Issue #17, opened 2013-06-27
Better onion skin


chchwy:
![OnionSkin](http://www.pencil2d.org/wordpress/wp-content/uploads/2013/06/14l53sj.jpg)

gordielachance:
Still would like to see this one !

feeef:
Great idea!

feeef:
Guys, I have implemented this feature! ;)

I didn't have much time to do it so it is a bit ugly in term of interface and I still have to cleanup the code a bit but it works! It will be in my next pull request.

You have to click on the small icon to toggle "absolute" / "relative"

![screenshot from 2015-12-18 17-28-06](https://cloud.githubusercontent.com/assets/383399/11901921/8d4f3c92-a5ae-11e5-80b3-a35bb8a46614.png)

![screenshot from 2015-12-18 17-28-26](https://cloud.githubusercontent.com/assets/383399/11901929/96b181a0-a5ae-11e5-9844-1bf2cbefbac2.png)

Have a happy Christmas animating! ;)

gordielachance:
Feeef, this is amazing :)
Great, great work. 
Really !

ManuAros:
Nice feeef, you make Pencil better and better. Have a very very Merry Christmas :)

feeef:
Thank you guys! :)

carlosgf:
A dream coming true.
The time-line is like the eyes of the animator.

... and now let us start animating :+1:

chchwy:
Implemented by feeef, closed.

PR #296, opened 2015-03-31
Add options to onion skinning so user may define the number of frames shown

thorjelly:
Implement this request: https://github.com/pencil2d/pencil/issues/17 . This first step adds these options to a setting menu. A better GUI interface, as discussed in the request, should be added later.

Adds options in Edit > Preferences > Tools so that the user may define how many previous and next frames are shown when previous and/or next onion skinning is toggled on. The user may also define a maximum opacity value and a minimum opacity value, which will be gradated between all of the frames shown. Replaces the previous options here which allowed the user to define the opacity values for the next or previous 3 onion frames explicitly; these settings were broken and did not work.


carlosgf:
@thorjelly 
Very nice for your help

PR #322, opened 2015-06-03
move translations to a dedicated directory resources/translations

mapreri:
as per https://github.com/pencil2d/pencil/commit/d570395207b3bdf6da425e9899d9767262da77e3#commitcomment-6552251

It currently fails to build, but looks like unrelated reasons. anyway, i couldn't test it.


mapreri:
thanks :)

PR #325, opened 2015-06-24
Fixes on bitmap brush engine.

feeef:
When working on a tablet, erasing with the back of the stylus worked but the current selected tool wasn't restored when using the point of the stylus again. It is fixed with this commit.

I also fixed the fact that when selecting a new tool, its settings where not updated correctly.

Finally, I made the pencil tool feel more like a real pencil.

Updates : I also fixed most of the tools options and made their controls appear or disappear depending on the option's availability for the selected tool. The non working options never appear.

Most of the bitmap brush engine is fixed in this version and is very usable! Only the polyline tool is not working completely as expected.

Thanks!


chchwy:
Nice work! @feeef

feeef:
Thanks Matt! :)

PR #342, opened 2015-11-07
Fixed colour behaviour in the Vector engine

feeef:
In the way the code is done right now, one has to use the colour palette when drawing on a Vector layer. Each vector stroke is linked to a colour in the colour palette so changing the colour in the colour wheel updates the current selected colour in the colour palette and all the vector strokes that have been made with this palette colour.

Changing this behaviour at this point requires a lot of refactoring so I left it like that as it is usable. It seems strange as it is different but I think it is great and can even be quite useful if we want to adjust colours in one go in the entire project. We just need to show users how it works.

I have also updated the onion skin system so it is displayed correctly across frames.


chchwy:
Thanks, @feeef

PR #347, opened 2015-11-20
Vector engine colouring

feeef:
I have made some updates on the Vector engine (mainly about colouring). The udates are explained in this video : https://goblinrefuge.com/mediagoblin/u/feeef/m/pencil-dev-update-colours-on-vector-engine/

Cheers.


MattJordan:
Very nice update.  Thanks for your contribution to the project.  And I'm glad you kept the default vector palette behavior since it really speeds up making color changes to an entire sequence.  And that's a great workaround for the bitmap layer.

PR #369, opened 2015-12-11
Weekly updates

feeef:
Here is my weekly contribution. The entire app starts to be pretty usable, I am quite happy with the result.

When you select an area with the select tool, you can transform your selection (move, scale, rotate) and cancel your transformation by clicking escape. The transformation is only applied on deselect.

For some reason, the move tool doesn't work well on the vector layer when a key is pressed, which makes rotation impossible on a vector selection. I will fix that for the next pull request.

Duplicating a frame now handles the entire frame and not just the screen area. It is also the case when you do select all. The entire frame is selected even if it is larger that the screen area.

Enjoy! :)

### DIFF
--- a/app/colorbox.cpp
+++ b/app/colorbox.cpp
@@ -5,7 +5,7 @@
 
 ColorBox::ColorBox( const QString& strTitle, QWidget* parent ) : BaseDockWidget( strTitle, parent )
 {
-    QVBoxLayout* layout = new QVBoxLayout(this);
+    QVBoxLayout* layout = new QVBoxLayout();
 
     m_colorWheel = new ColorWheel(this);
     m_colorInspector = new ColorInspector(this);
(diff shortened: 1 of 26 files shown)


carlosgf:
I am so happy that you are happy as well!

I read all your fixes and noticed this "Press Alt to temporary select Eyedropper tool"
... If I remember correctly, the Alt key was associated to the "blending" mode added years ago. It would work only when the SMUDGE tool is selected. Did you knew that?

I'll check it when I get the build working here.

THANKS A LOT!

feeef:
Thanks @carlosgf ! I am happy that you are happy that I am happy! ;)

I noticed a couple of bugs introduced because of these temporary keys. For example, you can't delete vector curves anymore. I will clean up all that for the next release... 

I will spend some time on pencil next week and then it will be holiday time for me. I hope to get to a pretty stable version before the end of next week for you all Pencil users to enjoy animating during the Christmas holidays. It is my Christmas gift! ;)

carlosgf:
:+1: 
Our gift to you @feeef  will be the animation videos done with Pencil2D!
☺

PR #381, opened 2015-12-18
Weekly updates

feeef:
Some fixes and 2 great features to make the timeline very usable.

- The ability to select and move frames around
- The ability to have onion skin match the keyframes (or not)

Enjoy! ;)

screenshot from 2015-12-18 17-28-06

screenshot from 2015-12-18 17-28-26


Jose-Moreno:
Great work François! It will surely come in handy :+1:  In the future we should see how this could be further extended so the animator can select how many (key) frames he /she wants the onion skin to show originating from the current frame under the time indicator like in TVPaint or the recent Krita.

feeef:
@Jose-Moreno I will add this feature next. I want to make it possible to extend the number of onion skin frames by dragging the last gray indicator over the timeline. The same way it is done in flash actually. I will also make those indicator be red on the left and blue on the right when the onion skin colors are on.

These 2 features should be quite easy to implement. They will come after the holidays! ;)

chchwy:
Great job @feeef.

PR #395, opened 2016-01-02
moved mEditor->setObject( object ) to correct position

JonasThomas:
Fixed my boo boo.  Happy new year.

### DIFF
--- a/app/mainwindow2.cpp
+++ b/app/mainwindow2.cpp
@@ -92,8 +92,9 @@ MainWindow2::MainWindow2( QWidget *parent ) : QMainWindow( parent )
 
     mEditor = new Editor( this );
     mEditor->setScribbleArea(mScribbleArea);
-    mEditor->init();
     mEditor->setObject( object );
+    mEditor->init();
+
 
     mScribbleArea->setCore( mEditor );
     mScribbleArea->init();


chchwy:
@JonasThomas, I have to say thanks for your contribution first, but I'm afraid that I'm going to revert this commit. 

Switch the order of `setObject()` and `init()` may lead other problems, we must ensure every Manager (ex. ViewManager or ColorManager) is created and init before `setObject()`, otherwise the managers will not got notified through `onObjectLoaded()`.

@tasaif has fixed the color problem in fc872de580 properly.

Hope you can understand :)

JonasThomas:
@chchwy   No prolblem the important thing is that it gets fixed overall, but the explanation is appreciated.
(I thought moving the setobject() befor the init() sort of stunk also  but it seemed to solve the problem in my head.  Ie it reloaded the colorPallete.
The fix that @tasaif does assumes that the first color in the colorPallette will always be be black.(This totally works for me btw)    
If someone has an itch to customize the colorPalette where the first color is not black, down the road that could be an issue.  I wondering if there should be a code comment for future develepors.

feeef:
@JonasThomas In the case of a custom color palette when opening an existing project, the palette is reloaded after the object is set through setObject() 

https://github.com/pencil2d/pencil/blob/master/app/mainwindow2.cpp#L494

JonasThomas:
@feeef Yep.... that's the rabbit hole I went down.. ;)   I was trying to rap my head around what was going on with setObject but there's so much going on it was hard for me to understand what was happening stepping into the code.

feeef:
I fixed this bug : https://github.com/feeef/pencil/commit/e5af91f2b21fe1f01dbd3b4ab454a875a39662ff

PR #398, opened 2016-01-03
Add dotted hover cursor

tasaif:
This branch adds a checkbox in the preferences
When the checkbox is on, the cursor is turned into two dotted circles for the brush, eraser, smudge, and pencil tools so that you have a preview of the stroke you're about to create.

The code is based largely on the circleCursors code


feeef:
Thank you very much @tasaif this is indeed a very good idea! :)

tasaif:
No problem

chchwy:
Thanks, @tasaif

Issue #385, opened 2015-12-21
Bug when drawing after moving bitmap objects with select and move tools


ManuAros:
1. Start fresh Pencil and draw something in frame 1
2. use select tool to select the object you drew
3. use move tool to move the object to a new location
4. Add one frame
5. Use pen tool to draw something in the new frame (2)
   and you will se that what you draw will be moved to about the same place 
   as you moved the object in frame 1

Comment: same behaviour even if you start with making two frames.
Comment: I don't see this behaviour on vector only bitmap layers

Tested on 32 bit Windows 10, Pencil2D Dec 19 2015 Build.

feeef:
This is one of the main bugs of the current build. It is very annoying. jonasthomas from the forum said he would have a go fixing it.

I have implemented the current transform feature so I can help understand how that works.

- The select tool is used to select
- The move tool is used to handle the transformation of the selecton
- The ScribbleArea handles the transformation at the BitmapImage / VectorImage level
- For a bitmap image (the problem is here!) the transformation display is generated by the CanvasRenderer (called in ScribbleArea paintTransformedSelection()). This is just a temporary display. It must be applied or canceled. If we go back painting or if we go to another frame without applying (deselect) or canceling (esc key) the transformation (ScribbleArea applyTransformedSelection() / cancelTransformedSelection()), the CanvasRenderer doesn't "switch off" its transformation mode.

ManuAros:
This can now be closed, tasaif fixed this issue.

PR #399, opened 2016-01-03
Fix select, move, brush bug

tasaif:
Apply changes of the move when switching to a new tool
This should fix the bug in #385

### DIFF
--- a/core_lib/interface/timelinecells.cpp
+++ b/core_lib/interface/timelinecells.cpp
@@ -9,6 +9,7 @@
 #include "layermanager.h"
 #include "playbackmanager.h"
 #include "preferencemanager.h"
+#include "toolmanager.h"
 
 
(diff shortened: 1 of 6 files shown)


ManuAros:
@tasaif 

Tested and verified working. Thank you for fixing this !

feeef:
Thank you @tasaif ! It is amazing to see more people involved in the devs! :)

I guess we should also apply transform changes when we leave the current frame / layer as well.

tasaif:
Should the transform changes be applied or canceled when switching between frames/layers?

feeef:
I guess it should only be canceled when pressing the esc key. Any orher "leaving" the transform mode action would apply the changes.

tasaif:
The move is now applied when switching frames/layers

feeef:
Thanks a lot! :) This will fix some very annoying behaviour that was introduced by the move tool!

chchwy:
Good job, @tasaif

PR #400, opened 2016-01-03
Bug fix

JonasThomas:
This fixes Black become while on the color pallet  when File->New is chosen.
However, it appears the nothing is drawing on the scribble area.

### DIFF
--- a/app/mainwindow2.cpp
+++ b/app/mainwindow2.cpp
@@ -1,4 +1,4 @@
-/*
+ /*
 
 Pencil - Traditional Animation Software
 Copyright (C) 2005-2007 Patrick Corrieri & Pascal Naidon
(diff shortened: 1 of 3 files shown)


JonasThomas:
basically file->new , load files,  needed to have init called after the object was set in the editor so the colorpallete freshes correctly.

tasaif:
Tested this PR
current master branch behavior:
- black brush strokes work
- new file
(color switches to white)
- change color to black
- strokes no longer work

your master branch:
- black brush strokes don't work from the onset

your bugFix branch:
- black brush strokes work
- new file
(color does not switch to white)
- strokes still don't work

tasaif:
Fixed the color thing in #405

JonasThomas:
Yah... that camera switching thing messed me up too. ;)  While the Black is White is fixed.  I noticed that some of the custom colors in the colorpalette many be turning black.   Are you seeing that as well?

tasaif:
I haven't noticed that yet

JonasThomas:
It seems like the Black is white issue has been by tasaif and merged into the master so it seems this is no longer required.

Issue #397, opened 2016-01-03
Black becomes white in color palette. [BUG]


JonasThomas:
ubuntu 14.04 64 bit 
b778b02eafc69fbec3cc755a496c95501163ba8d
I thought I had fixed this, but apparently not.
When the application starts the colored block in the first line starts out as black.
If file new is selected the black box become white.
If a file is saved and later retrieved it becomes white.
I think this the block of code that may be turning the block white is here:
ColourRef Object::getColour( int i )
{
    ColourRef result( Qt::white, "error" );
    if ( i > -1 && i < mPalette.size() )
    {
        result = mPalette.at( i );
    }
    return result;
}
(I'll figure it out and fix this)

JonasThomas:
I refactored the getColour code to be able to put a break point on the error condition and apparently that's not what's causing the problem....  hrrrg.

JonasThomas:
Black is white on file new is fixed with this. ee9964a806baad8ccd71197bac359603c1fb47a9
There seems to be a problem that nothing is drawing with the pencil after file new is selected.

JonasThomas:
The issue is that when file->new is selected the camera layer is automatically selected instead of the bitmap.

JonasThomas:
This seems to be working now with fixed by tasaif.
Thanks.

Issue #374, opened 2015-12-14
[BUG] - New document breaks the timeline


feeef:
After doing some work with Pencil, going to File -> New visually breaks the timeline.

ManuAros:
I don't know if this belongs here but I also noticed that when you are drawing and you start over with a new file (with File -> New) you can't start drawing before you move focus to another layer and back to bitmap layer. I don't know if this is because it's the camera layer that is set as default layer when you choose FIle-> New.

carlosgf:
Maybe I should close this #300 ?, which is the same as yours. Besides that it is best understood with your words.

(There is no need to keep duplicate issues) :)

feeef:
Sorry for the duplicate!

carlosgf:
No problem!

(I will close mine. You have the best vocabulary for explaining things)
:+1:

tasaif:
Screenshot please and steps to duplicate bug ("doing some work" is vague)

@ManuAros that should be a separate issue

feeef:
You're right @tasaif. This bug was primarily a bug reminder for myself :)

I haven't been able to reproduce this bug where the layout is broken. It was the keyframes from the previous project still being half visible in the timeline. I have been doing some work on the timeline display when implementing frames dragging so I may I fixed it.

The other thing was the camera layer being selected by default. It is not intuitive and looks broken. We should have the bitmap layer selected by default just like when the app starts up.

feeef:
I close this bug as I can't reproduce it. @ghost should re-open his bug as it was describing the rest of the bug. sorry about that!

PR #404, opened 2016-01-04
Start on bitmap layer

tasaif:
When creating a new document, bitmap layer should be selected by default
#374 #397 #409

### DIFF
--- a/core_lib/interface/editor.cpp
+++ b/core_lib/interface/editor.cpp
@@ -496,6 +496,10 @@ Status Editor::setObject( Object* newObject )
 void Editor::updateObject()
 {
     scrubTo( mObject->editorState()->mCurrentFrame );
+    if (layers() != NULL)
+    {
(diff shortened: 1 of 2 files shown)


JonasThomas:
Cool.. Look forward to giving this a try tonight after work.  (Are you on the forum?)

feeef:
Thanks @tasaif ! :)
As we are now a few contributors to this code, we should talk to organise ourselves. If you are not in the forum please join us : http://www.pencil2d.org/?post_type=forum

JonasThomas:
I was hoping to give this a try, but it hasn't been merged yet.  I suppose I can pull it from your repo.

tasaif:
Sorry

JonasThomas:
Not a big deal.  It is sort of a gotcha bug though.

JonasThomas:
Sweet...   This is one of those things that you don't even notice when your using the app, but sure is annoying when it doesn't work..  Tested all the permutation, where I noticed it... All better as near as I can tell.   Thanks,

PR #410, opened 2016-01-06
Fixes and improvements

feeef:
Some fixes and improvements. Mainly on the vector engine (we still need to improve the bucket tool on the vector engine)

I also updated the ugly icon to toggle onion skin match keyframes.

I fixed a bug where the view was not correctly reset after creating or opening a new document.

I fixed the palette loading


chchwy:
Thanks, @feeef !

Issue #353, opened 2015-11-22
Bug when switching tools while still drawing


Jabst:
Hey this might be somewhat complicated to explain but I will try my best to be the most concise.

When I draw a line using brush and I hotkey to the pen and continue to draw that line but when I switch back to the brush I notice that there is a straight line that links the last coordinate on the first use of the brush(x,y position on the canvas) and the first coordinate of the other use of the brush.

I will attach an image of the resulting canvas.
![bug](https://cloud.githubusercontent.com/assets/9280778/11325757/f1c3e8a0-914d-11e5-92e7-83c7043affb5.png)

carlosgf:
Same thing here. 
(drawing a path like a Serpent, you will get that new issue line between curves)

Win8.1x 64bits

PR #413, opened 2016-01-07
Drawing bug 353

tasaif:
Solving #353 

This PR fixes switching between brush and pen; however, it doesn't work the first time you draw a stroke (only subsequent times). Please don't merge this PR until I track down the source of this problem. The other stroke tools should be fixed before merging as well.


tasaif:
This shouldn't be a PR

PR #421, opened 2016-01-12
added option to type brush and feather size

CandyFace:
This should make it much easier to select precise values.

~~ideally i want to remove the smaller brush number and update the spinBox simultaneously instead. Right now it's only updating when you let go of the slider and of course when you insert a value or use the arrows or use the shortcuts to change size.~~
fixed it! now, as there is no need for the brush number to be shown twice anymore, i have removed the smaller text showing its value and instead made it so the SpinBox updates simultaneously with the slider.

the new UI looks like so:
![Visual representation]
(http://i.imgbox.com/4mDuOTZv.png)

### DIFF
--- a/core_lib/interface/spinslider.cpp
+++ b/core_lib/interface/spinslider.cpp
@@ -25,14 +25,9 @@ SpinSlider::SpinSlider( QString text, GROWTH_TYPE type, VALUE_TYPE dataType, qre
     QLabel* label = new QLabel(text+": ");
     label->setFont( QFont( "Helvetica", 10 ) );
 
-    mValueLabel = new QLabel( "--" );
-    mValueLabel->setFont( QFont( "Helvetica", 10 ) );
-
-    mValueLabel->setAlignment( Qt::AlignRight | Qt::AlignVCenter );
-
     mSlider = new QSlider(Qt::Horizontal, this);
-    mSlider->setMinimum( 0 );
-    mSlider->setMaximum( 100 );
+    mSlider->setMinimum( mMin );
+    mSlider->setMaximum( mMax );
     mSlider->setMaximumWidth( 70 );
 
     QGridLayout* layout = new QGridLayout();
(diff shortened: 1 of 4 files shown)


tasaif:
I've reviewed this pull request, and it is my opinion that it should be merged.

core_lib/interface/spinslider.cpp
```diff
@@ -85,21 +73,20 @@ void SpinSlider::setValue( qreal v )
     int value2 = 0;
     if ( mGrowthType == LINEAR )
     {
-        value2 = std::round( 100 * ( v - mMin ) / ( mMax - mMin ) );
```
tasaif: what's this line about?

core_lib/interface/spinslider.cpp
```diff
@@ -70,13 +58,13 @@ void SpinSlider::onSliderValueChanged( int v )
...
     {
-        value2 = mMin + v * ( mMax - mMin ) / 100;
+        value2 = mMin + v * ( mMax - mMin ) / mMax;
     }
     else if ( mGrowthType == LOG )
     {
-        value2 = mMin * std::exp( v * std::log( mMax / mMin ) / 100.0 );
+        value2 = mMin * std::exp( v * std::log( mMax / mMin ) / mMax );
     }
-    changeValue( value2 );
```
tasaif: let's not leave commented code hanging around
@CandyFace

tasaif:
I've reviewed and tested this PR.

CandyFace:
I've updated the PR, yeah honestly @tasaif i'm not sure why i changed 100 to mMax i somehow thought that the value was set to 100 which it hasn't. I've reverted that mistake. From what i understand, when LINEAR growth equation has been chosen, it makes sure that min and max matches the size of the slider, ie. not stopping only 2/3 on it's way for example.

chchwy:
Good work @CandyFace , Thanks.

CandyFace:
Ohh... that's what the QSignalBlocker did.. ah yeah i see. I wasn't sure about it's functionality but i can see that now. I should have looked it up like i did with the rest :3

Issue #430, opened 2016-01-14
Color wheel UI enhancement


CandyFace:
I've been playing with positioning of the color wheel and the color inspector, as the screen is much wider than it is tall, i think it would be ideal to move the color wheel next to the color inspector instead of having the them underneath each other. Although i'm not entire sure on that that yet... Maybe we could just split the color wheel and the inspector into separate windows, then people would be able to customize that themselves. 

I have so far made it possible to resize the color wheel, to fill less if needed. What i have achieved looks like the second/right image.. but my goal is to have the wheel aligned in center and also make the wheel thinner when resizing.

![Color wheel]
(http://i.imgbox.com/u98OxvBz.png) ![Resizing color wheel]
(http://i.imgbox.com/Rcs84hPL.png)

I'm not much of a math person, so i'll try to keep my boundaries within UI and interface and use my knowledge to develop those parts.

Jose-Moreno:
@CandyFace Hey man, just wanted to say THANK YOU for the tools size sliders in a previous PR!

This looks great too and It's really needed to be able to change from HSV / RGB / HEX particularly if you're working with other software without having a palette file to export. Keep up the great work!

tasaif:
@CandyFace should this be closed now?

CandyFace:
Yes please close it

PR #431, opened 2016-01-15
Centered color wheel and square, set contrains for widget #430

CandyFace:
Color wheel and color square has been centred to the widget and is resizable to a certain extend. 
Fixed the region focus when resizing the window, now it is possible to resize and still
use the colorwheel hue indicator.
m_wheelWidth has been renamed to m_wheelThickness. 
color label has been positioned below inspector instead of next to.
constrained the widget to not be able to resize to fill the whole application window.
#430


feeef:
Well done @CandyFace ! It looks great! :)

Issue #424, opened 2016-01-13
Grabbing Layer crashes Pencil2D


tasaif:
Grabbing any layer (bitmap, vector, camera, sound) to the bottom of the Layers panel crashes Pencil2D

- Kaiko

JonasThomas:
@taslf   I'm having trouble verify this bug.  I'm working if it's been fixed, or it's occuring when other weird things are happening.

JonasThomas:
@tasif http://www.pencil2d.org/forums/topic/bug-grabing-a-layer-crashes-pencil2d/#post-27788
Posted to the originator for futher info

JonasThomas:
Bug confirmed.   There is a segmentation fault that's occurring during a drag and drop event.   The seg fault occurs when a layer is grabbed and dropped outside of the frame of the application.    @chchwy  I don't think I have privileges to assign myself  this bug.. I'll take a look at this.

JonasThomas:
The segmentation fault occurs here:
void TimeLineCells::mouseReleaseEvent( QMouseEvent* event )

specifically in this section
 if ( m_eType == TIMELINE_CELL_TYPE::Layers && layerNumber != startLayerNumber && startLayerNumber != -1 && layerNumber != -1 )
    {
        mEditor->moveLayer( startLayerNumber, layerNumber );
    }
When I set up a break point of and drag the layer off the screen I get a layerNumber = -4
@feeef @tasaif     quick fix would be to change  !=-1  to >=0   although I suppose @chchwy would probably want the layerNumber set to -1

JonasThomas:
The real critter is in int layerNumber = getLayerNumber( event->pos().y() );

tasaif:
@JonasThomas should this be closed now?

JonasThomas:
Yes

PR #433, opened 2016-01-16
Fixes bug https://github.com/pencil2d/pencil/issues/424

JonasThomas:
Should this have been put in on a branch other than the master?


JonasThomas:
I think this should fix it. https://github.com/pencil2d/pencil/pull/433

JonasThomas:
initially I thought that the seg fault occurs when you drop and release a frame button outside the application frame, but also if you drop the button past the last row.(This fixes both situations) 
segmentationfault

chchwy:
Thanks @JonasThomas

PR #437, opened 2016-01-17
Added the ability to see the current zoom level and disable feathering

Spark01:
Brush feathering can be disabled and reenabled through a simple
checkbox, no harm done.
Zoom level is seen on the display options widget. It's not pretty but it
works.


Spark01:
I suppose you're right, I'll wait until I've added some more features/bugfixes before the pull request.

Issue #432, opened 2016-01-16
Alpha is not changing opacity of the current color palette


CandyFace:
When you change the number of the alpha channel, it does not change the opacity of the current color block.
here is a gif of the bug. First when you change one of the three other color values, the alpha is applied to the box. this happens on OSX 10.11.2 with Qt 5.5.1 btw.
![color bug]
(https://zippy.gfycat.com/SneakyAgitatedArmedcrab.gif)

PR #448, opened 2016-01-20
Alpha 432

tasaif:
#432 
This bug was very difficult to find. The signals and emits code is very convoluted, and I still don't understand the full implications of this bugfix. I encourage a stringent review of this PR.


CandyFace:
I tried to apply this commit to my own source but i'm getting "reference to non-static member function must be called. This is with my applied PR.

PR #463, opened 2016-01-30
Update translations & show zoom ratio on status bar.

chchwy:
1. Update translations from OneSky
2. Move zoom ratio to status bar from display option widget.


tasaif:
I don't know how to review this pull request. There are 10k line changes and touch areas that I don't understand.

CandyFace:
The first commit are just translations from Onesky service, there shouldn't be any critical stuff in there. although if there is any weird tags to be found in those translations, the compiler will fail to build. The merging seems to include old changes which have been applied already. They could potentially overwrite new stuff.. but the commit is so big, that it's unmanageable to review.

feeef:
Yes, it is mainly translations so it shouldn't affect the logic of the application. It would need to be tested manually.
I guess we can trust @chchwy .

chchwy:
I download the new translation ts files from OneSky and escape the html tag in commit 0aa2163b, something like <br> to &lt;br&gt; . 

The easiest way to review this PR is to launch Pencil on a non-english locale computer, and check the  translation of UI text :p

chchwy:
OK, i'm going to merge this PR.

feeef:
Thanks @chchwy !

PR #470, opened 2016-02-08
Bug fixing

feeef:
- Fixed eraser behavior when using it through the SHIFT + CTRL keyboard shortcut.
- Fixed crash when moving the first frame in the timeline (re-create a new first frame)


chchwy:
Thank you! @feeef

PR #474, opened 2016-02-12
Bug fixing

feeef:
- Fixed some bugs with the select tool that made a moving selection visually interfere with other layers.


feeef:
This fix was important for the usability of the select tool so I merged it.

PR #478, opened 2016-02-12
Clamped view scale (zoom) to [1%-10000%]

Squeenix:
You can currently zoom out to infinity or float overflow which is probably an issue.  All this does is clamp the zoom to 1-10000%.


chchwy:
Thanks @Squeenix !

PR #485, opened 2016-02-19
Bug fixing

feeef:
Fixes on the camera layer : 

- Camera border now follows zoom, pan and rotate the canvas
- Removed Camera border setting in display option. Every camera, as a layer has its own visibility
- Auto detect the last used camera when exporting. There is no dialogue asking you to select a camera anymore
- Fixed a bug where the camera border was not displayed after moving the camera layer


chchwy:
Good job, thanks @feeef

feeef:
Thanks @chchwy ! We are very close to the 0.6 version! Pencil2D starts to be very usable.

PR #492, opened 2016-03-08
Bug fixing

chchwy:
- Fix: menu items are not at correct state at application start.
- Fix no icons on menu items OnionNext/OnionPrev.
- Introduce a 2-way binding method to connect menu item and setting on/off.


Jose-Moreno:
@chchwy Great work!

PR #493, opened 2016-03-14
Fixed Export as Image Sequence without selecting the camera.

feeef:
Quick fix for image sequences export.

It is not necessary to select the camera when there is only one. If there are several cameras, the last created or selected one will be used for the export.

### DIFF
--- a/core_lib/soundplayer.cpp
+++ b/core_lib/soundplayer.cpp
@@ -60,6 +60,9 @@ void SoundPlayer::makeConnections()
 {   
     QObject::connect( mMediaPlayer, &QMediaPlayer::mediaStatusChanged, this, [ this ]( QMediaPlayer::MediaStatus s )
     {
+        // WARNING :
+        // This call is not supported in QT 5.3. Is it necessary?
+        //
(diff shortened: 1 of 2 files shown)


core_lib/soundplayer.cpp
```diff
@@ -60,6 +60,9 @@ void SoundPlayer::makeConnections()
 {   
     QObject::connect( mMediaPlayer, &QMediaPlayer::mediaStatusChanged, this, [ this ]( QMediaPlayer::MediaStatus s )
     {
+        // WARNING :
+        // This call is not supported in QT 5.3. Is it necessary?
+        //
```
chchwy: What is the error message?

feeef: The error message is : 'QObject* QObject::sender() const' is protected

chchwy: OK, I got it, I'll fix it in next PR, thanks.

feeef: Thanks Matt! ;)

PR #501, opened 2016-05-21
Added transparency checkerboard to color inspector preview box

scribblemaniac:
This adds a checkerboard background underneath the color preview box in the color inspector to make it easy to see when a color is transparent. Here's what the end result looks like:
Color Inspector Preview


Jose-Moreno:
@scribblemaniac Thank you for helping out! Every commit counts! :smile:

scribblemaniac:
@Jose-Moreno Glad I could help! 😄

PR #503, opened 2016-05-21
Improved mac experience

scribblemaniac:
A summary of the changes I made:
- Added a few missing resolutions to pencil.icns
- Created custom Info.plist for the application bundle.
- Registered for the QApplication:event to detect mac file opening (by dropping on the application icon or with Open With...). Some minor refactoring of main and mainwindow2 was required for this.
- Defined file associations for .pcl and .pclx files so that double clicking on them will open them up in Pencil.
- Added file icons for .pcl and .pclx files. They look like this now:
Mac file icons


chchwy:
:+1:

PR #507, opened 2016-05-29
Fix Linux build : The filename needs to be case sensitive

benjarobin:


chchwy:
Thank you @benjarobin !

Issue #506, opened 2016-05-28
Option to change language?


hankofficer:
Hi all,

I'm new to Pencil2d. I'd like to use Pencil2d in Traditional Chinese, but my WIndows 10 is in English. Is there any way to change language? If not I could try to change it in code.

Also I have cloned and built the project with Qt Creator 4.0.0, Qt 5.6 and MSVC 2015. Therefore I'm in Nightly Build May 28 2016.

Thanks,
Hank

Jose-Moreno:
@hankofficer Hello, thank you for your interest in Pencil2D. We already have everything setup and more than 7 languages to translate which have been contributed by our users. Our lead developer Matt Chang, has told us this week that he will be working to finalize the translation switch as well as other 2 requested topics. You can read more here: https://github.com/pencil2d/pencil/issues/505

For now, it would be an incredible help if you could also look out for bugs reported in the tracker and help us out, since there are not enough developers working on Pencil2D. 

I also think that the code has been migrated to QT 5, but I can't be sure, maybe @chchwy could tell you more about the technical requirements for making Pencil2D work in order to develop :smile: 

Thanks a lot for your interest in the software!

chchwy:
Hi @hankofficer,
if you don't mind waiting for about one week, this feature is under developing and maybe you could see it in the next nightly build. :)

hankofficer:
@Jose-Moreno Thanks! And sure, I'd love to contribute a few codes once I understand how the software works. I have tried Qt to develop a program before but I'm still not familiar with it. It will take times. Btw, If possible, I also want to stick with Pencil2D to make some animations instead of those expansive animation software I cannot afford. x(

@chchwy Good to know! I'm looking forward to next build. :D

PR #509, opened 2016-06-09
Update translation and Language switcher

chchwy:
- Update translation from OneSky
- Add a new language switcher as a drop-down list in Preference dialog.


chchwy:
#506

PR #518, opened 2016-06-24
Motified zooming when using mouse wheel.

hankofficer:
I think this would be nice to have, but it's probably just my opinion. I'd like to have a discussion about this. :3


chchwy:
How do the values of zoom ratio come from? just curious.

hankofficer:
@chchwy Most values are from Paint Tool SAI (I draw most of my arts with this program. XD). I also derived a few values from it.

PR #519, opened 2016-06-26
Fix Pencil2D not playing sound

chchwy:
- Use Qt Multimedia lib as sound player
- Sound files will be copied to data folder on saving
- Show different size of sound clip on Timeline
- Fix misc UI behaviors about sound layer


Jose-Moreno:
@chchwy Great work Matt!

chchwy:
@Jose-Moreno thanks!

PR #527, opened 2016-08-16
fix #511 title gets an * at the beginning when any change happens to …

### LINKED ISSUE #511: [Suggestion] Show unsaved sign on the program title
I think it would be nice to have a star sign after filename if the file haven't saved. e.g.
`C:/user/mybox/myartwork.pclx* - Pencil2D`

lordadamson:
…indicate unsaved changes

This includes modifications in the scribble area only though. not sure if that will be enough. provide me with any necessary feedback if needed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,7 +38,7 @@ Pencil2D.app
 # QtCreator
 
 *.autosave
-
+build-pencil-Desktop-Debug/*
 
 # Mac stuff
 
--- a/app/mainwindow2.cpp
+++ b/app/mainwindow2.cpp
@@ -205,6 +205,7 @@ void MainWindow2::createDockWidgets()
     makeConnections( mEditor, mColorPalette );
     makeConnections( mEditor, mDisplayOptionWidget );
     makeConnections( mEditor, mToolOptions );
+    makeConnections( mScribbleArea );
 
     for ( BaseDockWidget* w : mDockWidgets )
     {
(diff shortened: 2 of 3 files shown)


Jose-Moreno:
@lordadamson I don't have merging rights, as I'm not developing however @chchwy is the lead developer and he can approve all the incoming commits, so give it a bit of time since he's usually busy with his job but then he can review your commit and merge it to master. Thank you for helping out!

scribblemaniac:
Minor bug: it looks like you are adding two characters (asterisk + space), but only removing one of them (`...remove(0, 1)`). I can't actually test whether this is problem right now, but I would imagine that each time you save, a space gets added to the title.

Additionally it may be better to connect to new signals in `Editor::backup`, `Editor::undo`, and `Editor:redo`. I see two main advantages to doing it that way. First, any changes should go on the undo stack, so backup will theoretically be called for all file changes, regardless of whether or not they modify the scribble area. Second, imagine that a person makes modifications, saves, makes more modifications, and then undo all of the changes they made since they saved last. By checking the backupIndex value you could detect that the file is at the same state as when it was saved, and mark the title as saved accordingly. I have noticed some other programs that have this behavior. I'm not very experienced with the codebase here though so take what I have to say with a grain of salt.

lordadamson:
I found that actually I'm not using that function anywhere because I found that it's unnecessary because when saved the title gets replaced anyway with the `setWindowTitle( strSavedFileName );` in the `saveObject` function.
So yeah let me remove it.

And that sounds really good, I'd really like to implement that. I'm not that familiar with the code base myself, I was just introduced to it yesterday :)

chchwy:
Thanks for your contribution @lordadamson!

lordadamson:
You welcome!

PR #536, opened 2016-09-25
Fix memory leaks at importing images

edwrodrig:
Several object allocation without calling deleting it after.

Due object are used in the same scope, heap allocation is not needed, intead using stack allocation for better resource management (auto delete and file close).


Jose-Moreno:
@edwrodrig Whoah! You're awesome man! hope @chchwy can merge it soon! Thank you for contributing to Pencil2D! :smile:

chchwy:
Thank you! @edwrodrig

PR #543, opened 2016-10-15
Spinbox focus loss bug fix

GregorysonofCarl:
Here's a tiny baby commit.

This modifies the program so spin boxes in the tool options widget don't lose focus when typing values in with the keyboard. 
ex. You can now input 34 as the brush size without needing to click back into the spin box to type the second digit.


Jose-Moreno:
@GregorysonofCarl Thanks a lot! Every effort counts and so does yours! :smile:

PR #554, opened 2016-10-30
Improve import import and export dialogs

scribblemaniac:
Some very minor changes to the file type filters for importing and exporting. Changes:
- Modified formatting of export option labels to match import option formatting
- Added support for importing TIFF sequences
- Set the default import option to all image types. This is especially important for Mac, since the file type option is hidden by default, potentially leading people to think that only PNGs can be imported.


chchwy:
Thank you! @scribblemaniac

PR #556, opened 2016-11-02
Add Linux build doc

scribblemaniac:


scribblemaniac:
Good to know, I built it with 4 a while back so I assumed it was supported. I will update soon.

PR #563, opened 2016-11-07
Working ffmpeg call on Linux systems (patch)

ptrg:
This patch is working on my linux system without any problem.

You must/may add ffmpeg as dependency or as a recommended package for Pencil2D's .deb and .rpm packages.

ffmpeg is a standalone project/package so if you install Pencil2D on a Linux system (normally with a package manager) then you can not use it in a plugin-like directory structure (as you do it on windows or apple systems).

The windows/apple way may work if Pencil2D is installed in user's home directory (/home/auser/Pencil2D) then ffmpeg can be placed in a subdirectory (/home/auser/Pencil2D/plugins), but this way each user on the system must install a new copy of Pencil2D and ffmpeg.

In my oppinion the best solution on all systems would be to search ffmpeg on the PATH and then (if not found) ask the user to select the ffmpeg executable manually.

### DIFF
--- a/core_lib/movieexporter.cpp
+++ b/core_lib/movieexporter.cpp
@@ -89,7 +89,7 @@ QString ffmpegLocation()
 #elif __APPLE__
     return QApplication::applicationDirPath() + "/plugins/ffmpeg";
 #else
-    return "";// TODO: linux
+    return "/usr/bin/ffmpeg"; // ffmpeg is a standalone project.
 #endif
 }
 


chchwy:
Thanks for your contribution and suggestions @ptrg !

I believe that searching ffmpeg executable on PATH would be the right way on Linux, but Windows/Mac users are so lazy that we don't use package manager and expect that everything is included in release package.

However, It would be very welcome to send any further pull requests to improve user experience on Linux. Personally I don't use Linux so much, so we eagerly need more linux contributors.

And let's call @mapreri  for the dependancy issue :p

mapreri:
\*blink\* :)

Yes, anyhow in the Debian package there is already a ffmpeg from back when it was "working" before you removed the support to export movie while rewriting it (and I was too lazy to remove it).

Btw, to search from PATH there I suggest using [QStandardPaths::findExecutable](https://doc.qt.io/qt-5/qstandardpaths.html#findExecutable); that way you can also remove that logic "if Win then ffmpeg.ext elif mac ffmpeg else /usr/bin/ffmpeg" too (it adds .exe as needed by itself).

Issue #540, opened 2016-10-11
Official Roadmap, Developer Guide and Bugtracker Template [Help wanted]


Jose-Moreno:
Hey everyone, I'm glad that we've had some help here and there by many generous developers who have been curious enough to help @chchwy and @feeef develop both the main branch and the "Mypaint" branch, however I think it's time we start moving forward and for that I believe we need a few things to get sorted:

1) One of the things I've noticed we're missing and that many would-be developers have asked me for is an **OFFICIAL ROADMAP**. 

In the past I tried to ask people to collaborate in order to create one but that didn't work. So now I want to ask if you can propose how we can approach the creation of a unified roadmap so other developers that drop by know what's next.

Some open source projects have had success with public **Trello** boards, but **if you know or use a better app / website or else, please suggest it in the comments**. We should look to unify our efforts towards coding the software.

2) We desperately NEED a **DEVELOPER GUIDE**. There was one in the website on how to compile the source but it got buried due to the hacker / spammer dramatic arc. Not all is lost since that guide required some updating. However we really need to understand not everyone uses Qt, despite being so spread.

We particularly we require MacOSX and Linux guides for people who want to compile form source for those platforms. Who can help with this? we would upload this in the static github page so we don't need to fear another shutdown, but we mainly need consistency.

It should at least have:
+ How to download Pencil2D source (Github vs bitbucket, what and how to download)
+ How to download and install Qt framework (account creation, version number, install options)
+ How to compile from source (How to open the project
+ An explanation of how Pencil2D is constructed, is it modular? is it event driven? and a brief summary of each library / module / class. If someone wants to help modify the vector engine, where do they start? (This will take some time to pull off, but when it's done it will be worth it)
+ How to Test the source (how to compile / build and Execute)
+ (optional) How to create binaries to help test new features
+ Suggestions are welcome

3) Lastly thanks to a suggestion by another user in this thread #535 I created an **issue template** to help us out in guiding new people on how to report bugs. Ideally Matt can implement this through Github since he's the admin and repo owner.

For more info about that go here: https://github.com/blog/2111-issue-and-pull-request-templates

And to download the text template do it here: 
https://dl.dropboxusercontent.com/u/59671876/ISSUE_TEMPLATE.md

The file is already named as suggested by the previous guide and it's also a markdown file.

scribblemaniac:
Very nice write up! Here are a few of my personal thoughts on the matter.

[...]

PR #570, opened 2016-11-14
Shapes tool

Piehigh:
I have been playing with pencil 2d abit and i think that it is important to have a shapes tool like in paint or in ms office products. This would make drawing so much easier and would be a great help. Also I think that it would also be nice to have layers. This would mean i could have different body parts as a different layer so i could keep the body consistent while still being able to change the arms without changing the body.

shape tool box
image


Jose-Moreno:
@Piehigh Yooo! That is quite an interesting Idea. I don't oppose to it, however next time it would be nice to discuss your ideas for the software with everyone first here #540 The thing is we need to begin shaping Pencil2D into a proper animation tool. The vector department is a bit forsaken to be honest, so some pre-baked vector shapes would be nice. Although what is missing most is a bezier tool to allow for proper "inking"

The other thing that bothers me is, are you making a featurerequest? or are you comitting a change you made? why are there so many commits the same as previously contributed by other users?

If you're making a commit did you update your branch prior to making your own changes? I'm afraid I'm still learning how these Github differencial patches work, but it's weird that there are so many old commits that were already ...well....committed.

If you're just making a request, please don't try to merge another branch since it's a bit confusing. Thanks! :)
----
@chchwy can you take a look at this? you're the only one that can approve stuff and I'm not sure what's happening with this merge to be honest.

chchwy:
Hi @Piehigh Thanks for bringing up the issues. 
But you may misuse the "pull request" function, this is for developers who change the source code and want to send their changes to main repository. I am going to close this issues and we can continue discussion in #571

PR #572, opened 2016-11-14
Fix scrollbar not updating when project is loaded

scribblemaniac:


chchwy:
Thanks @scribblemaniac

Issue #244, opened 2014-02-10
[BUG] Pen tool in vector mode - undesired "merge" behavior


blurymind:
This is something that completely ruins newly created and previous lineart strokes in 80% of the time.
It would be better if Pencil2d at least did not alter the older line strokes that the artist is presumably already happy with. The auto glue behaviour should only affect the newly created line stroke.

I made a short video in order to illustrate where this automatic "merge" behavior simply ruins inking workflow.
http://youtu.be/ArnFs9gC4r4

My request is to disable automerge completely by default 
and/or let the artists disable it optionally with a tick box!

It makes inking work very frustrating and counter-productive.

Jose-Moreno:
**Dev Notes**
---
As of 2018 the vector engine has become beyond broken, since the original behavior described in this issue is no longer present (as merging was removed entirely) the engine needs a complete overhaul considering the newest changes made it almost unusable. Once this engine has been enhanced please check on this issue again to consider it as a **feature request** rather than a bugfix in order to have an optional merging behavior for vector control points on fills and strokes.

PR #590, opened 2017-01-16
Added yes-no option for vector merge

martinvanzijl:
A fix for #244. I added a checkbox called "Merge" to the Tool Options panel. When it is checked, new vector lines are merged with existing ones. If unchecked, this merging does not happen.

By default it is un-checked.

This checkbox is only visible when both 1) a vector layer is selected and 2) either the pencil or pen tool are selected. (The pencil and pen tool are the only ones which can do vector merging).

Below is a screenshot:

bug_244_new_checkbox

Below is a video demo (zipped):

[Bug_244_Fix_Demo.zip](https://github.com/pencil2d/pencil/files/707642/Bug_244_Fix_Demo.zip)

What do you think?


Jose-Moreno:
@martinvanzijl I think it's great, surely a great step in the right direction. The vector tools definitely need more love and at least people won't feel constrained to close it. Although i'd also say that a handy feature would be a modifier would to turn ON the function temporarily, because sometimes you need the vectors to close and fill them with the bucket tool and some others to just "ink" on the fly. 

Great job and thanks a lot for your contribution!

Issue #545, opened 2016-10-16
[BUG] Bitmap drawing transferring over to newly opened file with PCLX extension


Jose-Moreno:
### --Issue Summary--

Highly specific problem when opening a saved blank file wih the .PCLX extension. Does not happen with .PCL files.
### --Actual Results

When you draw in a file with any color, except black, save your file, then open a blank file, the strokes from the first keyframe will transfer over to the newly opened file and will shift position on the canvas, even if the canvas remains the same size between files.
### --Expected Results--

Every blank file or files with the first frame empty should behave independently from any previously worked or saved file. The user shouldn't need to worry about possible bitmap information transferring over to a file that has the first frame blank. 
### --Steps to reproduce--

1. Create a NEW file.
2. Select the bitmap layer
3. Draw a stroke on the first frame with any color from the palette except black and any tool (pencil, pen, brush).
4. Save with PCLX extension. I'll name it BugTest_stroke for reference.
5. Create another NEW file.
6. Add a new frame on bitmap layer
7. Draw a stroke on second frame for comparison. First frame MUST be empty.
8. Save file with PCLX extension. I'll name it BugTest_blank
9. Now just re-open "BugTest_blank.pclx" via File > Open
10. You'll notice now that the first frame will be populated with same stroke from BugTest_stroke shifted in position.
### --System Information--
- Pencil2D Version:
  Nightly Build Oct 10 2016
- Operating System:
  Windows 7 Ultimate 64bits
- RAM Size:
  4 GB
- CPU:
  AMD Phenom II x4 965 BE
- Graphics Tablet:
  Wacom Intuos 5 Pro with driver: 6.3.15-3

Jose-Moreno:
Report Update:

user goody2shoos on our community forum reported a connected problem in this thread: http://www.pencil2d.org/forums/topic/recurring-pasting-bug/

> I’ve been sometimes getting this weird bug where an old image of mine from another Pencil file keeps popping up while working. The weirdest thing is that the tiger doesn’t even exist anymore since I continued the drawing on it’s file. I think it occurs when I either copy and paste something from another frame, or make a new frame. This has also happened with a drawing in the same Pencil file.
> 
> Something to note is that I copy and pasted a lot of frames while working on the tiger animation.

I think this has to do with how the images are being cached into memory, but I have no idea what exactly would be the problem. Hopefully these steps are enough to pinpoint the problem :smile:

CandyFace:
[...]

PR #591, opened 2017-01-17
Fix for bug #545 (sticky frame)

martinvanzijl:
This is a fix for #545. The temporary working folder is now cleared to prevent this bug occurring.

### DIFF
--- a/core_lib/structure/object.cpp
+++ b/core_lib/structure/object.cpp
@@ -48,6 +48,12 @@ Object::~Object()
     {
         delete mLayers.takeLast();
     }
+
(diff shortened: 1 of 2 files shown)


scribblemaniac:
Thanks for another contribution! I have a couple of issues with this PR though. First, you should call `createWorkingDir` after `deleteWorkingDir` to make sure the working directory exists and is properly initialized. Second, the `dataDir` variable in `deleteWorkingDirectory` should be renamed to something like `workingDir` so it is not confused with the directory at `mDataDirPath`.

chchwy:
Merged, thank you!! @martinvanzijl

Issue #565, opened 2016-11-10
[Bug] Ghost polyline remains in new canvases


linganesan:
### --Issue Summary--
I guess, this is the follow up of issue #445 
When I create a new canvas/document in pencil2D, it come up with the 'Ghost' polyline, which I was drawn before. Details of this 'Ghost' Polyline was mentioned in the issue #445 

### --Expected Results--
Should be remove the ghost polyline in the new canvas

### --Steps to reproduce--
1.Select the Polyline tool
2.Draw with the Polyline tool in the canvas
3.Without double clicking the left button of the mouse, press '**ESC**' 
4.Now your drawn polyline seems to fixed in the canvas
5.Create a new canvas/document (File -> New)
6.Warning dialog box will be prompt, select 'No'
7.The earlier drawn polyline still appear in the new canvas
8. If you use any other tool except Polyline tool in the new canvas, that Ghost polyline will be stick with the canvas. 
### --Video or Image Reference--

https://www.youtube.com/watch?v=zxI0MO8_4no&feature=youtu.be

### --System Information--
+ Pencil2D Version: Pencil2D-Nightly Build

+ Operating System: Fedora

+ RAM Size: 8GB

+ CPU:  2.40GHz Intel Core i-7

I guess, this behaviour may helps the developers to identify what's going on with this 'Ghost' Polyline scenario. 
---

Jose-Moreno:
@linganesan This hasn't been worked at all. Since we were hoping to have MyPaintLib integrated I'm thinking this isn't as prioritized as other stuff, like export. To be honest I'd rather redo the whole polygon tool borrowing code from vector tools, than to try to fix the one that we have, but We have to depend on a developer for it. If you know someone who is willing to help with that, please let them know Pencil2D needs help! Thanks for the report :smile:

CandyFace:
I've tested this numerous of times while working on #611 and I can't replicate this bug anymore. I'd say it's been fixed. Great work @martinvanzijl

PR #598, opened 2017-02-06
Fix for #565. ESC key now clears polyline.

martinvanzijl:
Fix for #565. Pressing the ESC key now clears the in-progress polyline.


chchwy:
Thanks for your contribution!

Issue #595, opened 2017-02-01
[BUG] Pencil tool pressure sensitivity issue, stroke not drawn with minimal stylus pressure.


Jose-Moreno:
### --Issue Summary--

When using Pencil tool, you can only achieve faint strokes after putting about 45~50% pressure on your stylus.

### --Actual Results

Quite literally when you try to draw with a stylus pressure under 45~50% the Pencil tool will not draw a single pixel. It only begins to count after pushing really hard.

### --Expected Results--

Pressure sensitivity opacity should work from 0 to 100 pressure similar to how the Pen tool works with size only. It should draw a stroke with the faintest of pressure put on the stylus.

### --Video or Image Reference--

![image](https://cloud.githubusercontent.com/assets/4912907/22513819/59accf02-e86b-11e6-91c0-976d53d67539.png)

### --Steps to reproduce--

1. Select Pen tool
2. Check Pressure checkbox []
3. Try to draw a faint line.
4. Press harder until you see a line.

### --System Information--
+ Pencil2D Version:

Pencil2D Nightly Build Nov 2 2016
+ Operating System:

Windows 7 Ultimate
+ RAM Size:

4GB
+ CPU:

+ Graphics Tablet:

Wacom Intuos Pro 5 PTH-450
Driver: 6.3.15-3

---

CandyFace:
This was fixed with #599, so i'm closing this issue.

PR #599, opened 2017-02-07
Removed division from opacity and set feather to 1

### LINKED ISSUE #596: [BUG] Pencil tool stroke should be fully opaque without pressure activated
### --Actual Results

When using the alpha slider on the color wheel, Using fully opaque (255) alpha values make the pencil tool transparent. When dropping to a value of 7 it dissapears

### --Expected Results--

When creating a stroke with the Pencil tool it's opacity should be fully opaque (255 alpha) when the "pressure" checkbox is not activated. Also the alpha interpolation should work so that when it reaches to 0 the stroke is fully transparent and 255 fully opaque.

image

### --Steps to reproduce--

I'll update later. But you just have to draw and change the alpha value to notice this issue. Also turn off the pressure checkbox to test this.

### --System Information--
+ Pencil2D Version:

+ Operating System:

+ RAM Size:

+ CPU:

+ Graphics Tablet:

---

CandyFace:
This should fix #596 and maybe also solve #595

test3

## Changes:
* Removed division from opacity in ScribbleArea
* feather has also been set to 1 in PencilTool, as it seemed to have no influence on the look, except the cursor radius.
* The opacity pressure has been modified to be consistent with how it works in brushTool. I see no reason why the Pencil should act differently.

Is there a a reason as to why the opacity was being divided before drawing the stroke?

### DIFF
--- a/core_lib/interface/scribblearea.cpp
+++ b/core_lib/interface/scribblearea.cpp
@@ -1066,7 +1066,7 @@ void ScribbleArea::drawPen( QPointF thePoint, qreal brushWidth, QColor fillColou
     qreal offset = 64;
 
     QRadialGradient radialGrad( thePoint, 0.5 * brushWidth );
-    setGaussianGradient( radialGrad, fillColour, opacity/2, offset );
+    setGaussianGradient( radialGrad, fillColour, opacity, offset );
 
     QRectF rectangle( thePoint.x() - 0.5 * brushWidth, thePoint.y() - 0.5 * brushWidth, brushWidth, brushWidth );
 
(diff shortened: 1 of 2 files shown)


Jose-Moreno:
@CandyFace Awesome! Thanks a lot Oliver!

chchwy:
Great job! @CandyFace

CandyFace:
@chchwy 
After having tested this some more, it does fix the the alpha problem for the Pencil tool but now it's broken in Brush and Pen. I didn't catch that in my testing because it sorta still works if alpha value is low enough.

I refuse to believe though that dividing the opacity is the correct way do it.

CandyFace:
update: The issue seems to be caused by the strokes being drawn on top of each other, instead of replacing with source only. This means that opacity does act correctly but because the strokes are being multiplied onto each other, you can't see it. This behaviour is odd though, as the composition mode is set to Source in all the tools correctly. Investigating...

Issue #547, opened 2016-10-16
[BUG] Imported sound(s) can only play and be deleted from frame they begin


Jose-Moreno:
### --Actual Results

**Issue A**
A sound will not play from any frame different than the frame where the key frame begins. Apparently the sound layer imports the sound as a long "single" frame. 
**Issue B**
A second issue related to this is that you can't delete the sound "frame" unless you are located at the exact frame where the sound starts.
### --Expected Results--

**Issue A**
We should be able to play a sound from any point in time and any frame in the timeline. It should resume the sound like a normal player instead of getting muted.
**Issue B**
Pencil should be able to delete the sound "frame" wherever the playhead / scrub is located. However I understand this might be a bit more complicated to implement since we would require a sound "scrubbing" functionality like what you see in video editors (note: I'll link here later a feature request for sound scrubbing since it's imperative for lyp sync)
### --Steps to reproduce--

**Issue A**
1. Create new file
2. Create new sound layer
3. Import (short) sound on first frame.
4. Play normally from the first frame and make sure you get sound.
5. Position your scrub / playhead on frame number 2 and hit play. The sound will not play anymore.
**Issue B**
follow previous steps 1 through 4
5. Position your scrub / playhead on frame 2 and press the "remove frame" button
6. The scrub will move 1 frame back but the sound "frame" will remain there
### --System Information--
- Pencil2D Version:
  Nightly Build Oct 10 2016
- Operating System:
  Windows 7 Ultimate 64bits
- RAM Size:
  4 GB
- CPU:
  AMD Phenom II x4 965 BE

chchwy:
This issue has been fixed by @martinvanzijl, I'm going to close it.

PR #600, opened 2017-02-07
Fix for #547. Can play and delete sound from any frame in its length.

martinvanzijl:
Fix for #547. You can now play a sound from any frame in its duration, not just its starting frame.
	
You can also delete and move the sound from any frame.

**Code Changes:**

- Added "mCheckForSoundsHalfway" flag to PlaybackManager. Set to true when starting playback - to check whether to play any sound from part-way through.
- Added "getKeyFrameWhichCovers(pos)" to Layer. This returns the keyframe which covers pos. E.g. a sound clip starting at frame 1 and ending at frame 6 covers frames 1, 2, 3, 4, 5, 6. 
- Updated Layer methods "isFrameSelected(pos)" and "setFrameSelected(pos)" to use "getKeyFrameWhichCovers". I could possibly move this logic change to the LayerSound class and keep Layer as is.
- Added "playFromPosition()" to SoundClip to play sound from part-way through.

### DIFF
--- a/core_lib/managers/playbackmanager.cpp
+++ b/core_lib/managers/playbackmanager.cpp
@@ -64,6 +64,10 @@ void PlaybackManager::play()
 
     mTimer->setInterval( 1000.0f / mFps );
     mTimer->start();
+
+    // Check for any sounds we should start playing part-way through.
+    mCheckForSoundsHalfway = true;
+
     emit playStateChanged(true);
 }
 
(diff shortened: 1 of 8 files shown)


Jose-Moreno:
@martinvanzijl This is fantastic Martin! Hopefully @chchwy Can check it out to see if everything works properly and merge it right away :smile:

chchwy:
Well done, @martinvanzijl

PR #601, opened 2017-02-09
Improve compatibility with older Mac OS X systems

scribblemaniac:
Setting `QMAKE_MACOSX_DEPLOYMENT_TARGET` will enable weak linking, which will improve support for older Mac versions (10.4+ probably). [Qt documentation on the subject](https://doc.qt.io/qt-5/osx-deployment.html#macos-version-dependencies).

### DIFF
--- a/app/app.pro
+++ b/app/app.pro
@@ -101,6 +101,8 @@ macx {
     FILE_ICONS.files = ../icons/mac_pcl_icon.icns ../icons/mac_pclx_icon.icns
     FILE_ICONS.path = Contents/Resources
     QMAKE_BUNDLE_DATA += FILE_ICONS
+
+    QMAKE_MACOSX_DEPLOYMENT_TARGET = 10.3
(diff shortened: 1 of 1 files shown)


scribblemaniac:
Nevermind, I forgot that we're using C++11. This will not work. I really should try compiling things before I commit them 😉

chchwy:
Also Qt officially dropped support for 10.6, so 10.7 would be the minimum requirement.

scribblemaniac:
Oh, where did you see that?

chchwy:
http://doc.qt.io/archives/qt-5.5/osx.html
> Qt 5 applications may be deployed to Mac OS X versions 10.6 "Snow Leopard", but support is limited.

And I also checked the supported platforms of Qt 5.8
http://doc.qt.io/qt-5/supported-platforms.html
minimum OS X version is 10.9

PR #602, opened 2017-02-11
Palette UI changes

CandyFace:
palette

- Palette widget and items take up much less space. We don't want to waste all that space on nothing.
- Added checkerboard background so all colors can be seen. 
- Added a palette preference button 
   - Switch between list mode and a NEW tile mod.
   - Change swatch size (small, medium and large)

Other things that can't be seen in the gif:
Creating a new color in tile mode will not ask about name, the new color will appear unnamed if switched to list mode though. For those who don't care about naming their colors, this should be great.

### Future improvement:   
Colors should be draggable in both list and tile mode.
Double click to add color from color picker or color wheel.


Jose-Moreno:
@CandyFace Awesome contribution Oliver! can't wait to see it on a binary! :smile:

PR #606, opened 2017-02-14
Some tablets are not recognized

CandyFace:
Some tablets are not recognized because TabletEvent only expects a pointerType. This PR adds a NoDevice condition for tablets that are not recognized as a Stylus, likewise PointerType: Cursor is not fail proof either, my tablet was recognized as a Cursor even if it was a pen.

My non Wacom tablet works now, so this should definitely increase compatibility for those who use other tablet brands.


chchwy:
Cool!

Issue #609, opened 2017-02-19
Selection rectangle not shown when dragging up or left from original point


martinvanzijl:
### --Issue Summary--
Using the select tool, if drag up or left from the original point where I clicked, the selection rectangle is not shown. It works fine when I drag down and right.

### --Actual Results
The selection rectangle is only shown when dragging down and right.

### --Expected Results--
The selection rectangle should be shown no matter which direction you drag in.

### --Video or Image Reference--
Video (zipped):
[Selection_Rectangle_Bug.zip](https://github.com/pencil2d/pencil/files/697173/Selection_Rectangle_Bug.zip)

NOTE: The yellow and red circles around the mouse cursor are added by my screen capture software - just to show where I am clicking/releasing the mouse.

### --Steps to reproduce--
1. Open new animation.
2. Choose the select tool.
3. Click and drag up and left on the canvas. The selection rectangle will not be shown.

### --System Information--
+ Pencil2D Version:
Nightly Build Dec 16 2016
(Still occurs in Feb 13 Nightly Build)

+ Operating System:
Windows 7 64-bit

+ RAM Size:
4 GB

+ CPU:
Intel i5-2410M @ 2.30 GHz

+ Graphics Tablet:
N/A - Was using the mouse.

### --Code Snippets--
Looks like the code below needs changing. I've been developing a fix, and I'm hoping to make a pull request soon.

```C++
        case ScribbleArea::NONE:
            mScribbleArea->mySelection.setBottomRight( getCurrentPoint() );
            break;
```

CandyFace:
Tested, it works great. Good job :)

CandyFace:
As mentioned, tested and works. Closing.

PR #610, opened 2017-02-19
Fix for #609 - select rectangle drawn in any direction

martinvanzijl:
Fix for #609. The selection rectangle is now drawn in any direction.

Video demo:
(https://github.com/pencil2d/pencil/files/786061/Selection_Rectangle_Bug_Fix.zip)

**Code Changes:**
I store the original point clicked in selecttool.cpp and use it to calculate the selection rectangle. I used a (static) global variable for this; I can try making it a class field instead if you prefer.


Jose-Moreno:
@martinvanzijl Thanks a lot for your contribution Martin! It's incredible! :smile:

core_lib/tool/selecttool.cpp
```diff
@@ -5,7 +5,9 @@
 #include "toolmanager.h"
 #include "selecttool.h"
 
-QPointF gSelectionOrigin;
+// Store selection origin so we can calculate
+// the selection rectangle in mousePressEvent.
+static QPointF gSelectionOrigin;
```
CandyFace: Why a global variable?
Couldn't you just store the variable as a private in the header or in the function itself.

Global variables should be avoided if possible imo.

chchwy: Agree with @CandyFace, this is not necessary. However I'll merge the code anyway and refactor later.

Issue #486, opened 2016-02-19
[BUG] Drawing Tools Stroke Latency (Lag) & Interpolation Issue


Jose-Moreno:
**--Issue Overview--**

There are two related problems that happen in both vector & bitmap layers. 

First the stroke latency: basically every drawn stroke lags behind the cursor or pointer. 
Second, the Stroke interpolation: Speed is a factor that modifies how the lines are interpolated, thus giving it enough precision to mimic the stroke that was input from a graphic tablet.

Both problems become altered when the Pencil2D window layout, and thus the canvas size are reduced as shown in the video demo. 

It's important to say that in previous version 0.5.4b both **Stroke Latency** & **Stroke Interpolation** were present but it was to such a minor degree that they were barely noticeable for the regular user.

**--Video Demo--**
https://www.youtube.com/watch?v=9pTepQSgLCQ&feature=youtu.be

**--Steps to reproduce--**

**Stroke Latency & Stroke Interpolation**
1) Open Pencil2D
2) Select a drawing tool (i.e Pencil)
3) Select either a Vector or a Bitmap Layer
4) Try to draw as "fast" as a normal stroke with pencil and paper would feel like
5) Notice both the stroke lagging behind the cursor and the poor interpolation
6) Reduce the Pencil2D Window Layout to Half it's size (make it to occupy half the screen)
7) Try the same kind of strokes as in step 5.
8) Notice the reduced latency and interpolation problems.

**--Expected Results--**
**Stroke Accuracy** ( _Latency & Interpolation_ ) should be independent of:
- The Window layout size
- The Canvas Size
- The Document / Camera resolution

Purely from a "User Experience" standpoint the stroke drawing interpolation method should not be apparent to the user, or it can throw off their drawing perception with the program and limit their freedom while creating artwork.

**--System Information--**
Pencil2D Nightly Build 15 Jan 2016 / Nightly Build 16 Feb 2016 (Run & built with Qt Creator 3.4.2 based on Qt 5.5.0)
Windows 7 64bits SP1
ATI Radeon HD 6770 1GB
AMD Phenom II X4 965 BE 3.4 GHZ
4 GB RAM

feeef:
Running Pencil2D from QTCreator, you may have latency problems comming from debug data being processed all the time. Especially debug messages. You should try to run the same test with a release binary, away from the IDE. Wait for the next nightly build.

Jose-Moreno:
[...]

PR #611, opened 2017-02-20
Brush/Tools changes, AA Brush setting and more..

CandyFace:
## What's fixed/changed:

- Opacity strokes being drawn on top of each other instead of being continuous
for Pen and Brush.
- Separated painting/brush Anti-Aliasing from the Preferences
Anti-aliasing setting.
- Set fixed step size for pen. In the future I want to control this via a slider, but it should be hidden in a popup window, so for now it's fixed instead of relying on the feather value.

## What's removed:
- Feather from Pen

## What's added:
- Checkbox to enable AA for brushes, this is used for Pen only. AA is
redundant for Brush and Pencil when feather is enabled and is otherwise
enabled.
- Consistent dap ability across Pencil, Brush and Pen. Before this was only possible with brush.
- Dotted Cursor for Pen tool.

brush

### DIFF
--- a/core_lib/interface/scribblearea.cpp
+++ b/core_lib/interface/scribblearea.cpp
@@ -1165,95 +1165,16 @@ void ScribbleArea::liquifyBrush( BitmapImage *bmiSource_, QPointF srcPoint_, QPo
     delete bmiTmpClip;
 }
 
-void ScribbleArea::drawPolyline( QList<QPointF> points, QPointF endPoint )
+void ScribbleArea::drawPolyline(QPainterPath path, QPen pen, bool useAA)
 {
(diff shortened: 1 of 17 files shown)


CandyFace:
## What's been fixed:
the tool lagging behind cursor problem #486, I figured that the problem was something universal, since it happened to every tool. I found  the smoothing pos variable inside the strokemanager and simply removed it. I see no reason why the stroke should be smoothed, it should always be where the cursor is located. If anything, this should only happen when using a tablet but even so, the smoothing was way to strong.

The polyline tool has been fixed too, no more artifacts from the stroke.

## What's been added:
- AA checkbox for Polyline tool.

PR #618, opened 2017-02-28
Added progress dialog (with cancel option) for exporting image sequences

martinvanzijl:
Added a progress dialog for exporting image sequences. Shows how far through the export is, and you can stop the export by clicking "Abort".  Below is a video demo:

[Progressbar Demo.zip](https://github.com/pencil2d/pencil/files/805477/Progressbar.Demo.zip)

The reason for this is:
1. When exporting image sequences, I couldn't tell how far Pencil was through, so I just had to wait till it became responsive again. This sometimes took a few minutes for very long animations.
2. There was no way to cancel the export, so I would either have to kill the application or just wait till the export eventually finished.

### DIFF
--- a/app/mainwindow2.cpp
+++ b/app/mainwindow2.cpp
@@ -782,6 +782,13 @@ void MainWindow2::exportImageSequence()
 
 
     int projectLength = mEditor->layers()->projectLength();
+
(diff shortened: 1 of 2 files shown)


chchwy:
@martinvanzijl Thanks for your contribution :)

Issue #620, opened 2017-03-02
[Request] Allow to import image sequences with adjustable frame spacing


Jose-Moreno:
### --Issue Summary--

Right now importing an image sequence of any length **will import each image for every 2 frames**. This means it is imported on two's. This is useful on some cases when bringing frames drawn by hand, however the user should have the ability to choose if they want to import it on one's, two's, three's or a custom frame spacing between images.

### --Expected Results--

There should be a window dialog that allows the user to specify the amount of separation between each image. We could use a window dialog similar to the one used when exporting, I think those are made with QT widgets.

Edit: The caption could say something like "Import an image every **[X]** frames"

![image](https://cloud.githubusercontent.com/assets/4912907/23493645/4ce7acac-fedb-11e6-9d2b-76ecc6f8be9a.png)

### --Video or Image Reference--

![image](https://cloud.githubusercontent.com/assets/4912907/23493257/6cd934b6-fed8-11e6-9948-1f2e9d9ba4cb.png)

### --Steps to reproduce--

1) Go to File > Import > Image Sequence.
2) Choose your images
3) You'll get an even amount of frames similar to the image reference above.

### --System Information--
+ Pencil2D Version:

Pencil2D Nightly Build November 2 2016

+ Operating System:

Windows 7

CandyFace:
Sounds like a good request, I'm going to work on this :)

Jose-Moreno:
@CandyFace Thanks a lot! Other animators had asked me for this but I was trying to put up a decent request before submitting.

CandyFace:
![test](https://cloud.githubusercontent.com/assets/1045397/23554591/139cbb58-0026-11e7-922a-6952329c320d.gif)

Alright should be done now. PR to be comitted soon :)

Jose-Moreno:
@CandyFace Awesome! also that hand icon looks quite great!

CandyFace:
I'm closing this since it has been implemented.

PR #621, opened 2017-03-03
Added dialog to adjust frame spacing for imported image sequence

CandyFace:
test

PR for feature request #620

### DIFF
--- a/app/app.pro
+++ b/app/app.pro
@@ -49,7 +49,8 @@ HEADERS += \
     pencilapplication.h \
     exportmoviedialog.h \
     app_util.h \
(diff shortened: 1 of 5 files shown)


scribblemaniac:
For your screencap, you choose to import a frame every 5 frames, but it looks to me like the final result is every 6 frames. I think either the dialog box needs to be reworded to something like "# of spaces between every imported frame", or the input box needs to be set to a minimum of 1 and you change the code to move one less time (because the inserted frame counts as one frame).

CandyFace:
Ah .. yeah you are correct. I've adjusted it now :)

PR #623, opened 2017-03-09
fix typo s,sufficent,sufficient,g

mapreri:
I'm actually not sure what to do with the translations, that contains that typo.  will they be updated automatically from onesky?


mapreri:
Can you also update the translations?  (no idea how that is done)

chchwy:
Done. I do it manually at the moment.
- Run lupdate to generate a new pencil.ts, upload it to OneSky.
- Download ts files from OneSky and overwrite them in repo.

PR #624, opened 2017-03-17
Refactor title save indicator and add file icon to main window title

scribblemaniac:
Instead of manually prepending the asterisk to the window title to indicate file changes, this lets QT handle it. On Window and Linux I don't think there will be any noticeable difference from this change, but on Mac, the close button will change when the current file has unsaved changes. I have also added a file icon to the window title (again probably mac only), which will be greyed out slightly when there are unsaved changes, as is typical of Mac applications.

### DIFF
--- a/app/mainwindow2.cpp
+++ b/app/mainwindow2.cpp
@@ -524,7 +524,7 @@ bool MainWindow2::openObject( QString strFilePath )
     mRecentFileMenu->addRecentFile( object->filePath() );
     mRecentFileMenu->saveToDisk();
 
-    setWindowTitle( object->filePath() );
+    setWindowTitle( object->filePath().append("[*]") );
     setWindowModified( false );
     setWindowIcon( QFileIconProvider().icon(strFilePath) );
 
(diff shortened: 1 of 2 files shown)


chchwy:
Thanks @scribblemaniac, I agreed that let Qt handle the save status is better.
And as I checked, close button & file icon changes are not available on Windows, so we might need to find a way to let Win/Linux users know the file has been saved or not (like an asterisk).

scribblemaniac:
> And as I checked, close button & file icon changes are not available on Windows, so we might need to find a way to let Win/Linux users know the file has been saved or not (like an asterisk).

That stuff is indeed just for mac. On windows/linux, qt should add the asterisk automatically. That is based off of the docs, I haven't tested it yet, but this code should theoretically handle this already.

chchwy:
Could you give a link for the documentation? I didn't see the asterisk.
I'll be happy if we don't need to deal with it.

scribblemaniac:
Oops, seems like I need to read it more carefully. I pushed a new commit that should hopefully make it show up now. The documentation is here:
https://doc.qt.io/qt-5/qwidget.html#windowModified-prop

chchwy:
Did a quick test, once the asterisk shows, the asterisk will never disappear even after saving file.

scribblemaniac:
> Did a quick test, once the asterisk shows, the asterisk will never disappear even after saving file.

Hmmm, okay I'll look into it again

scribblemaniac:
What a pain, sorry for this. The asterisk should show *and* disappear now.

chchwy:
Now it works very well. Thanks @scribblemaniac

PR #625, opened 2017-03-22
Fix syntax

mapreri:
a379816fefaa5bb9ecd7e313b701d19bb2d4e3e2 added pencil.ts to TRANSLATIONS, but doing so it broke the syntax of the list.


chchwy:
oh...my bad. thanks mate.

PR #626, opened 2017-03-24
Updated Hungarian translation

ptrg:
Also other translations changed here because of the regeneration of translations...


chchwy:
Well done! Thanks @ptrg

PR #617, opened 2017-02-26
Add script for uploading nightly builds to Google Drive

J5lx:
As per discussion in #582. This script will allow automatic nightly builds to be uploaded to GDrive via CI. It requires Python 3 and the [google-api-python-client](https://pypi.python.org/pypi/google-api-python-client) Python package. (I originally tried to do it with simpler means and no additional dependencies but the amount of complexity in Google’s authorization API is honestly quite ridiculous, so I gave up.)

Obtaining the right API credentials and setting everything up is a pretty tricky too, but I’ll make a detailed guide on that tomorrow.

PR #627, opened 2017-04-02
make travis ready for google drive integration

CandyFace:
I've tested everything aside from the upload process to google drive, since the private key is not set up for my account. if @chchwy has made the preparations mentioned in #617, my assumption is that it'll work when the PR has been merged.

The changes to the Travis file should in the end upload two files to google drive, one for linux and another for mac. The zip files are named like this: pencil2d-os-date where os is linux or mac and date is year-month-day

As i've previously said, Travis-CI doesn't support Windows yet, so at some point i'll see if I can get it setup on Appveyor too but that's a task for another day.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,7 +74,7 @@ after_success:
       zip -r "pencil2d-linux-$(date +"%Y-%m-%d").zip" app/;
      fi'
   - 'if [ "$TRAVIS_OS_NAME" == "osx" ]; then
-      zip -r "pencil2d-osx-$(date +"%Y-%m-%d").zip" app/;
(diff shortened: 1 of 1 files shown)


chchwy:
Thanks @CandyFace, testing right now.

CandyFace:
Hmm... console spits out 
"The previous command failed, possibly due to a malformed secure environment variable.
Please be sure to escape special characters such as ' ' and '$'.
For more information, see https://docs.travis-ci.com/user/encryption-keys."

I wonder if it's going to work... did you setup the key correctly?

chchwy:
Looking at the issue, probably I didn't escape all special characters properly.

CandyFace:
Baah.. I too found a typo in the script.. fixing.. get ready for PR

PR #630, opened 2017-04-06
remove makefile and fix for broken OSX build

CandyFace:
Hopefully this fixes the OS X build. It's untested.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,7 +74,7 @@ after_success:
       cd app;
       echo "Removing Makefile";
       rm Makefile;
-      cd..;
+      cd ..;
       echo "Zipping...";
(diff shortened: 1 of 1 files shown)


chchwy:
Thanks @CandyFace that's great.
Is the `macdeployqt` working with Qt 5.7? I don't have a mac at the moment so I can't check it.

CandyFace:
Yes and no, here's why: my observations has showed me that after you've qmake and make, the build are missing some name changes in its info.plist file, this essentially means that the package is corrupt and macdeployqt therefore can't locate it's binary.

I'm trying to figure out what qt creator does that the CLI command doesn't do that makes this happen.

PR #633, opened 2017-04-07
Fix Linux upload

J5lx:
In the recent CI build I noticed that right now the Travis script still tries to upload the ZIP file on Linux even though we switched to tar.gz.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -93,7 +93,7 @@ after_success:
       wget -P Pencil2D.app/Contents/MacOS/plugins https://evermeet.cx/ffmpeg/ffmpeg-3.2.4.7z;
       7z x Pencil2D.app/Contents/MacOS/plugins/ffmpeg-3.2.4.7z -o"Pencil2D.app/Contents/MacOS/plugins";
       rm Pencil2D.app/Contents/MacOs/plugins/ffmpeg-3.2.4.7z;
-      
+
       echo "Copying necessary Qt frameworks";
       macdeployqt Pencil2D.app;
       echo "Removing Makefile";
@@ -110,7 +110,7 @@ after_success:
   - cd "$TRAVIS_BUILD_DIR/util"
 
   - 'if [ "$TRAVIS_OS_NAME" == "linux" ]; then
-    python nightly-build-upload.py 0BxdcdOiOmg-CcU1WOFpCOFBvVXc "$TRAVIS_BUILD_DIR/build/pencil2d-linux-$(date +"%Y-%m-%d").zip";
+    python nightly-build-upload.py 0BxdcdOiOmg-CcU1WOFpCOFBvVXc "$TRAVIS_BUILD_DIR/build/pencil2d-linux-$(date +"%Y-%m-%d").tar.gz";
   fi'
   - 'if [ "$TRAVIS_OS_NAME" == "osx" ]; then
     python nightly-build-upload.py 0BxdcdOiOmg-CeVpTY294cXdLZ2c "$TRAVIS_BUILD_DIR/build/pencil2d-mac-$(date +"%Y-%m-%d").zip";


Jose-Moreno:
@J5lx Hey Jakob, sorry to intrude, I was trying to download a nightly build from the Google Drive for Linux to see If I could run it on a laptop I got, but I noticed almost every tar.gz build is empty. Do you know what could be happening? Or is this something that is still on the works?

https://drive.google.com/drive/folders/0BxdcdOiOmg-CcU1WOFpCOFBvVXc

J5lx:
Yeah, I also noticed that recently but I haven't had time to look into it yet. I'll see what I can do about it.

J5lx:
I think I found the culprit: 685a0d7729543834db607ee63957eef5aaa4b9c0 changed the output directory of the binary from app/ to bin/ without updating the Travis scripts. For the record, this affects mac builds as well. I’ll submit a PR.

J5lx:
Alright, PR is up. For your Linux testing venture you can use the build I generated just now: <https://drive.google.com/open?id=0BzrRR4_F38j0NzdhRUI5YTB5d00>

Jose-Moreno:
@J5lx Hey, sorry for the delay. I'm testing the file you linked on Linux Mint 17.2 'Rafaela' Xfce 64-bit, I downloaded and extracted the tarball. I only get one file branded as an "executable", however when I try to run it nothing happens. I tried running it through the software installer, through the terminal but that's it, it doesn't do anything.

I conversely tried one of the debian maintained packages on their "unstable" catalogue here, which is about the same date (April 19 2017): https://packages.debian.org/sid/pencil2d

But I ran into the beauty of trying to install dependencies via terminal (after trying out with the package installer several times) and since it's a debian based distro, there's a lot of issues a normal user runs into to be able to install any kind of "unstable" software, which means I couldn't run any nightly build for Pencil2D on my Linux distro.

Anyway, is there anything I can do to test if the Linux builds work? I mean, do they work for you? After you download them from the google drive I mean. Thanks in advance for any input or advice you have.

J5lx:
Yeah they work just fine for me. Most likely it’s in fact some dependency issue.

So, firstly you are right that the file in the tarball is just an executable, it’s not any kind of installable package. To see what’s going wrong, is there any kind of output when you run that executable from a terminal? What is the exit code (`$?`)? Also, the output of `ldd /path/to/pencil2d-linux-2017-05-14/Pencil2D` could contain some hints if libraries are missing.

I’ll also see if I can set up a VM with that Mint release to look around myself. Right now the Linux builds are in fact quite barebones compared to OS X and Windose so I’ll try and look into some more polished solutions (e.g. AppImages).

J5lx:
Your problem is likely caused by missing libqt5multimedia5. I also created a PR for switching Linux builds to the AppImage format which should help avoid such issues, see #671.

Jose-Moreno:
@J5lx Thanks for your prompt response. Yeah while using the terminal on the uncompressed folder I tried `sudo install Pencil2D /home/myname/pencil2D/` just now and it copied the same file over that newly created folder. When I used the `ldd `command it said: `not regular file`. Not even asking for dependencies or giving a more detailed error. I'm sure I was doing something wrong, but we also have to consider that not all people using linux might be power users.

Now trying the appimage you linked in your PR, after changing the permissions (and allowing it to "run as a program") It worked! My laptop is really old, so this is fantastic news indeed.

Generally It's very difficult for me to deal with MAC or Linux user problems since I don't know the proper procedures to install on those systems. Now If someone asks I can also give them broader support options before passing their issues directly through the tracker. Great job and thank you for your assistance :smile:

mapreri:
> I conversely tried one of the debian maintained packages on their "unstable" catalogue here, which is about the same date (April 19 2017): https://packages.debian.org/sid/pencil2d
> 
> But I ran into the beauty of trying to install dependencies via terminal (after trying out with the package installer several times) and since it's a debian based distro, there's a lot of issues a normal user runs into to be able to install any kind of "unstable" software, which means I couldn't run any nightly build for Pencil2D on my Linux distro.

OOI, what would be the actual problem?  I'd expect that if you are running a recent enough debian-based distribution a simple `apt install ./file.deb` to be enough.  ("recent enough" as it'd need the dependency as listed in the package, so a new enough libc6, a new enough libstdc++6, etc).

Jose-Moreno:
@mapreri Well, basically as I understand you shouldn't run "unstable" stuff on "stable" versions of Debian. conversely It was a bit of a pain trying to get that to work. I haven't used linux for a long time so I'm not really great with it, I did go and search how to do it, but the Debian guide on how to install "Unstable" gave some hints on how to approach that, despite the "if you break it it's your fault" tone. 

At first I did try the "deb" file from this link:
http://ftp.us.debian.org/debian/pool/main/p/pencil2d/pencil2d_0.5.4+git20170419.616acfc+dfsg-1_amd64.deb
But it didn't work for me, which was wierd. It didn't bring me any kind of notable error, it did the same as the other executable, nothing. I tried enabling permissions, I tried running it with the software manager. Then I finally indulged and went with the terminal. It didn't do anything.

Then I tried the hard-mode route, and went on trying to enable the sources from pencil2D, but it still wouldn't do anything. Tried the `apt-get install unstable pencil2d` and other similar stuff. Didn't work. Tried using aptitude, no luck. Tried with the GUI Package manager for Mint, it kept alerting me about a badly formatted source file which I changed everytime by editing the /source.list file but in the end, I wasted 2 hours trying to do it. Mortal beings shouldn't need that much time to setup a non-canonical piece of software.

I'll try at another time to work out the debian sources again, but that's what happened in my particular case.

Issue #550, opened 2016-10-16
Improve support for dark themes


J5lx:
> **Note:** This issue is about general problems with existing dark theme support on macOS and Linux. For macOS-specific issues, see #1229. For dark mode support on Windows, see [QTBUG-72028](https://bugreports.qt.io/browse/QTBUG-72028).

When using Pencil with a dark theme (in this case [Adapta-Nokto](https://github.com/adapta-project/adapta-gtk-theme) via the gtk style of Qt), there currently are some components that don't adapt properly. Most of this was originally mentioned in https://github.com/pencil2d/pencil/issues/540#issuecomment-253976402 and then simply extracted into this separate issue upon suggestion of @scribblemaniac.

![Pencil with the Adapta-Nokto theme](https://cloud.githubusercontent.com/assets/2063777/19416395/791b87b0-938f-11e6-8a48-d1160efd35ee.png)

These are the issues that I see there right now, in no particular order:
- [x] The main menu bar looks like it has some custom styling overwriting the theme that should be removed, e.g. the text colour is set according to the theme but the background color clearly isn't. Submenus are not affected and look just fine. See #634.
- [x] The "Display" pane apparently uses some kind of custom-made (and therefore unthemed) buttons, which probably should be exchanged with Qt's standard buttons. See #634.
- [x] The contrast in the system information section of the about dialog ranges from bad to completely nonexistent with dark themes, depending on the exact theme. See #1229 for a screenshot.
- [x] The timeline needs support for theming; currently it doesn't adapt at all. Will probably be solved together with #533
- [ ] Some icons contain fragments of white where they transition from colour to transparency, indicating that they were originally designed for light backgrounds. These icons should be edited or remade accordingly. Obvious examples:
  - The move tool
  - The brush tool
  - The playback controls
  - The "Duplicate Frame" button
  - The zoom entries in the View menu
  
  In some other icons those fragments are also present, but barely even noticeable.

  See #603 and #1356.
- [ ] Some icons don't contrast well with the dark background and should be redesigned accordingly. Obvious examples:
  - The select tool
  - The restore and close buttons on each pane

  See #603 and #1356.

Some additional remarks:
- The icon-related might be solved by switching to a monochrome icon theme, which is what most other programs explicitly supporting both light and dark themes do (e.g. MyPaint, Krita, MuseScore). These programs show light icons when using a dark theme and dark icons when using a light theme
[...]

PR #634, opened 2017-04-07
Remove custom styling

J5lx:
This removes custom styling of the menu bar and the display option buttons, which makes Pencil play along much more nicely with system themes (i.e. #550). Before:

Before

After:

After

### DIFF
--- a/app/ui/displayoption.ui
+++ b/app/ui/displayoption.ui
@@ -34,20 +34,6 @@
      <property name="toolTip">
       <string>Horizontal flip</string>
      </property>
-     <property name="styleSheet">
-      <string notr="true">QToolButton { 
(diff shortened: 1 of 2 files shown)


CandyFace:
Eh weird about those &amp;ersands, can't you remove them? or at least make it consistent across all menu items at the first position?

Good job though :)

J5lx:
I’m still looking into it. I think I’m starting to see a pattern behind what Qt Creator did there, but I’m not quite sure yet.

J5lx:
I give up. I rebased the whole thing to include only the relevant stuff, performing the changes in an external editor. When I do something, I want to do it right, and what Qt Creator did there seems wronger the longer I look at it. Ready to merge now.

PR #641, opened 2017-04-13
Hand tool improvements

scribblemaniac:
Just a couple of improvements for the hand tool. It now uses the native Qt::OpenHandCursor, and changes to Qt::ClosedHandCursor while dragging. Rotation behavior with alt click has been basically rewritten. I honestly am not even sure if it was functioning before, because it just more or less seemed to go wherever. I think you will agree that the new behavior is much more intuitive when you try it.

### DIFF
--- a/core_lib/tool/handtool.cpp
+++ b/core_lib/tool/handtool.cpp
@@ -25,13 +25,15 @@ void HandTool::loadSettings()
 
 QCursor HandTool::cursor()
 {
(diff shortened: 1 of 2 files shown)


CandyFace:
It did somewhat work before, it rotated when you held down alt and moved the mouse, however only clockwise and it didn't matter how you moved the mouse.

Looking forward to try this.

PR #649, opened 2017-04-18
Specify Nightly target directories as env vars, AppVeyor ver.

J5lx:
Okay, here goes the AppVeyor edition of #648. The env vars you need to set in AppVeyor for this repo are:
- `WIN32_NIGHTLY_PARENT` = `0BxdcdOiOmg-CcUEwS1R0WFhwM0E`
- `WIN64_NIGHTLY_PARENT` = `0BxdcdOiOmg-CSVlqc3JNQV9hVGs`

Issue #530, opened 2016-09-02
please allow using system zlib


mapreri:
since when I started packaging pencil2d for Debian, I'm carrying [a local patch](https://anonscm.debian.org/git/collab-maint/pencil2d.git/tree/debian/patches/remove-zlib.patch?h=debian/0.5.4_git20160902.50c5ddb%2bdfsg-1) to use the system zlib (I remove the whole 3rdparty/zlib from the tarball).

Now, that patch was a quick thing that just removes everything zlib-related and adds a `LIBS += -lz`, nothing clever.

I'm really dumb when it comes to qmake, so to me it's going to take a good while to come up with something usable also in other platforms and that falls back sanely to the bundled copy.  I wonder if somebody can come up with something here :)

Thanks in advance!

Jose-Moreno:
@mapreri I really wouldn't know where to start either to be honest. Maybe @chchwy or @feeef know any better? If anyone else is using Linux distros, please respond.

scribblemaniac:
Your patch works fine on macs since zlib is installed on the system. I would do something like this though to support the homebrew version of zlib:

```
unix: LIBS += -lz
macx {
  LIBS += -L/usr/local/opt/zlib/lib -lz
  INCLUDEPATH += /usr/local/opt/zlib/include
}
```

I have no idea how it would be setup for windows, so someone else will have to help you there.

mapreri:
@scribblemaniac `/usr/local/opt/zlib/lib` sounds a bit weird to me: it's /usr/local, which according to the FHS is a place where system administrator puts their own stuff, according to whatever logic they like most (except the usual bin, etc, share,...).
Then, is that present on all system?  IMHO the build system should check for the presence of those libraries; for example in linux I'd try to use pkg-config with something like
```qmake
CONFIG += link_pkgconfig
findPackage(zlib){
    LIBS += -lz
}
```

scribblemaniac:
@mapreri It is not present on all systems. On a clean installation, zlib is in `/usr/lib` I think, which will already be in the search path. Adding `/usr/local/opt/zlib/lib` is so that the the Homebrew (a popular package manager for mac, which uses `/usr/local`) version will be used if possible. The advantage of that version over they system version is that it is updated more often.

chchwy:
No system zlib on Windows, and that's why zlib is included in pencil respository. 
If it is more common to use system zlib on linux, we could make zlib as a windows only build option.

J5lx:
[...]

PR #651, opened 2017-04-19
Fix appveyor.

CandyFace:
After much trial and error, I finally figured a combination of quotes and backslashes that works on windows.

### DIFF
--- a/app/aboutdialog.cpp
+++ b/app/aboutdialog.cpp
@@ -44,13 +44,18 @@ void AboutDialog::init()
     lay->addWidget(logoText);
     lay->addWidget(aboutText);
 
+#define STRINGIFY(x) #x
+#define TOSTRING(x) STRINGIFY(x)
+#define S__GIT_TIMESTAMP__ TOSTRING(GIT_TIMESTAMP)
+#define S__GIT_COMMIT_HASH__ TOSTRING(GIT_CURRENT_SHA1)
+
 #ifdef GIT_EXISTS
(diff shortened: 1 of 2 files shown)


J5lx:
Well, *compiles* on Windows, not *works* on Windows… Everything else would have been too good to be true, I guess.

The about dialog without a git timestamp

(Note that while the screenshot was taken with Wine I’ve also confirmed the missing timestamp in Windows 10)

At this point I haven’t looked into the cause, though. FWIW, the [port](https://github.com/J5lx/pencil/tree/meson) to the [meson build system](http://mesonbuild.org/) I originally mentioned in https://github.com/pencil2d/pencil/issues/530#issuecomment-293109549 works splendidly, see the 2017-04-18 builds at <https://drive.google.com/open?id=0ByqfJUKSPYQtZHBzTjllVUpreUU> (the 2017-04-19 ones are based on this PR, so they do not work).

CandyFace:
Hahahaha pff.. it really was too good to be true. I can't test whether it shows the text or not since I don't have a windows nearby with VS.

J5lx:
For testing you can use [Wine](https://winehq.org/) or a VM with a [Windows Insider build](https://insider.windows.com/), both are free (at least as in free beer). That’s what I use when I desperately need to hack on Windose stuff. You can also use #649 to extract your own AppVeyor builds.

CandyFace:
I would have done if I had enough space left. I have Wine though but not enough space to get  

I've found an elegant solution though, using stringification.

J5lx:
Can confirm it works. Great Job!

PR #652, opened 2017-04-19
fix #555 replace special characters with '?'

### LINKED ISSUE #555: [BUG] File corruption when using certain characters as layer names
### --Issue Summary--
When I use either [U+FFFE](http://unicode-table.com/en/FFFE/) or [U+FFFF](http://unicode-table.com/en/FFFF/) characters as a layer name, it saves fine. However when I try to open the saved file Pencil cannot read the file.

### --Expected Results--
Either files containing this characters should be loaded correctly by Pencil, or people should not be allowed to input them when entering the layer name.

### --Steps to reproduce--
1. Click one of the links in the issues summary
2. Copy the symbol
3. Open a new Pencil file
4. Edit a layer name and paste the character into it
5. Save the file
6. Try and reopen the saved file

### --System Information--
+ Pencil2D Version:
Built from source (20891445508a9e238f67fd3491ec8a409684f9fa)

+ Operating System:
macOS Sierra

CandyFace:
This PR fixes #555 which occurred when special characters were inserted into layer labels, causing your save to become corrupted.

It looks through all ascii characters, if the inserted character is above, then replace with '?'

### DIFF
--- a/core_lib/structure/layer.cpp
+++ b/core_lib/structure/layer.cpp
@@ -474,11 +474,23 @@ void Layer::mouseRelease( QMouseEvent* event, int frameNumber )
 void Layer::editProperties()
 {
     bool ok;
+    QString replace;
     QString text = QInputDialog::getText( NULL, tr( "Layer Properties" ),
                                           tr( "Layer name:" ), QLineEdit::Normal,
                                           mName, &ok );
     if ( ok && !text.isEmpty() )
     {
+        for (int i = 0; i < text.size(); ++i)
+        {
+            QChar check = text.at(i);
+            unsigned char c = *(unsigned char*)(&check);
+            if (c >= 127) {
+                replace.append("?");
+            } else if (QChar(check).isPrint()) {
+                replace.append(QChar(c));
+            }
+        }
+        text = replace;
         mName = text;
         setUpdated();
     }
--- a/core_lib/structure/layer.cpp
+++ b/core_lib/structure/layer.cpp
@@ -473,10 +473,6 @@ void Layer::mouseRelease( QMouseEvent* event, int frameNumber )
 
 void Layer::editProperties()
 {
-    QStringList illegalChars;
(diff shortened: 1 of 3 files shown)


CandyFace:
Although this may not be the best solution since it removes all non ascii and latin characters.

scribblemaniac:
I think replacing all non-ascii characters is definitely overkill and not the correct way to solve this issue. It makes pencil practically useless for anyone who uses a non-latin alphabet (cyrillic, kanji, hangul, etc.).

chchwy:
We cannot do in this way. QChar is a 16 bits unicode character, cast it to a unsigned char is definitely wrong.

CandyFace:
Alright i've scratching that, working on a better solution.

CandyFace:
Pff way to go overcomplicating the task... after having scratched my head for a while I realized, I can't see those characters at all, so no matter what I do, the parser will never accept those symbols. 

This new implementation checks the string using regex whether it contains any special characters from U+FFEF to U+FFFF, if it does, then it replaces those with nothing.

core_lib/structure/layer.cpp
```diff
@@ -473,24 +473,25 @@ void Layer::mouseRelease( QMouseEvent* event, int frameNumber )
...
+
     bool ok;
-    QString replace;
     QString text = QInputDialog::getText( NULL, tr( "Layer Properties" ),
                                           tr( "Layer name:" ), QLineEdit::Normal,
                                           mName, &ok );
     if ( ok && !text.isEmpty() )
     {
-        for (int i = 0; i < text.size(); ++i)
+        if (text.contains(regex))
```
scribblemaniac: Any reason why you don't just do `text.replace(regex, "")`?

CandyFace: Hmm now that you mention it, I guess not. I did it this way because initially the regex was super greedy and I didn't want it to always do unnecessary computation.

chchwy:
Confirmed it works with CJK characters.

Issue #129, opened 2013-07-30
EDITED: DeleteCurrentLayer shortcut under PreferencesPanel doesnt work


ghost:
Last build 30July
Win7 64bits

Assigning a shortcut keyboard for the feature "DeleteCurrentLayer" doesn't work and the letter shortcut doesn't appear on the Main Menu-> Layer, after having customized it.

But deleting a Layer does work if going directly to the Main Menu bar--> Layer --> DeleteCurrentLayer

Jose-Moreno:
@feeef @tasaif @chchwy This bug still exists in ** Nightly Build Jan 15 2016**

CandyFace:
Fixed in PR #658.

Closing.

PR #658, opened 2017-04-26
#129 fix delete current layer shortcut not working

CandyFace:
FIxed deleting current layer shortcut not working, related to #129


chchwy:
Thanks @CandyFace

Issue #552, opened 2016-10-17
[BUG] Layer transparency while camera/sound layer is selected


scribblemaniac:
### --Issue Summary--

Upon selecting the camera or sound layer, all other layers have a slight transparency, but this is not desirable behavior.
### --Actual Results

When the camera layer or a sound layer is active, all layers get a slight transparency applied to them.
### --Expected Results--

The expected behavior when the camera or sound layer is selected is no transparency or filtering of any kind within the render area. It should display _exactly_ what the camera would render (esp. for the camera layer). This transparency is normally applied to all other layers so you can focus on the strokes in the selected layer, which makes sense for bitmap and vector layers but not for sound or camera layers. When the user selects the camera they are conveying that they want to focus on what the camera would render, which is all of the layers not none of the layers.
### --Video or Image Reference--

|  | Current Behavior | Expected Behavior |
| --- | --- | --- |
| **Viewport<br>(Camera/<br>sound<br>layer<br>selected)** | ![viewport screenshot](https://cloud.githubusercontent.com/assets/3461051/19427197/b2952fd0-93ff-11e6-8406-cf75be10e9db.png)Red value: #FF3936 | ![rendered result](https://cloud.githubusercontent.com/assets/3461051/19427218/cc1df78e-93ff-11e6-974f-e90feaaef058.png)Red value: #FF0000 |
| **Render** | ![rendered result](https://cloud.githubusercontent.com/assets/3461051/19427218/cc1df78e-93ff-11e6-974f-e90feaaef058.png)Red value: #FF0000 | ![rendered result](https://cloud.githubusercontent.com/assets/3461051/19427218/cc1df78e-93ff-11e6-974f-e90feaaef058.png)Red value: #FF0000 |
### --Steps to reproduce--
1. Open a new project
2. Draw or import something that is not white onto the bitmap or vector layer
3. Select the camera layer
### --System Information--
- Pencil2D Version:
  Nightly Build October 11 2016
- Operating System:
  macOS Sierra

Jose-Moreno:
@scribblemaniac Ah. I somehow knew this was going to be brought up after re-reading my own comments on the other issue. 

This is not a bug. It's the relative transparency system in effect. I talk about it here: 

https://www.youtube.com/watch?v=GJFW9hzny_M

To make it render "fully opaque" you just have to press the top timeline circle and make it tint black, instead of grey.

[...]

PR #660, opened 2017-04-28
fixed an issue that would render layers translucent on a camera layer

CandyFace:
It was mentioned in #552 that some time ago, whenever you selected a camera layer, everything would be shown "as if rendered" 

Prior to this PR only the selected layer would be shown fully opaque.

demonstration of the PR:
opacity

### DIFF
--- a/core_lib/canvasrenderer.cpp
+++ b/core_lib/canvasrenderer.cpp
@@ -360,20 +360,14 @@ void CanvasRenderer::paintTransformedSelection( QPainter& painter )
 
 void CanvasRenderer::paintCurrentFrame( QPainter& painter )
 {
-
-    bool isCamera = false;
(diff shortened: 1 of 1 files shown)


core_lib/canvasrenderer.cpp
```diff
@@ -361,19 +361,23 @@ void CanvasRenderer::paintTransformedSelection( QPainter& painter )
 void CanvasRenderer::paintCurrentFrame( QPainter& painter )
 {
 
+    bool isCamera = false;
+    if (mObject->getLayer(mLayerIndex)->type() == Layer::CAMERA) { isCamera = true;}
+    else {
+        isCamera = false;
+    }
```
scribblemaniac: How about:
```
bool isCamera = mObject->getLayer(mLayerIndex)->type() == Layer::CAMERA;
```

PR #669, opened 2017-05-13
regression: re-added macdeployqt fix

### LINKED ISSUE #668: Latest and Previous monthly build for Pencil2d fails to launch on OS X 10.10
### --Issue Summary--

Download Application from main site. Attempt to launch. It fails to start with the message:

"Pencil2D cannot be opened because of a problem. Check with the developer to make sure Pencil2D works with this version of OS X. You may need to reinstall the application. Be sure to install any available updates for the application and OS X."

Please note, if I open package contents and attempt to start the application from the command line, here is the error message I get: 

/Volumes/User\ Data/XXXXX/Downloads/Pencil2D\ 2/Pencil2D.app/Contents/MacOS/Pencil2D ; exit;
dyld: Library not loaded: /usr/local/Cellar/qt@5.7/5.7.1/lib/QtWidgets.framework/Versions/5/QtWidgets
  Referenced from: /Volumes/User Data/XXXXX/Downloads/Pencil2D 2/Pencil2D.app/Contents/Frameworks/QtSvg.framework/Versions/5/QtSvg
  Reason: image not found
Trace/BPT trap: 5
logout

Note that the QT framework libs are in the Frameworks directory for the application build. So this is probably a build bug, referencing library paths on the build host instead of within the package contents. 

### --Actual Results

I expected the application to start. It did not. 

### --Expected Results--

It should have started. 

### --System Information--
+ Pencil2D Version:

+ Operating System:

+ RAM Size:

+ CPU:

+ Graphics Tablet:

---

CandyFace:
This should fix #668

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -98,6 +98,11 @@ after_success:
 
       echo "Copying necessary Qt frameworks";
       macdeployqt Pencil2D.app;
+      echo "applying macdeployqt fix";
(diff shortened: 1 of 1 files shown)


paranoidfactoid:
Yup, this works. Thanks.

PR #671, opened 2017-05-15
Linux packaging improvements

J5lx:
This might come a little out of nowhere but basically all it does is bring the Linux nightlies on equal footing with the Windows and macOS ones. The current .tar.gz nightly builds only contain the executable itself and expect the user to take care of all dependencies, while AppImages are completely self-contained. It’s sort of like a macOS App Bundle in a single file. To use it, a user only needs to download it, grant execute access (i.e. `chmod +x`) and execute it. This also takes care of https://github.com/pencil2d/pencil/pull/633#issuecomment-301342447.

A nightly build in AppImage format (generated via this PR) is available at <https://drive.google.com/open?id=0ByqfJUKSPYQtNS1BM1Rrak5fUFU>.


pencil2d.desktop
```diff
@@ -0,0 +1,9 @@
+[Desktop Entry]
+Version=1.1
+Type=Application
+Name=Pencil2D
+GenericName=Animation Software
+Comment=Create traditional hand-drawn animation using both bitmap and vector graphics
+Icon=pencil2d
+Exec=Pencil2D
```
mapreri: this should be `Exec=Pencil2d %f`, shouldn't it?

mapreri:
Let me thank you for the .desktop file, I'm not sure why I've never done a PR for it… :)

pencil2d.desktop
J5lx: Probably, I added the desktop file mainly because linuxdeployqt wants one, so I didn’t really pay attention to that. I’ll go over it again and see how I can improve it for normal installation.

J5lx:
Please don’t merge yet, I’m going to work on the desktop file a little more.

pencil2d.desktop
mapreri: If you are going to tweak the .desktop file, please add a Keywords field, I suggest `Keywords=picture;draw;animation;` but feel free to expand of course.

J5lx: Alright I’ll do that. Just [keep in mind](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-1.1.html#recognized-keys):

> The values […] should not be redundant with the values of Name or GenericName.

So I won’t add animation.

CandyFace: You could call it 'animate' instead, or change the Generic name.

J5lx:
Okay, I’m done, PR is ready to merge. I added the following things:

- A MIME package. This allows file managers and other programs on Linux to recognize Pencil files as such and automatically open them with Pencil.
- Install support. Think `make install`. For now it’s Linux-only since I am not familiar with the customs of other platforms. The prefix can be set by passing `PREFIX=…` to qmake.

pencil2d.desktop
J5lx: Actually the point of that sentence is not to avoid some dangerous name clashes or anything, in fact the reason for discouraging redundancy is pretty boring: Because it doesn’t provide any value. I believe you are mostly a macOS guy, so I suppose you are about as clueless about the Linux ecosystem as I am about the macOS ecosystem. Desktop Entries on Linux are somewhat like shortcuts on Windows or (from what I read on Wikipedia) Aliases on macOS. They contain information about a certain target such as an application or a link, and they are often placed in a central location which serves as a database of sorts. The Desktop Entries in that location can then be read by programs to find suitable applications for opening certain file types or to provide application launchers, among other things. That’s where Keywords comes in. Sometimes you might look for an application but can’t quite remember which one it was. In that case you can often search through available applications, mainly through Name (e.g. Firefox) and GenericName (e.g. Web Browser). Keywords however is for cases where you want to provide additional searchable key words, so that the entry can be found with not-so-specific searches.
So putting animation in Keywords isn’t very helpful simply because it can already be found in GenericName.

For reference, this is how the specification explains Keywords:

> A list of strings which may be used in addition to other metadata to describe this entry. This can be useful e.g. to facilitate searching through entries.

Whew, that turned into a novel!

J5lx:
I also added German translations to the desktop entry and the MIME package. Unfortunately, OneSky supports neither format, so translations for those files will have to be added manually.

Jose-Moreno:
@J5lx One thing I tested yesterday was importing sound on the appimage you initially provided yesterday, but everytime I imported a sound file it didn't bring any. Also exporting mp4 didn't create any kind of file. And later on I tried on a windows build for May 4th it did work. 

Is it possible that ffmpeg wasn't packaged on the Appimage? I recall there was a patch to fix the deployment via Travis, but I don't know much else.

J5lx:
Yeah that is a possibility. Some part of my head is still confused by the fact that Pencil uses the FFmpeg binary rather than the libraries. I’ll look into it.

@chchwy, @CandyFace, @ other developers: What do you think of using FFmpeg libraries (libav*) directly rather than the FFmpeg binary? I haven’t really used the libraries myself so far, but if there is no opposition I might eventually take a look at some point.

J5lx:
@Jose-Moreno I haven’t looked into FFmpeg bundling yet, but as for mp4 export it seems that FFmpeg fails to execute correctly when the file does not have an .mp4 file extension.

CandyFace:
@J5lx I do not know why we're bundling a static binary of ffmpeg rather than using its libraries directly either but I fully support your proposal.

Jose-Moreno:
@J5lx @CandyFace As far as I know libav is originally a fork of FFMpeg that has had some bothersome legal issues with it's parent project. On the other hand FFMpeg also has libraries that allow this behaviour... https://en.wikipedia.org/wiki/FFmpeg#Libraries  particularly `libavcodec `and `libavformat`

In FFMPEG's FAQ there's an explanation on how to use them:
http://ffmpeg.org/faq.html#Are-there-examples-illustrating-how-to-use-the-FFmpeg-libraries_002c-particularly-libavcodec-and-libavformat_003f

J5lx:
The FFmpeg libraries were what I meant when I wrote libav*. Actually I haven’t heard anything from the Libav fork in ages. Thanks for the link to the FAQ.

Jose-Moreno:
@J5lx Ah my apologies, I've been looking at the Libav project because I had to install it for another software to work (according to the developers) so I didn't get your libav* reference at first. Oops :sweat: 

Either way I think it's the best to think of having the libraries rather than the binary itself. If you need something let us know so we can see how to help out, have a great week! :smile:

J5lx:
Alright, FFmpeg is now bundled and should work. @Jose-Moreno, could you check whether your issues are gone? The build at https://drive.google.com/open?id=0ByqfJUKSPYQtNS1BM1Rrak5fUFU is updated.

core_lib/movieexporter.cpp
```diff
@@ -90,6 +90,16 @@ QString ffmpegLocation()
 #elif __APPLE__
     return QApplication::applicationDirPath() + "/plugins/ffmpeg";
 #else
+    QString ffmpegPath = QStandardPaths::findExecutable(
+        "ffmpeg",
+        QStringList()
+            << QApplication::applicationDirPath() + "/plugins"
+            << QApplication::applicationDirPath() + "/../plugins" // linuxdeployqt in FHS-like mode
+    );
```
J5lx: I’m not sure here about the multiline syntax, so if a specific style has already been adopted elsewhere and I just couldn’t find it let me know.

Jose-Moreno:
@J5lx Ok, just tested it, now it exports the video even without explicitly typing the .mp4 file extension, the sound that I imported into pencil is also exported in the final movie BUT weirdly enough the sound does not play inside Pencil2D and the keyframe container does not change shape or shows the waveform like before. 

Regarding the other issue I just mentioned I think I'll have to file a linux-specific bug report considering the sound import visualization was working on a windows version from 10 days ago. However right now it is a bit difficult for me to confirm if this is a problem with how the appimage works or a problem with the source code itself. I'll try to see how to install the debian version of Pencil2D I quoted before and compare. If it doesn't work I'll file the bug report with that info.

On the matter of the ffmpeg being bundled and working with the appimage under Linux Mint, I think we can say it's solved. Thanks a lot!

Jose-Moreno:
@J5lx It seems I have to correct myself again. Apparently the windows version you had on your google drive repo which I was using yesterday was also suffering from the same problem as the linux version regarding the sound import visualization, so it is a source code issue. 
However testing mp4 export only brings movies about 5 seconds long, which was an issue that was resolved last year after implementing the feature.

To test this I opened QT creator and ran the latest master source. This problem is not happening with the latest version. I guess I'll have to wait whatever travis builds next to fully confirm if it's a linux issue or just a difference in the repos.

Once again thanks for all your efforts. I'm afraid I have to stop testing for today since I have to finish other work, but I'll try to be around to test more things when needed. Cheers!

J5lx:
I believe the sound import doesn’t work because (unlike movie export) it doesn’t have any special logic for locating the FFmpeg binary yet, it just takes whatever it finds in the PATH. I’ll take care of that.

chchwy:
Sorry for late reply. In my opinion, a main drawback to use ffmpeg as static library directly is dependency. It make pencil2d harder to compile. At the moment Pencil2D is extremely easy to compile because there is no 3rd party dependency except Qt. All libraries included.

Once the pencil2d is bundled with ffmpeg library, Pencil2D could be compiled only if the ffmpeg development environment is settled down, headers, libs, dlls, a specific compiler, which I personally don't want to. Don't forget Pencil2D is a cross-platform app so there are 3 platforms we need to deal with.

In the flip side, how much benefits will we get from a static linked ffmpeg? does it run much faster? use less memory? easier to maintain? 

However, it is still not a bad idea to do a built-in video exporter rather than an external library. I would recommend to create an abstract video exporting interface with many implementations, for example a ffmpeg exporter on linux or a AVFoundation exporter for Mac, if anyone want to contribute. :p

chchwy:
@J5lx give me a ping if these branch is ready for merging.

J5lx:
It is ready for merging.

As for the exporter, I agree that a library-based exporter would complicate build-time dependencies slightly, but in turn it will also simplify run-time dependencies since the linker will take care of everything and we’ll no longer need our own *platform-dependent* where-the-hell-might-this-binary-be logic.

Even when it comes to the build-time dependency I don’t think it is a deal breaker. I already started experimenting with a library-based exporter, and right now it is entirely optional: If you don’t have the FFmpeg libraries, the library-based exporter simply won’t be built. For most contributors this probably won’t be much of a problem because the exporter is not exactly part of the core functionality anyway (plus, for now the binary-based exporter will still be built anyway) and they can start hacking on other parts of the program without it. And if they do want to build the exporter, it is rather trivial to obtain the libraries:
- On Linux they are practically always installed by default and will be found automatically by qmake so it’s a non-issue. On some distributions special development packages are needed but installing them is absolutely trivial and one of the most fundamental skills of any Linux developer. 
- On macOS it is fairly trivial to install them using Homebrew. One environment variable needs to be set for qmake to find them.
- For Windows, pre-built libraries [are available](https://ffmpeg.zeranoe.com/builds/). One variable needs to be passed to qmake so it knows where to look.

The main advantage of using libraries would be that it is a much, much cleaner solution: No heaps of temporary files, no home-baked file format handling code (for WAV), and generally more control and precision (most notably more precise progress which takes the entire process equally into consideration rather than just the process of saving the frames as temporary files). Also, IMO it’s just more natural to use libraries when available: Why rely on an external program when there is a dedicated API, an *application programming interface*, specifically made for, well, application programming?

As for memory consumption and speed, I can’t say for sure right now. Probably it will be roughly comparable. As mentioned before there will no longer be lots of temporary files consuming space, though.

J5lx:
BTW, my WIP on the library exporter is at <https://github.com/J5lx/pencil/tree/export-lav>. To enable the new exporter, CONFIG+=EXPORT_LAV must be passed to qmake and it must know where to find the libraries.

- On Linux it looks for the pkg-config packages, so those need to be available. On Arch Linux they are in the ffmpeg package, on a modern Ubuntu they are in libavcodec-dev, libavutil-dev and libswscale-dev.
- On macOS it also looks for the pkg-config package. An easy way to obtain it is via Homebrew (i.e. `brew install ffmpeg`). When building with the homebrew package, the PKG_CONFIG_PATH environment variable must contain /usr/local/lib/pkgconfig.
- On Windows, download and extract the dev package from <https://ffmpeg.zeranoe.com/builds/> and run qmake with `LAV_DIR=C:\path\to\ffmpeg-x.x.x-winxx-dev`. At runtime, the dlls from the shared package must be present.

Known issues:

- No audio support
- Only H.264 supported for video
- No muxing support (i.e. no container format)
  - This is a problem for many players. The only players that were able to play my test files are mplayer and ffplay.
- Packaging for the nightly builds is currently not very efficient, since the exporter and the FFmpeg binary both come with their own version of the libraries. This is especially true for Windows builds which are now at 50-55MiB.

Speaking of which, nightly builds with the new exporter are available at <https://j5lx.de/pencil-builds-j5lx>. Due to lack of a Mac, I haven’t been able to test the macOS builds though.

chchwy:
Agree with you that a library-base exporter would be cleaner, and faster definitely.

I would prefer it to be an **open** option, not stick to a specific av library like ffmpeg. I don't like the ffmpeg because just as you mentioned, to compile ffmpeg by ourselves is freaking difficult, and then we have to use precompiled binaries which means it has been already compiled by a specific compiler.

chchwy:
Alright, thanks @J5lx great work. I'll create a new thread about video exporter.

PR #675, opened 2017-05-17
Integrate file choosers in import/export dialogs

### LINKED ISSUE #673: Import/Export file chooser field instead of dialog
Currently the file dialogs for import and export are handled slightly differently depending on the type of media. Exporting images and image sequences shows an options dialog first and a file chooser second, exporting movies and importing image sequences does it in the opposite order. But instead of figuring out the One True Dialog Order it might be a good idea to just turn the file path into another option like this:

screenshot from 2017-05-16 03-14-01

If others are okay with this, I can come up with a proper widget and adjust the dialogs. Thoughts?

J5lx:
Resolves #673. In the end I went for an ImportExportDialog that contains the file chooser and serves as a base class for all the individual dialogs. In most cases switching the individual dialogs over was rather trivial, but I completely restructured the image sequence import dialog to match the structure of the other dialogs.

This is what it looks like:
Import image sequenc dialog
Export movie dialog
Export image sequence dialog
Export image dialog

BTW, this PR depends on (and includes) #674 and will probably be much nicer to review once that one gets merged.


Jose-Moreno:
@J5lx Man this looks awesome! This solution looks like a great step in the right direction! Thanks a lot :smile:  I really wish that I could have time to learn C++ and Qt properly to help everyone a bit more with the coding of this project, though.

J5lx:
I rebased the PR on master, so the changes that are part #674 are no longer part of the diff.

chchwy:
Is it necessary to double quote the file path? I think is more neat without the double quote.

chchwy:
Also we probably should remove the text `(Set by Camera)`. The export resolution doesn't have to equal to the camera's resolution, it is ok to export a 1920x1080 image with a 640x480 camera if you don't mind to get a blur image.

J5lx:
As for the text (Set by the camera), it’s already present in master:
screenshot from 2017-05-19 16-20-55
But I agree that it’s not very useful and I’ll remove it in a second.

The double quotes are for clarity when selecting more than one file, it’s what the Qt file chooser dialog does as well:
screenshot from 2017-05-19 16-13-28
They are only “eyecandy”, though, so I can remove them if you like. Just tell me if you want to get rid of them or not.

CandyFace:
If possible, a comma for separation and no double quotes is what I would prefer.

scribblemaniac:
:+1: for the double quotes. There are plenty of people who use spaces in their filenames, so the double quotes help visually and programmatically segment the list.

Issue #678, opened 2017-05-26
[BUG] Pressure sensitivity checkboxes are disabled / misplaced


Jose-Moreno:
### --Issue Summary--

Pressure sensitivity is working by default with graphic tablets. However attempting to disable the pressure using the appropriate checkbox doesn't have any result. This happens on every tool.

### --Actual Results

Using the pressure checkbox doesn't do anything. On Paint tool the pressure checkbox does not do anything BUT the feather checkbox is enabling and disabling the pressure.

### --Expected Results--

Pressure sensitivity checkbox should allow the user to enable or disable the pressure sensivitity to achieve different effects.

### --Steps to reproduce--

1. Open Pencil2D
2. Select a drawing tool (i.e. Pen tool)
3. Draw. Notice you have pressure enabled.
4. Press the checkbox to disable pressure.
5. Draw again. Notice you still have pressure sensitivity enabled.
6. Select Brush tool.
7. Draw. Notice you might not have pressure on.
8. Disable Pressure and Enable Feather.
9. Draw. Notice you now have pressure enabled for the Brush tool.

### --System Information--
+ Pencil2D Version:

Pencil2D Nightly Build May 20 2017

+ Operating System:

Windows 7 Ultimate

+ RAM Size:

+ CPU:

+ Graphics Tablet:

Wacom Intuos Pro 5
---

CandyFace:
Posting this (slightly edited) comment I made on the PR^ here, as it's relevant to the issue.

Pressure doesn't do anything for brush unless feather is on. I recall it was discussed whether Brush should have size pressure or not and I believe that it concluded that pressure should only affect feather at the time.

Because of that and the fact that there's no constant control for feather blending yet, makes it feel a bit broken. 
Therefore whenever you need feather, set pressure on too, otherwise it won't work. I want to add more customisation for the tools but I don't want to make changes before we've implemented the MyPaint library as such change might become obsolete.

ryyharris:
[2019 Review] I downloaded the Pencil2D v0.6.3 version for MacOS and the pressure sensitivity works for me. I tried to reproduce the bug using the steps above, but nothing went wrong. 
However, I fond that wen I ran the Pencil2D v0.6.3 in the virtual box ubuntu emulator, the pressure sensitivity did not work at all, whether or not the pressure checkbox was marked.

scribblemaniac:
[...]

PR #679, opened 2017-05-27
fixed brush dap artifact and pressure toggle

### LINKED ISSUE #677: [BUG] stabilization function create artifacts on drawing strokes
### --Issue Summary--

### --Actual Results

### --Expected Results--

image

### --Steps to reproduce--

1. Open Pencil2D
2. Select Pen or Pencil tool
5. Draw again.

### --System Information--
+ Pencil2D Version:

+ Operating System:

Windows 7 Ultimate

+ RAM Size:

+ CPU:

+ Graphics Tablet:

Wacom Intuos Pro 5
---

CandyFace:
This fixes #677 #678

Notice though that pressure still won't do anything for brush unless feather is on. I recall it was discussed whether Brush should have size pressure or not and I believe that it concluded that pressure should only affect feather at the time.

Because of that and the fact that there's no constant control for feather blending, it feels a bit broken. Therefore whenever you need feather, set pressure on too, otherwise it won't work. I want to add more customisation for the tools but I don't want to implement anything yet if we intend to integrate mypaint.

### DIFF
--- a/core_lib/interface/scribblearea.cpp
+++ b/core_lib/interface/scribblearea.cpp
@@ -418,10 +418,10 @@ void ScribbleArea::tabletEvent( QTabletEvent *event )
 
     // Some tablets return "NoDevice" and Cursor.
     if (event->device() == QTabletEvent::NoDevice) {
-        currentTool()->adjustPressureSensitiveProperties( pow( ( float )mStrokeManager->getPressure(), 2.0f ),
+        currentTool()->adjustPressureSensitiveProperties( mStrokeManager->getPressure(),
                                                       false );
     } else {
-        currentTool()->adjustPressureSensitiveProperties( pow( ( float )mStrokeManager->getPressure(), 2.0f ),
+        currentTool()->adjustPressureSensitiveProperties( mStrokeManager->getPressure(),
                                                       event->pointerType() == QTabletEvent::Cursor );
     }
 
--- a/core_lib/tool/brushtool.cpp
+++ b/core_lib/tool/brushtool.cpp
@@ -350,7 +350,7 @@ void BrushTool::drawStroke()
     {
         qreal brushWidth = 0;
         if (properties.pressure ) {
-            brushWidth = properties.width * m_pStrokeManager->getPressure();
+            brushWidth = properties.width * mCurrentPressure;
         }
         else {
             brushWidth = properties.width;
@@ -395,8 +395,6 @@ void BrushTool::paintVectorStroke()
         // Clear the temporary pixel path
         mScribbleArea->clearBitmapBuffer();
         qreal tol = mScribbleArea->getCurveSmoothing() / mEditor->view()->scaling();
-
-        mStrokePressures.append(0.01);
 
         BezierCurve curve( mStrokePoints, mStrokePressures, tol );
                     curve.setWidth( properties.width );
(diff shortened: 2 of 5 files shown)


Jose-Moreno:
@CandyFace I get what you mean. But I don't know how long will MyPaint integration take. To be honest I doubt Francois (feeef) has had time to work on it ever since last time.

I'm seriously thinking of setting up a bounty on bountysource for that. But I wonder if it should be an ok solution. Like we can ask for money to fill that bounty and allow someone to work on it to get payed, but I also would like for Matt (chchwy) to overview the integration since he knows the program better than most.

I'll see what we can do for this.

CandyFace:
@Jose-Moreno 
I've spoken to feeef recently and it's true that he hasn't worked on the integration since the last time he showed it. However, a lot of the integration is done, what's keeping us from getting it implemented as I see it, is that you can't zoom or whenever you do so (by enabling the old way of doing it), the quality of the canvas degrades. Aside from that we also need to get it up to date with the master.

If we really want Pencil2D to be usable and recognized as a animation software rather than merely a prototype for doodling, then we need to either rework the current brush engine (bitmap and vector) and do a lot of the calculations ourselves for drawing and applying anti-aliasing, or complete the integration of MyPaint. 

Given that feeef has already worked a lot of the latter, I think it would be silly to start on the former, rather we should get more contributors to look into that problem and fix it.

Jose-Moreno:
@CandyFace I completely agree. I've already investigated into the bountysource possibility, to my surprise pencil2D had already a team page that was unclaimed, so I made a claim to allow us to handle it. This way we could post a bounty on that integration and ask a dev to finish it and be paid for it. I know this is a huge undertaking and feeef has helped a lot. I'd really like him to finish integrating the feature but without monetary incentive I know it is really hard to convince anyone when survival is at hand.

Once I've been answered by the bountysource team I'll let you all know to see what our next step could be in order to secure community support.  I feel a few engines and features need to be refactored in order to allow taking more complex enhancements e.g vector engine as we  talked before.

I also have to apologize to everyone since I haven't really had the time to put myself and report every feature request that I compiled, including the separate bug list I have also gathered lately. I just need to turn in a milestone in a project I'm working on next week and hopefully I can have 2 days free to do this, but I can't promise much for now.

J5lx:
@Jose-Moreno This might be a little late but I’d like to point out that BountySource seems to be somewhat half-dead. One of their featured teams, the diaspora project, is already looking into alternatives such as [FreedomSponsors](https://freedomsponsors.org) right now. References:
- Very slow response times: <https://twitter.com/Bountysource/status/772629418103615488>
- Only barely maintained: [https://github.com/bountysource/core/issues/1077#issuecomment-284277780](https://github.com/bountysource/cor%65/issues/1077#issuecomment-284277780)
- diaspora core team members looking into a replacement “given the amount of issues and the state of maintenance it's in”:
 <https://discourse.diasporafoundation.org/t/adopting-bithub/650/2>

Nonetheless the general idea of rewarding development is good and AFAICT BountySource is still functional at the very least, just keep in mind that it’s not in the best shape right now.

Jose-Moreno:
@J5lx Thanks for letting me know about this issue. 

I just thought It was the best type of service we could use, considering that the team is not a Foundation or some similar figure. I'll look for similar options then, but what I think would be important is to allow for both community and developers to have ease of mind while dealing with  income-based development. I want everyone to feel secure that both their pledges and development is going to be rewarded as transparently as possible.

Also don't worry about being late with this info, I made the claim yesterday, so It might take a month for us to get a response, so I have time to look for other options. I'll let you guys know what I find soon.

Issue #165, opened 2013-08-05
To add a way for RESETING the "OPEN RECENT" files


ghost:
Last build 04Aug2013
Win7 Ultimate, SP1 64bits

To add a new button/shortcut that allows Pencil to RESET to zero the list of recent files.
When finishing a project/document  there is no need for it to be listed on the OPEN RECENT Menu any more. 

Resetting this list will allow a visual update of the work already in use.

J5lx:
#165 got merged and this feature is now available in master, so shouldn’t this be closed?

CandyFace:
Indeed it should, I just forgot about the issue ;)

PR #686, opened 2017-06-06
clear recent files list #165

CandyFace:
Clear recent files list button:
screen shot 2017-06-06 at 20 53 02
request from #165 

Informative dialog box:
screen shot 2017-06-06 at 20 53 23

I've also fixed orientation and zoom not being reset when creating a new document,
as requested from the google drive doc: 
**when creating a new file, canvas state (rotation and zoom) should be reset (currently pan is reset).**

It was commented with no info as to why, I assume it must have been broken some time ago. I've tested it numerous of times since I re-added it though and it seems to work fine now.

As for the button, I've implemented it a preference key like all the other preference options, though since it's a button and not something to set or toggle, I wasn't sure whether this would be the right way to do it. Do comment if you believe I could have done it differently.


app/informativedialog.cpp
```diff
@@ -0,0 +1,19 @@
+#include "informativedialog.h"
+#include "ui_informativedialog.h"
+
+InformativeDialog::InformativeDialog( QString title, QString description, QWidget *parent ) :
+    QDialog( parent ),
+    ui(new Ui::InformativeDialog)
+{
+    ui->setupUi( this );
+    ui->horizontalLayout->setSpacing(0);
```
J5lx: Isn’t this already set in the ui file?

app/preferencesdialog.cpp
```diff
@@ -581,6 +595,11 @@ void FilesPage::autosaveNumberChange(int number)
     mManager->set(SETTING::AUTO_SAVE_NUMBER, number);
 }
 
+void FilesPage::clearRecentFilesList(bool b)
+{
+    mManager->set(SETTING::CLEAR_RECENT_FILES_LIST, b);
+}
+
```
J5lx: I, too, am not 100% happy with using a preference key for this. I don’t think we should squeeze the button into this structure no matter what. Also, I got the success message two times once; I couldn’t reproduce it since then but I suspect it could be somehow related to this.

app/mainwindow2.cpp
```diff
@@ -380,6 +381,13 @@ void MainWindow2::updateSaveState()
     setWindowModified( mEditor->currentBackup() != mBackupAtSave );
 }
 
+void MainWindow2::clearRecentFilesList()
+{
+    mRecentFileMenu->clear();
+    InformativeDialog *informDialog = new InformativeDialog("Notice!", "You have sucessfully cleared the list");
+    informDialog->exec();
```
J5lx: Personally, I’d find it more intuitive to simply enable/disable the button depending on whether the list is empty (already clear) or not, though a dialog is certainly more explicit.

app/informativedialog.h
```diff
@@ -0,0 +1,23 @@
...
+{
+    Q_OBJECT
+
+public:
+    explicit InformativeDialog(QString title, QString description = QString(), QWidget *parent = 0 );
+    ~InformativeDialog();
+
+private:
+    Ui::InformativeDialog *ui;
+};
```
J5lx: This dialog feels a little like reinventing [the weel](https://doc.qt.io/qt-5/qmessagebox.html).

CandyFace: This is somewhat true, although the idea with a custom dialog box is that you can do whatever you want. A QMessageBox is fixed to whatever settings it has. For example having the "ok" button to the right.

My goal is to make our own unique UI scheme that fits Pencil when we've fixed the tools and other things with higher priority, thus having standard buttons, toggles and such has to go.

app/mainwindow2.cpp
CandyFace: I had this initially and while I too think it's more intuitive, I found it annoying that the computer should decide whether or not that button should be clickable. I can change it though if you are against that.

app/preferencesdialog.cpp
CandyFace: Right, that's what I thought too. I'll see if I can come up with a better solution without storing it into a preference key. 

Interesting, haven't experienced that but i'll look into it.

app/informativedialog.cpp
CandyFace: indeed, I have removed setSpacing now for the upcoming commit.

app/informativedialog.h
J5lx: I see. In that case the question here is whether we want consistency within Pencil or within each user’s operating system. Personally I feel that OS-level consistency is much more important unless a completely custom UI such as Blender’s is used, since most users have much more experience with the conventions of their system rather than the conventions of Pencil. That’s why so many toolkits and systems have [Human](https://developer.apple.com/macos/human-interface-guidelines/overview/themes/) [Interface](https://community.kde.org/KDE_Visual_Design_Group/HIG) [Guidelines](https://developer.gnome.org/hig/stable/) [and](https://developer.microsoft.com/en-us/windows/apps/design) [such](http://www.s-and-b.ru/syshlp/motif_guide/MotifStyleGuide/TOC.html).  In general I figure that Qt already does a rather decent job at consistency, there already are “standard buttons, toggles and such” and there are native styles for Windows and macOS as well as a GTK+ style for Linux. That’s just my two cents anyway.

app/mainwindow2.cpp
J5lx: I see. I’ll think about it some more.

app/informativedialog.h
CandyFace: Guidelines are helpers and instructions to stay within a certain design, they're not rules. They meant to be bend, twisted and broken in the most evil ways :P I know about Human interface guidelines and all that UI/UX stuff but I still don't agree with most of them. 

I'm not saying we should go the blender route, that would be extreme and maybe not even good, but a more personal UI should be possible even if it has to inherit OS-level consistency. I haven't looked too much into it yet but it's on my personal roadmap.

chchwy: Is there any new feature will be added into InformativeDialog? QMessagebox::information() is doing the same job.

CandyFace: It's not quite the same, as I mentioned in my first reply, the button is centred and it doesn't have the icon. You can't control those things as far as I know with the standard QMessageBox.

Ideally I just want the text aligned with the icon if I were to use the standard QMessageBox::Information, but even that I can't do.

app/mainwindow2.cpp
J5lx: I came to the conclusion that ultimately I’m in favour of enabled/disabled, but it could be a good idea to toggle the text as well like some programs do, e.g. between “Clear” (enabled) and “Currently empty” (disabled) or something like that. That’d be clearer and at the same time it feels like one part of the program (the preferences dialog) is actually “communicating” with the other (the recent files list) rather than just showing a confirmation all the time no matter what. It would still be nice to have a third voice on this, though.

CandyFace:
Eh sorry for leaving this open for so long, I was trying to implement it in-between my final exams, didn't work out :P

Anyway, I've removed the informativeDialog, and made it possible for the button to update and disable when the list is empty.

chchwy:
Cool, I'm going to merge it since it gets the job done.

Ultimately I hope this clear button could be a part of the menu like this:


But let's think about it later. Good job @CandyFace

PR #687, opened 2017-06-15
Bucket tool - Added tolerance slider, refactored a bit

CandyFace:
Applies to bitmap only:
High tolerance will blend more with alpha, which seems to work nice for pencil, lowering the tolerance makes the algorithm less strict about when to fill through small gaps.

fill

although gif (early) doesn't show it, there's also a spinbox.
 
The vector flood fill is a lot more complicated than the bitmap, so I haven't touched it yet.


Jose-Moreno:
@CandyFace Hey man, nice job! I think this is a great fix that was very much needed by this tool. A few questions: 

1. Does the tolerance range form 0 to 255? 
2. Do you think it's possible to add a numerical representation of the tolerance before merging the commit?

Thanks a lot for your hard work!

If this gets merged I was thinking of making a sort of livestream event on facebook live to showcase the advances that have been made on Pencil2D, so people understand that there's been progress. Hopefully we can show this and all the other advancements that you guys have made over the past months :smile:

CandyFace:
Morr, as mentioned below the gif, there's also a box to show the specific value (known as spinbox) :) it ranges from 0-100.

Jose-Moreno:
@CandyFace I'm sorry I saw it on my email and didn't read that! That's even better then :smile:

CandyFace:
I've refactored, removed and added more stuff:
- removed Anti-Aliasing checkbox in vector mode
- Refactoring
- Pencil Vector mode:
    - added contour fill checkbox. (check gif)
    - removed brush slider and box for vector mode (doesn't do anything)
fillcontour

Jose-Moreno:
@CandyFace That's actually a really nice Idea, I usually use the lasso tool to create organic shapes when not drawing. This way the Pencil tool can become more useful while using the vector layers. Normally you only would use it to create "invisible" lines, but this additional mode feels right. Great work!

app/tooloptionwidget.cpp
```diff
@@ -47,14 +47,34 @@ void ToolOptionWidget::updateUI()
...
+        {
+            mSizeSlider->setVisible(false);
+            mBrushSpinBox->setVisible(false);
+            mUsePressureBox->setVisible(false);
+        }
+    } else {
+        if (currentTool->type() == PENCIL)
+        {
+            mFillContour->setVisible(false);
+        }
```
chchwy: Probably we need to find a way to organise all these visible/invisible switchings.

Issue #195, opened 2013-09-19
Pencil includes itself in "Sound & Video" menu categores


hughsie:
In Linux, Pencil has a file pencil.desktop with "Categories=Graphics;2DGraphics;Development;AudioVideo;"

This means it gets put into the "Sound and Video" groups on some desktops, and _both_ the "Audio" and "Video" menus on other desktops.

Could you please either drop AudioVideo  to make it "Categories=Graphics;2DGraphics;Development;" or add AudioVideo  to make it "Categories=Graphics;2DGraphics;Development;AudioVideo;Video"

I don't think Pencil2D is really an Audio/Video program, and I don't really think it's a Development program either, although I 100% agree with "Graphics;2DGraphics;"

Comments welcome, thanks.

Jose-Moreno:
@mapreri Sorry to bother you Mattia with a Pencil2D issue, but do you know how to change the category of an application in linux distros? Is it a problem of the source code or just how the distribution categorizes it? It's just so we can at least advance a bit with this issue. Thanks :smile:

mapreri:
Well, it comes from the .desktop file, as the OP wrote.

pencil2d "ships" one from the util/linux-debian-packager.sh but I don't know who uses that script.  In Debian I added [my own .desktop file](https://anonscm.debian.org/git/collab-maint/pencil2d.git/tree/debian/pencil2d.desktop?h=debian/0.5.4%2bgit20170323.ec4712d%2bdfsg-1) (although I realize only now that I forgot to add a respective .mime file…)

CandyFace:
I don't think the linux-debian-packager is used anymore? or it has been updated since then because nowhere does it say "Development;AudioVideo".

J5lx:
I don’t have any idea where the categories named above come from. I searched for that combination on the internet but I didn’t find anything related to Pencil2D. Anyway I think it’s safe to say that this was introduced somewhere downstream. On top of that I recently introduced a [a more recent desktop entry](https://github.com/pencil2d/pencil/pull/671/files#diff-37648b414abdcad00bf7f64e37d08dfb) to the repo anyway as part of #671, which is now automatically installed when using `make install` on Linux. That entry is currently using `Categories=AudioVideo;AudioVideoEditing;Graphics;2DGraphics;VectorGraphics;RasterGraphics;Qt;
[...]

PR #688, opened 2017-06-20
Cleanup util/

J5lx:
Firstly, I removed the FFmpeg wrapper script that was needed as a workaround for a bug in linuxdeployqt since I fixed that bug a while ago.

Secondly I removed a bunch of packaging files from util/ as explained in https://github.com/pencil2d/pencil/issues/195#issuecomment-306255441. In addition to the files mentioned in the comment I also removed the windows build/deploy scripts, since I believe they are now obsolete thanks to the introduction of AppVeyor builds. Correct me if I’m wrong, though.


chchwy:
Could you please keep win-deploy.ps1? It's useful when I want to pack the app manually.

J5lx:
Sure, I amended the commit accordingly.

chchwy:
Thanks.

PR #690, opened 2017-06-21
Fix compilation with MSVC 12.0

J5lx:
I caught another compiler error that I originally missed because it appears to be specific to MSVC. The parameter type of ToolManager::setTolerance was changed from qreal to int, but the isnan check on the parameter was not removed, so the compiler didn’t know whether it should cast the parameter to float, double or long double. Since int can never be NaN to begin with, I removed the check.


J5lx:
For reference, here’s the error: <https://ci.appveyor.com/project/J5lx/pencil/build/1.0.95/job/4spc0qkyrkcrw2s7#L281>

PR #693, opened 2017-06-23
ToolOption button visibility refactoring.

CandyFace:
from comment in #687 

Is that fine @chchwy or did you have something else in mind?

### DIFF
--- a/app/tooloptionwidget.cpp
+++ b/app/tooloptionwidget.cpp
@@ -51,29 +51,7 @@ void ToolOptionWidget::updateUI()
     mToleranceSpinBox->setVisible(currentTool->isPropertyEnabled( TOLERANCE ) );
     mFillContour->setVisible( currentTool->isPropertyEnabled( FILLCONTOUR ) );
 
-    auto currentLayerType = editor()->layers()->currentLayer()->type();
-
-    if(currentLayerType == Layer::VECTOR)
-    {
-        // TODO: implement merge behavior
-//        mVectorMergeBox->setVisible( currentTool->isPropertyEnabled( VECTORMERGE) );
-        mToleranceSlider->setVisible(false);
-        mToleranceSpinBox->setVisible(false);
-        mUseAABox->setVisible(false);
-
-        if (currentTool->type() == PENCIL)
-        {
-            mSizeSlider->setVisible(false);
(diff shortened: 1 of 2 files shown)


app/tooloptionwidget.cpp
```diff
@@ -304,6 +282,41 @@ void ToolOptionWidget::onToolPropertyChanged( ToolType, ToolPropertyType eProper
...
+    else
+    {
+        switch (propertyType)
+        {
+            case PENCIL:
+                mFillContour->setVisible(false);
+                break;
+            default:
+                break;
+
```
J5lx: Empty line, inconsistent with first switch.

J5lx:
Personally I think simply factoring part of the visible/invisible switchings into a separate method doesn’t make all that much of a difference; it is still quite a mess in that it first disables *everything* and then manually does the setVisibility thing, potentially twice for some tools. I’ll ponder this myself some more.

CandyFace:
Given that widgets are disabled/hidden by default, I think that we can simply remove the disableAllOptions method.

chchwy:
Thanks @CandyFace for your quick response.

Firstly, this is a minor issue at the moment, so it's totally ok to leave it as it is.

`ToolOptionWidget` is a widget shared by all tools, it will be better to avoid all explicit 
`if ( currentToolType == OneTool )`  and `currentLayerType == OneType` kind of conditions. It is why the `isPropertyEnabled()` was introduced, a more general way to show/hide UI controls.

These **if conditions** reveals another problem of our code, the pencil tool on vector layer and bitmap layer are 2 different tools. In this case, pressure and brush width are only valid on bitmap's pencil, not on vector's pencil. But because there is no way to distinguish these 2 modes, we ended up getting hundreds of  `if ( bitmap) else (vector)` everywhere. This UI show/hide is only the tip of the iceberg. However, it's not an urgent issue probably we can deal it later.

CandyFace:
But the tools don't have to share the same functionality, thus you would need some kind of explicit condition to tell whether to use one over another? Making the tools share same functionality across the different layer types doesn't fix the problem, it just ignores it. I understand that having explicit conditions everywhere is a messy solution too however unless you split up the tools into classes, say penciltool (bitmap) and contourtool (vector) or something, i'm not sure how you are going to avoid such a problem.

Issue #691, opened 2017-06-23
The shortcut of tools in tooltips are hard coded


chchwy:
### --Issue Summary--
We have some tooltips showing when users move the cursor over a tool button
for example: "Pencil Tool **(N)**: Sketch with pencil"
But once user customise the shortcut , tooltip doesn't reflect the change.

### --Expected Results--
The tooltips should reflect shortcut changes.
For example, a user change the shortcut of pencil to [Ctrl+P]
The tooltips should be "Pencil Tool **(Ctrl+P)**: Sketch with pencil"

### --System Information--
+ Pencil2D Version:  0.6 alpha

Related source file: `toolbox.cpp`

PR #694, opened 2017-06-23
#691 shortcuts are no longer hardcoded

CandyFace:


chchwy:
#691

chchwy:
The shortcuts don't update until a restart is because tooltips need to be updated manually lol.  The function  `MainWindow2::setupKeyboardShortcuts()` will be triggered after shortcut changes.

However, I'm thinking that removing the shortcut from tool tips. Users could still know the shortcut from the main menu, then we don't have to worry about it. Thoughts?

Jose-Moreno:
@chchwy I don't think it's a big deal if the default shortcuts are not reflected on the tooltips after changing them. However it would be worse to lack those entirely. From a user stand point it's better to have the tooltips pointing at the default shortcut layout even if you change them than not having them at all. 

When you as a user manage to change the shortcuts it's because you have a certain familiarity with the software, but if you don't know which ones are default then it's difficult to learn the tools. However if this is too much of a hassle then we have to make it extra clear that the shortcut layout is on the preferences.

People could look at the shortcut list on the preferences, but let's face it, a rough 90% of our user base does not do that. I've had to point out a lot of times, options that are specifically located in the preferences simply because people do not look for them.  They just open Pencil2D and expect it to work "out of the box".

chchwy:
I think you're right @Jose-Moreno 
Let's keep them and i'm going to merge this pr.

CandyFace:
I think the shortcuts are good to have when you hover over the tools, it's much easier than opening preferences and checking out the shortcut list and it is also a somewhat common design choice being used by various applications too, thus I think it should stay. 

I have been looking into the first special cases but haven't been able to find a solution so far. It is indeed trickier than it first seems, initially I thought that adding "None" to empty keys in kb.ini would work, but it doesn't accept strings so that wasn't possible. Then I tried to check whether the QKeySequence would be empty an then replace it's content, but that no good either as it's just a reference to the shortcut key (which doesn't exist) 

The second special case shouldn't be that hard, it's just a matter of calling the toolbox  update when you've hit assign from the shortcut list (or so I imagine) that should be enough to trigger an shortcut update.

chchwy:
User could get shortcut information from main menu, not necessary to open preference window 


But I agree that it's good to show shortcuts in tool tips.

Issue #622, opened 2017-03-03
Free software directory license auditing


mangeurdenuage:
### --Issue Summary--
Hello I am giving a hand for the Free Software directory (FSD).
https://directory.fsf.org/wiki/Main_Page
And I was doing a collection of software tools that are meant for art purposes.

I stumbled on pencil and was gladly surprise by it so I began to audit it to see if I could add pencil to the directory and into the said collection but unfortunately for now pencil doesn't have the necessary files to be added in the FSD 

To be validated in the FSD pencil needs each file to contain a preamble of the license it uses (even scripts and content like image for example (CC-BY-SA is co) .
For what I have read pencil is under GPLv2 for now.
Further more I have also noticed that third parties software/libraries like "zlib" and "quazip" are missing their licenses in their folders.

### --Expected Results--

Could someone be kind enough to clarify the situation by adding a preamble of the license for each files and adding missing licenses for third parties libraries, please.
If you want an example of how it should look like please see the libreJS source code (or any other gnu software) 
https://ftp.gnu.org/gnu/librejs/librejs-6.0.13.tar.gz

### --Why all this boring paperwork instead of coding ?--
Licenses like the GPL are meant to protect the freedom of the users (and developers since they are users too) unfortunately over the years corporation or business have abused licensing via legal loopholes so that tools like pencil can be a tool of injustice instead of being a tools of creation.
The work so that software can be protected is infinitively more beneficial than letting people use the tools, that where shared in most innocent ways, in a malevolent purpose.
Well that's just my opinion.

I would like that Pencil be a part of the FSD.
Thanks you for given a bit time for this.
I wish you good luck and
Happy Hacking

Jose-Moreno:
@mangeurdenuage Hi. So far, we'd have to see what's the status on each individual license. This might take some time, since most of the volunteers are busy, including myself, and we don't have an stablished "core team" per say. 

Hopefully we can bring this to clarity in a timely manner. Thanks for taking interest in the project and for looking to protect those who code it and use it as well.

mangeurdenuage:
Hello,
Thank you for your response.
--we'd have to see what's the status on each individual license--
I have seen some files are under the copyright of one of the developers of synfig if I don't mistake.
They have been update to GPLv2+

-This might take some time,-
I know and understand, if you need help with clarification about licenses please do ask whenever you want I know there are a lot of misconceptions about licensing.

[...]

PR #695, opened 2017-06-28
About dialog links to GPLv3, should actually be GPLv2

J5lx:
Just something that I happened to run across due to https://github.com/pencil2d/pencil/issues/622#issuecomment-311619489. Better ~safe~ correct than sorry


chchwy:
#622  @Thanks @J5lx

PR #696, opened 2017-06-29
Movetool deselection properly resets

CandyFace:
For this to work I had to rebind the play button, otherwise it's going to override the action.
It has been bound to Ctrl+Return or CMD+Return on Mac.

Now when you press Enter/Return, the transformation is applied and when you press Escape, the transformation is discarded and reset. Before it would only remove the selection but not update the position until you had clicked on the canvas.


chchwy:
Thanks @CandyFace

PR #698, opened 2017-07-05
Added preference for color palette background

martinvanzijl:
The checkered background makes the color labels hard to read on my PC (I'm on Windows 7).

example-on-my-pc

So I added a preference in the "Tools" page to set it to either checkered or white.

Screenshot below:

color-palette-background-preference-screenshot

Video below:
[color-palette-background-preference-video.zip](https://github.com/pencil2d/pencil/files/982653/color-palette-background-preference-video.zip)

### DIFF
--- a/app/colorpalettewidget.cpp
+++ b/app/colorpalettewidget.cpp
@@ -29,6 +29,7 @@ GNU General Public License for more details.
 #include "colormanager.h"
 #include "colorpalettewidget.h"
 #include "ui_colorpalette.h"
+#include "preferencemanager.h"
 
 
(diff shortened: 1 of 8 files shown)


chchwy:
Probably we should fix it rather than adding a new preference option.
I think the white-grey grid background is not suitable for list mode.

scribblemaniac:
I think the checkered background should only be applied to the color boxes. This was one of the things I was planning on changing but I obviously haven't gotten around to yet.

CandyFace:
I agree that it should be fixed rather than being worked around. The reason it's checkered though is so you can see all colors properly, but it's true that it makes text harder to read in list mode.

chchwy:
Since #700 is a better solution for readability, I'm gonna to reject this PR.
Still say thank you @martinvanzijl for raising this issue.

PR #699, opened 2017-07-05
Load the FPS value when opening an animation

martinvanzijl:
When opening an animation, load the "currentFPS" value if present. For example if the saved file has:

    <currentFps value="12"/>

...then set the FPS value in the editor to 12.

This will save me having to remember and adjust the FPS each time (I usually have a sound-track timed to a certain fps).

Video below (apologies, this runs quite slow):

[load-fps-value-video.zip](https://github.com/pencil2d/pencil/files/1125805/load-fps-value-video.zip)

### DIFF
--- a/app/mainwindow2.cpp
+++ b/app/mainwindow2.cpp
@@ -1168,6 +1168,8 @@ void MainWindow2::makeConnections( Editor* pEditor, TimeLine* pTimeline )
     connect( pTimeline, &TimeLine::soundClick, pPlaybackManager, &PlaybackManager::enbaleSound );
     connect( pTimeline, &TimeLine::fpsClick, pPlaybackManager, &PlaybackManager::setFps );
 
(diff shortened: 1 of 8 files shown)


chchwy:
`currentFps` is a legacy field, I remember it is used by 0.4.4 and some early 0.5.x versions.

chchwy:
Thanks @martinvanzijl

J5lx:
Wait a minute, according to the OP this PR is about restoring the framerate of saved files on load, but since  currentFps is a legacy field as @chchwy said and it isn’t written anywhere in the current code, all this does is reset the framerate to 12fps when loading files written by recent versions of Pencil. Isn’t this a little incomplete or am I overlooking something?

chchwy:
You're right @J5lx. I'm doing the missing part, writing fps and other information into xml on saving the file.

PR #700, opened 2017-07-06
Update color palette background

scribblemaniac:
The main widget background is back to white, while the individual color swatches have a checkered background which is visible for transparent/translucent colours like so:
screen shot 2017-07-06 at 12 59 48 am

This is an alternative approach to #698. (@martinvanzijl Sorry, I'm not trying to compete with/upstage you 😛  I just think this is a better solution to the ugly background as it displays translucent colours better).

Edit: Woot, got #700 🎉


chchwy:
cool @scribblemaniac

martinvanzijl:
@scribblemaniac No problem, your solution sounds good.

PR #701, opened 2017-07-07
Automatically name new palette colors

scribblemaniac:
When a new palette color is created, a name from the [NBS/ISCC Centroids color dictionary](http://tx4.us/nbs-iscc.htm) that best describes the color is automatically selected instead of asking for the name. The name can still easily be changed by double clicking the entry in list mode. In addition to smoothing out the workflow a bit, these changes will also be very beneficial when adding support for importing palette formats which don't require names (ex. .gpl). And as usual these days, here's a screenshot:
screen shot 2017-07-07 at 12 39 30 am
Because of the dictionary selected, some colors may seem to have slightly off names, that's just how it is. Additionally there are currently no adjectives to differentiate brightness, but this could be added fairly easily using the L value, I just haven't done this yet because I'm not really sure if I want to make the names any longer.


app/colorpalettewidget.cpp
```diff
@@ -294,6 +294,66 @@ void ColorPaletteWidget::updateGridUI()
...
+    // Convert XYZ to CEI L*u*v
+    // Helper function for the conversion
+    auto f = [] (const double a) { return a > 0.008856 ? cbrt(a) : 7.787*a + 16/116; };
+    // XYZ tristimulus values for D65
+    const qreal xn = 95.047,
+                yn = 100,
+                zn = 108.883;
+    const qreal l = y/yn > 0.008856 ? 116*cbrt( y/yn ) - 16 : 903.3*y/yn,
+                u = 500 * ( f( x/xn ) - f( y/yn ) ),
+                v = 200 * ( f( y/yn ) - f( z/zn ) );
```
J5lx: It’s been some time since I last worked with colorspace math so I won’t even begin to pretend that I understand this, but either way there are way too many magic numbers in here for my liking. It’d be great if you could state the reference/source at the very least.

scribblemaniac: Of course. The algorithms are from here:
https://www.cs.rit.edu/~ncs/color/t_convert.html#RGB%20to%20XYZ%20&%20XYZ%20to%20RGB
And xn, yn, and zn are from here:
https://en.wikipedia.org/wiki/Illuminant_D65
As you can see most of the numbers are just part of the formula, don't ask me what they do because I don't know either 😝

J5lx: It would be nice to have that as a comment in the source code for future reference, so the people who might have to work on the code in the future have something to go by, that’s all I meant. Since I also used that kind of math in the past without understanding the theories behind it I can’t really blame you on that part 🙂

scribblemaniac: Added! Thanks for the feedback.

chchwy:
I'm thinking, are all of these colour names need to be translatable? Just thinking...

scribblemaniac:
> I'm thinking, are all of these colour names need to be translatable? Just thinking...

I figured it would be a good idea. If people are using pencil in another language, and all of their colors are named in English, it's a) going to be inconsistent with their expected behavior and b) annoying because they will have to name rename all the colors they add. I realize there are quite a few to translate, but they are quite simple phrases so it shouldn't be too difficult to do.

PR #702, opened 2017-07-07
Upload nightly builds only from master branch or when forced

J5lx:
See [`1d67de0` (comment)](https://github.com/pencil2d/pencil/commit/1d67de0751e1b91bc0ee765a2ab6f73a58f65ff1#commitcomment-22953496). The way I implemented it, nightly builds can still be uploaded by setting the environment variable FORCE_NIGHTLY_UPLOAD to yes (because that’s what I want).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,10 +15,6 @@ os:
   - linux
   - osx
 
-branches:
-  only:
-  - master
-  
 python:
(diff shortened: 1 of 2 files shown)


chchwy:
Thanks it is helpful @J5lx

PR #703, opened 2017-07-09
Add Clear Menu item to recent files list

scribblemaniac:
Now there's a 'Clear Menu' item in the recent files submenu that functions similarly to the preferences button added in #686, except in a better place.
screen shot 2017-07-09 at 1 29 01 am

### DIFF
--- a/3rdlib/quazip/quazip.pro
+++ b/3rdlib/quazip/quazip.pro
@@ -10,6 +10,8 @@ CONFIG += staticlib
 
 INCLUDEPATH += ../zlib
 
+QMAKE_CXXFLAGS += -std=c++11
+
 # Input
 include(quazip.pri)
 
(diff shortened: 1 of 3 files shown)


3rdlib/quazip/quazip.pro
```diff
@@ -10,6 +10,8 @@ CONFIG += staticlib
 
 INCLUDEPATH += ../zlib
 
+QMAKE_CXXFLAGS += -std=c++11
+
```
scribblemaniac: It does not compile fine on some compilers. With Clang 4.0.0 on Ubuntu, it does not work for me. I probably haven't noticed this until now because I use other compilers (ones that default to C++11 maybe?). I'm getting some errors with QuaZIP about "QAtomic", which probably requires the `std::atomic` type that is only found in C++11+, and this change fixes everything. I will try to push it upstream after some more testing.

J5lx: Alright, good to know!

chchwy: MSVC just simply ignored the `-std=c++11` with a warning, so basically it's fine.

chchwy:
Cool @scribblemaniac

PR #704, opened 2017-07-15
Add option to lock widget positions

scribblemaniac:
This adds an action to the Windows menu that can be toggled to prevent the QDockWidgets from being dragged around or popped out.

Requested by Charmly here:
http://www.pencil2d.org/forums/topic/is-there-a-way-to-lock-the-windows/

Edit: Also thanks @Jose-Moreno for bringing this one to my attention with the Pencil document.

### DIFF
--- a/app/mainwindow2.h
+++ b/app/mainwindow2.h
@@ -91,6 +91,8 @@ public:
 
     void importMovie();
 
+    void lockWidgets( bool shouldLock );
+
     void preferences();
     void helpBox();
     void aboutPencil();
--- a/app/mainwindow2.cpp
+++ b/app/mainwindow2.cpp
@@ -352,6 +352,14 @@ void MainWindow2::createMenus()
         winMenu->addAction( action );
     }
 
+    winMenu->addSeparator();
+    QAction *lockWidgets = new QAction( tr( "Lock Windows" ), winMenu );
+    lockWidgets->setCheckable( true );
+    winMenu->addAction( lockWidgets );
+    connect( lockWidgets, &QAction::toggled, this, &MainWindow2::lockWidgets );
+    bindActionWithSetting( lockWidgets, SETTING::LAYOUT_LOCK );
+
+
     // -------------- Help Menu ---------------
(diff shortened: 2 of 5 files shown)


CandyFace:
This could probably be done in a more dynamic way, for example to allow the user to specifically choose which panels to lock. Aside from that, the code looks good 👍

app/mainwindow2.h
```diff
@@ -91,6 +91,8 @@ public slots:
 
     void importMovie();
 
+    void lockWidgets( bool shouldLock );
```
J5lx: Shouldn’t this go into the slots section since it is used with connect()?

scribblemaniac: You're right, it probably should. You'll notice that most of the other functions in this class have a similar issue, so it will require a bit more refactoring to fully correct the issue.

J5lx: I see. This makes me wonder: Does placing a method in the slots section actually do anything or is it just semantic sugar for stuff like code completion? I mean, if it actually did something we probably should have gotten warnings or bugs by now… Anyways, I agree that we should fix all of those sooner or later either way, even if it only makes the code more comprehensible.

CandyFace: StackOverflow https://stackoverflow.com/questions/35879514/public-functions-versus-public-slots mentions that prior to Qt5, slots were the only way to communicate with signals, now you can just use normal functions, so my guess is that it's just syntactic sugar.

scribblemaniac: @CandyFace Good to know, thanks for the research!

chchwy:
Good work! @scribblemaniac

app/mainwindow2.h
J5lx: Actually I just found out it’s *not* only syntactic sugar, it affects the generation of the meta object. That means that the way it currently is only `&MainWindow2::lockWidgets` will work in `connect()` calls, but `SIGNAL(lockWidgets(bool))` will not work because `SIGNAL()` relies on the meta object which currently doesn’t know that lockWidgets is a slot.

PR #705, opened 2017-07-15
Brush opacity/regression

CandyFace:
### Fixed (again)
There has been a bit of experimentation to get rid of this problem. To be able to draw a brush stroke continously with the same alpha, the stroke has to be drawn in one draw call. This is very important, as it will otherwise cause alpha blending (dabs intersection) when using < 100% alpha like shown below:
screen shot 2017-07-15 at 11 41 50

This caused some other problems I had to fix too though, for example keeping the grid on top and animating during playback.

### New
Grid is hidden during playback.

### Other
- Refactoring

### Remaining problem?
Because of the way the paintBItmapBufferRect is called (while holding down the button) the alpha blending behavior can't be avoided during playback.


chchwy:
@CandyFace sorry I don't really get this PR. 
How could I get the dabs you show in the image?

CandyFace:
it happens whenever you paint with less than 255/100% alpha. For example set alpha to 50 or something and notice the dabs intersecting.

dabs

chchwy:
That's a little bit weird because the alpha blending is not that bad on my side. (it's July 11 nightly build on my MacBook)
giphy 1

Probably it is a platform specific issue?

CandyFace:
O_o okay that's weird, considering i'm testing on a Macbook too, it can't be a platform specific issue as we're on the same platform. I've tried to replicate what you're getting on the 11 July NB and I still get the same visible dabs.

CandyFace:
Update: useFeather was missing pieces of code to make it update properly on launch, this made it look like it was disabled, even though it wasn't. This means that if you haven't touched the checkbox before, then it is ENABLED. You should be able to replicate the problem by:

1. Enable feathering
2. Disable feathering
3. Draw

Result:
Notice visible dabs.

You can also try the following:
1. delete your pencil config here (on OS X): ``/Users/"Your Name"/Library/Preferences/com.pencil.Pencil.plist``

2. Open Pencil2D, notice that feather seems to be off
3. Draw, notice reasonably good alpha blending
4. Enable, then disable feathering
5. Draw and notice visible dabs.

CandyFace:
I've included brush for continuous alpha blending again, when feather is off.
alpha blending

Composite_Source instead of Composite_SourceOver is what caused the problem, though the other changes which has been applied in this PR are still relevant.

Jose-Moreno:
@CandyFace Great work Oliver! Yesterday I was giving support to a person on the facebook fan page and I was testing the brush. Normally the feather option should help to create a brush outline both smooth or hard, similar to photoshop or krita brush settings. This is used to control the "edge" of your paintings, although I doubt a lot of people use Pencil2D to paint, there are some who do.

Maybe feathering should be always turned on, and leaving the slider to control how "feathered" or smooth the brush is could suffice. This is different than being anti-aliased though, you can have a sharp / hard brush with aliasing, or without it and it will only affect how the pixels are displayed, but not it's main purpose.

Seriously great work, and thank you for taking the time to do such a robust check up on these features!

CandyFace:
@Jose-Moreno Thanks :) 
I agree that the feather tick is redundant. Given the functionality of the brush, it should always be on.

Anti-aliasing shouldn't be used on feathered brushes though because of the way feathering has been implement, it practically does it when it applies the radial gradient that causes the density to change.

You can see what I mean here:
screen shot 2017-07-20 at 16 40 31

First stroke (top to bottom) is done with feathering (and AA off) and second is done without feathering (and AA on)

chchwy:
Sorry @CandyFace haven't got time to review this pr, will do it in 2 or 3 days.

chchwy:
Sorry for late reviewing, good job @CandyFace

core_lib/tool/penciltool.cpp
```diff
@@ -229,14 +236,14 @@ void PencilTool::paintAt( QPointF point )
...
-        qreal opacity = 1.0f;
+        qreal opacity = 1.0;
         mCurrentWidth = properties.width;
         qreal brushWidth = mCurrentWidth;
 
         BlitRect rect;
 
         rect.extend( point.toPoint() );
-        mScribbleArea->drawPencil(point,
+        mScribbleArea->drawPencil( QPoint( qRound(point.x() ), qRound(point.y() )),
```
chchwy: @CandyFace  Did you remember why you replace the QPointF with QPoint?

CandyFace: Thinking in context of pixels, you can't have a half pixel, so there should be no reason why the point you're drawing is floats.

I believe that's why I changed it.

chchwy: Just did some small tests. Please check the results as below:
Zoom 845.6%, Pencil width 2


The left image is using QPointF, the Right image is using QPoint.

In the right image, the stroke has some "sudden moves", which is the timing of floating numbers just cross the 0.5, for example, from 1.49 to 1.52. It's a very small move, but causes a whole pixle shift after rounding to integers.

Since Qt it quite ok to deal with floating-point positions. I think we should go back to QPointF. I'll do it later, just let you know. :)

CandyFace: Ah yes my mistake, it's affecting the antialiasing of the brush... I didn't think about that when I made the change. Do change it back to floating-points ;)

chchwy: Just a reference: 2226afd16406f0d9d50df6390e29c22e89392dd9

PR #707, opened 2017-07-16
Fix segfault when repopulating RecentFileMenu after clearing it

J5lx:
The call to QMenu::clear() wasn’t doing anything we didn’t do ourselves anyway except for deleting QActions we still needed. Instead we are now simply removing the clear actions without deleting them right away.

### DIFF
--- a/core_lib/interface/recentfilemenu.cpp
+++ b/core_lib/interface/recentfilemenu.cpp
@@ -37,7 +37,8 @@ void RecentFileMenu::clear()
     {
         removeRecentFile( filename );
     }
-    QMenu::clear();
+    removeAction( mClearSeparator );
+    removeAction( mClearAction );
     mRecentFiles.clear();
     mRecentActions.clear();
 }
--- a/core_lib/interface/recentfilemenu.cpp
+++ b/core_lib/interface/recentfilemenu.cpp
@@ -31,6 +31,12 @@ QMenu( title, parent )
     mClearAction = new QAction( tr("Clear"), this ); // share the same translation
 }
 
+RecentFileMenu::~RecentFileMenu()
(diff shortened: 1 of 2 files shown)


J5lx:
While I was at it I also fixed a potential memory leak due to the clear actions not being deleted.

core_lib/interface/recentfilemenu.cpp
```diff
@@ -31,6 +31,12 @@ QMenu( title, parent )
     mClearAction = new QAction( tr("Clear"), this ); // share the same translation
 }
 
+RecentFileMenu::~RecentFileMenu()
+{
+    delete mClearSeparator;
+    delete mClearAction;
+}
+
```
chchwy: It's not leaking actually. A QMenu will delete all owned QActions on destroying. Sort of Qt's memory management convention. But it is ok to do it explicitly though.

scribblemaniac:
> While I was at it I also fixed a potential memory leak due to the clear actions not being deleted.

There was no risk of a memory leak here because a parent object is set for both of the actions. When the parent object is deleted, its child objects are automatically cleaned up.

Otherwise good job, I obviously didn't catch this issue and I'm glad to see that it is already resolved.

core_lib/interface/recentfilemenu.cpp
scribblemaniac: @chchwy Ah, just a moment faster than me 😞

chchwy: :p

J5lx:
Ah, didn’t know that, and it isn’t immediately obvious from the code, too. Thanks for telling me 👍

PR #708, opened 2017-07-17
Sound button fix

martinvanzijl:
Fixed the sound button so that it toggles sound on and off when clicked. 

Also added the following icon for when the sound is turned off.

sound-disabled

Video below:

[sound-button-fix.ogg.zip](https://github.com/pencil2d/pencil/files/1153783/sound-button-fix.ogg.zip)


chchwy:
Thanks @martinvanzijl

Issue #683, opened 2017-06-02
Apply a consistent code style


J5lx:
Something that really stands out to me when working on Pencil’s code is that there is no universal code style. While there are certain conventions that are more prevalent than others, it seems that it is never universally consistent. Since I think that everyone will agree that any style is better than no style at all, I compiled a list of the various conventions in use so we can have a discussion as to what to use (I’m aware I might be opening Pandora’s box here). Once there is consent on the style to use, we can unify the style of the existing code and add an [editorconfig](http://editorconfig.org) to help new contributors stay consistent with it (Qt Creator plugin available at [editorconfig/editorconfig-qtcreator](https://github.com/editorconfig/editorconfig-qtcreator)). I’d happily volunteer to take care of the unification and the editorconfig myself.

Ordering is arbitrary, numbers are only there to make it easier to refer to a particular item if necessary.

1. Indentation
    1. four spaces per indent (currently prevalent by far)
    2. one tab per indent
2. End of file
    1. No newline at end of file
    2. One newline at end of file (currently prevalent by far)
    3. Two newlines at end of file (why would anyone want that?)
3. Whitespace in parentheses
    *(sometimes also in other kinds of brackets)*
    1. One space after each opening parenthesis and before each closing parenthesis: `( example )` (currently prevalent)
    2. No spaces after opening parenthesis or before closing parenthesis: `(example)`
4. Whitespace before parentheses
    *(sometimes also before other kinds of brackets)*
    1. No whitespace before the opening parenthesis of function calls, one space before the opening parenthesis of control structures (if, while etc.) (currently prevalent by far)
    2. No whitespace before any opening parenthesis
5. Braces
    1. Always one newline before the opening brace, never omit optional braces (variation of Allman) (currently prevalent by far)
    2. One newline before mandatory opening braces (e.g. function declarations), one space before optional opening braces (e.g. conditionals), never omit optional braces (OTBS)
    3. On rare occasions variations of the above with no whitespace before opening braces or omitted optional braces
6. Trailing whitespace
    1. No trailing whitespace anywhere (currently prevalent by far)
    2. Have indentation even on otherwise empty lines, otherwise no trailing whitespace
    3. Trailing whitespace? Who cares!
7. Aligning consecutive assignments
    1. Don’t align consecutive assignments (currently prevalent by far)
    2. Align consecutive assignments
8. Multi-line function calls / declarations
[...]

PR #709, opened 2017-07-18
Update sound key-frame length when fps changes

martinvanzijl:
Update sound key-frame lengths when the FPS is changed. (Since the amount of frames a sound covers depends on FPS).

Video below:

[update-sound-length-fix.ogg.zip](https://github.com/pencil2d/pencil/files/1157181/update-sound-length-fix.ogg.zip)


core_lib/structure/soundclip.h
```diff
@@ -41,8 +41,17 @@ class SoundClip : public KeyFrame
...
+    void setDuration(const int64_t &duration);
+
+    void updateLength(int fps);
+
 private:
     std::shared_ptr< SoundPlayer > mPlayer;
+
+    // Duration in seconds. This is stored to update the length of the
+    // frame when the FPS changes.
+    int64_t mDuration;
```
J5lx: Unrelated to the PR, but this reminds me we should really add some actual API docs at some point. @scribblemaniac [experimented with doxygen some time ago](https://github.com/pencil2d/pencil/issues/540#issuecomment-255045764).

scribblemaniac: @J5lx That's something I've been meaning to talk about further for a while now, especially in #683.

chchwy:
Thanks mate @martinvanzijl

PR #717, opened 2017-07-29
Fix compiler warnings

J5lx:
All those warnings bugged me so I decided to take them on. Lots of unused parameters and variables, bogus casts and other stuff. Now all remaining warnings are caused by zlib and QuaZip. I tried to find the most sensible solution for each case, but since I needed to venture into parts of the codebase that are unfamiliar to me it would be great if someone who is familiar with them could check these changes as well.


core_lib/structure/layer.cpp
```diff
@@ -474,7 +474,7 @@ void Layer::mouseRelease( QMouseEvent* event, int frameNumber )
 
 void Layer::editProperties()
 {
-    QRegExp regex("([\uFFEF-\uFFFF])+");
+    QRegExp regex("([\\xFFEF-\\xFFFF])+");
```
CandyFace: Have you tested this? as far as I know when you use \x and unicode characters, then you have to encapsulate them in braces.  So it should be [\x{FFEF}-\x{FFFF}] instead.

J5lx: The documentation doesn’t mention braces and it seemed to work fine when testing. What you are probably thinking of is QRegularExpression which uses PCRE, see http://www.pcre.org/current/doc/html/pcre2syntax.html#SEC3. QRegExp uses its own syntax.

CandyFace: Yup you're right, I mistakenly thought that they used the same syntax.

chchwy:
Good work! @J5lx

PR #721, opened 2017-08-02
fix selection rect not being removed when pressing backspace

CandyFace:
note: this only occurred when selecting vector curves.


chchwy:
Thanks @CandyFace

Issue #655, opened 2017-04-22
Cursor being drawn outside window frame


CandyFace:
### --Issue Summary--

bug 1:
While using a tool with a cursor, if you zoom too far in and hit space to update the cursor size, the program will hang.

I assume it's because the cursor is not being clipped to the frame of the window, so when you can't see it anymore, it's constantly trying to update a huge area.

enhancement:
The cursor (the outline circle) should be clipped to the window.

### --Actual Results
Application hangs.

### --Expected Results--
Application shouldn't hang

### --Video or Image Reference--
![outside](https://cloud.githubusercontent.com/assets/1045397/25303744/5ad283ea-2759-11e7-9af8-3f4bdf4b5011.gif)

Notice how the application starts lagging severely or hangs entirely.
Also notice that the cursor is being draw outside the window frame...

### --System Information--
+ Pencil2D Version:
commit: dbc6bda372472340f26fea72a98f5732f64eb2c2
date: Tue Apr 18 13:57:10 2017

+ Operating System:
OSX 10.11.6

+ RAM Size:
8GB

+ CPU:
1.7 GHz Intel Core i7

CandyFace:
it's essentially a part enhancement and part bug. 
The zoom in which ends up freezing the application is definitely a bug, but the cursor not being clipped the the window is intentional. However, my assumption is that implementing the enhancement will fix the bug and vice versa.

CandyFace:
Update:
There doesn't seem to be any way of controlling the rendering priority of a QCursor, and it will therefore be rendered on top of everything as far as I can tell. The solution here is therefore to implement the circle as a normal qpixmap and render in the ScribbleArea paintingEvent.

CandyFace:
Implemented, closing.

PR #725, opened 2017-08-11
Draw circle cursor within canvas area

CandyFace:
Circle cursor is now drawn only within canvas area, see #655

Also changed such that cursor size is affected by zooming.

You can see the difference below:
## Old behavior
canvas cursor-old

## New behavior
canvas cursor

### DIFF
--- a/core_lib/interface/scribblearea.cpp
+++ b/core_lib/interface/scribblearea.cpp
@@ -83,6 +83,8 @@ bool ScribbleArea::init()
     mOffset.setY( 0 );
     selectionTransformation.reset();
 
+    updateCanvasCursor();
+
     tol = 7.0;
 
     setMouseTracking( true ); // reacts to mouse move events, even if the button is not pressed
@@ -141,6 +143,7 @@ void ScribbleArea::settingUpdated(SETTING setting)
 void ScribbleArea::updateToolCursor()
 {
     setCursor( currentTool()->cursor() );
+    updateCanvasCursor();
     updateAllFrames();
 }
 
@@ -605,6 +608,7 @@ void ScribbleArea::mouseMoveEvent( QMouseEvent *event )
             ToolPropertyType tool_type;
             tool_type = event->modifiers() & Qt::ControlModifier ? FEATHER : WIDTH;
             currentTool()->adjustCursor( mOffset.x(), tool_type ); //updates cursors given org width or feather and x
+            updateCanvasCursor();
             return;
         }
     }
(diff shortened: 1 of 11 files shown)


core_lib/interface/scribblearea.cpp
```diff
@@ -617,6 +621,9 @@ void ScribbleArea::mouseMoveEvent( QMouseEvent *event )
 
     currentTool()->mouseMoveEvent( event );
 
+    // used to clear and update canvas cursor
+    update();
```
chchwy: Is it possible to update only a small rectangle rather than update the whole canvas?

chchwy:
Thanks for your work @CandyFace. Generally I like this idea, it is more natural to draw the circle of brush size just in canvas. Other than that, the canvas update area issue is relatively minor atm because something else consumes more time. Maybe we can worry about it later.

core_lib/interface/scribblearea.cpp
CandyFace: I've tried that but it's not fast enough to perceive the cursor if you move too fast and will therefore leave a fragments of the cursor. It doesn't matter how big the rect is.

CandyFace:
Yeah I'm thinking, if possible, we could make use of Qt OpenGL and use off-screen rendering to render to the cursor faster and possibly other things too, for now though I think this is okay.

chchwy:
I have some clues about performance. It is not related to Qt's rendering engine, it is because we copy the buffer image too many times during the process of drawing.

You could have a look at `BitmapImage::extend()` and count how many times it is called in a stroke.

CandyFace:
Ah yes sure enough, it's being called a lot per stroke. Depending on the speed of the stroke, 20-25 times... that's quite excessive. We could probably gain quite a bit by optimizing that. Another thing would also be to switch to QPixmap instead of QImage because it's faster to render and allows for hardware acceleration. https://techbase.kde.org/Development/Tutorials/Graphics/Performance#An_alternative_approach_to_the_problem

Jose-Moreno:
@CandyFace @chchwy Nice job! Now that you talk about hardware acceleration, if you guys consider this for the future, please make it optional. We have a lot of users with low-end computers, some even hailing from Kenya, so if we go and diminish their ability to keep up with the software it will be a huge impact for them who actually use the app. Just a thought. Thank you for the hard work as usual!

CandyFace:
@Jose-Moreno It will be optional if we implement it. Just like you see it in most raster and vector applications. You'll be able to toggle it somewhere in preferences. :)

PR #746, opened 2017-08-29
Fixed sound not starting over when loop is active

### LINKED ISSUE #733: [Bug] Sound not looping when using "loop" toggle
### --Actual Results

Sound continues until the end of the file while the playhead continues to loop. The sound doesn't start again.
### --Expected Results--

The sound should loop from the beginning everytime the playhead does. It should work when range is turned on as well.

### --Steps to reproduce--

1. Import a sound
2. Create a keyframe on another frame to mark the ending of your animation
3. Press the "Loop" button
4. Press the playback button
5. Sound plays until the end of the sound file

Alternatively---

1. Import sound
2. Turn on range
3. Press on LOOP button
4. Press play.
5. Sound plays until the end of the sound file
### --System Information--
+ Pencil2D Version:

commit: 3cc8eba25490d360cc058555cb1eab3cb23ce667
date: 2017-08-22_13:18:18

+ Operating System:

Windows 7 Ultimate

+ RAM Size:

+ Graphics Tablet:

---

CandyFace:
fixes #733
### Future improvements:
Changes on the timeline made while playback is on, are not updated until it has been stopped again.
This can cause various problems that may crash the program. 
A simple solution could be to stop playback whenever you click on the timeline.

### DIFF
--- a/core_lib/managers/playbackmanager.cpp
+++ b/core_lib/managers/playbackmanager.cpp
@@ -147,7 +147,7 @@ void PlaybackManager::playSounds( int frame )
             }
 
             // Set flag to false, since this check should only be done when
-            // starting play-back.
+            // starting play-back, or when looping.
             mCheckForSoundsHalfway = false;
         }
         else if ( layer->keyExists( frame ) )
@@ -156,6 +156,16 @@ void PlaybackManager::playSounds( int frame )
             SoundClip* clip = static_cast< SoundClip* >( key );
 
             clip->play();
+
+            // save the position of our active sound frame
+            mActiveSoundFrame = frame;
+        }
+
+        if ( frame >= mEndFrame )
+        {
+            KeyFrame* key = layer->getKeyFrameWhichCovers( mActiveSoundFrame );
+            SoundClip* clip = static_cast< SoundClip* >( key );
+            clip->stop();
         }
     }
 }
@@ -191,6 +201,7 @@ void PlaybackManager::timerTick()
         if ( mIsLooping )
         {
             editor()->scrubTo( mStartFrame );
+            mCheckForSoundsHalfway = true;
         }
         else
         {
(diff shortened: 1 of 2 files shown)


chchwy:
Cool Thanks @CandyFace

Kaikogf:
@CandyFace 
The implementation you made (hitting the timeline will stop the playback) is working fine.
But there is a similar situation/bug as the one described by @Jose-Moreno described on his post (that was already closed): 

I have tested the following:
. Open Pencil2D
. Select bitmap layer and add a few of them (the more the better for testing accordingly with the length of the sound file)
. Import a sound file (choose a file that starts almost immediately after hitting playback, just for testing purpose)
. Activate Loop
. Hit Playback

Two results here: 
- when the timeline reaches the last frame it will loop to the first frame and the sound file will start all over again
- when in playback mode, and before the timeline reaches its end frame, if you hit the _**START**_ button (_left arrow on the playback buttons_) several times, the timeline pointer will rollback to the first frame but the sound file will continue to play without interruption. So no Loop here


Pencil2D version used:
Last build 31August2017

Kaikogf:
Not sure fi what I wrote here was mentioned by you !

CandyFace:
> when the timeline reaches the last frame it will loop to the first frame and the sound file will start all over again

This is the expected result as far as I know. The player looks for the last frame in the timeline, a frame, no matter its length it will still only count as one frame, so you'll need to have another other frame placed after the sound for it to play through. I'll look into this...

> when in playback mode, and before the timeline reaches its end frame, if you hit the START button (left arrow on the playback buttons) several times, the timeline pointer will rollback to the first frame every time but the sound file will continue to play without interruption. So no Loop here

I can't seem to reproduce this... I will try some more but here it seems to work as expected.

Kaikogf:
I forgot to tell you that the first ["when the timeline reaches the last frame it will loop to the first frame and the sound file will start all over again"] part was OK no issues here: 
👍 


The second part goes like this: 
Considering the first part [mentioned above] this second part is an issue. 
Here the player is not able to look for the last frame in timeline because you hit the START button arrow BEFORE the player reached the last frame, so that is why it does not reset the sound track to start from the beginning as it did on the first part (when it looped to the first frame because it reached the last frame)

Hope this help you to understand what I wrote. If not let me know. I will try with different example/words

Issue #745, opened 2017-08-28
Sound playback is displaced when playing from anywhere but start


CandyFace:
### --Issue Summary--
http://www.pencil2d.org/forums/topic/timeline-and-audio-syncing-issue/#post-29591
> I’m having issues in regard to audio getting slightly behind animation. Specifically if after 4-5 seconds of animation a word was spoken, the animation would come later to mouth said spoken word. If I tried to sync it up in Pencil 2D, the exported result would be the audio continually lagging behind said visual speech.
> 
> Cause of problem I suspect is processing. I’ve noticed if you pause the timeline anywhere after 2-3 seconds and play from the point it stopped, the audio backtracks a little, which in my opinion points to the fact that I believe Pencil 2D isn’t processing it very well. Any thoughts?

### --Actual Results
Sound playback gets displaced when you resume from any point but the start

### --Expected Results--
Sound playback should be consistent no matter where you start it.

### --Video or Image Reference--
editor:
https://youtu.be/x8K3KTFUSxo

export:
https://youtu.be/PqwtD_AcRNw

### --System Information--
Unknown but probably a nightly build.

+ Operating System:
Windows 10

PR #751, opened 2017-09-06
Sound playback changes

CandyFace:
sound-test

Prior to this the timeline didn't consider the length of frames (for sound), so it would stop whenever the last frame had been hit, even if the length of the track was much longer. Now the pos + the length of the track is taken into account.

edit: ignore the first commit, the "real" changes are in the bottom one.


Jose-Moreno:
@CandyFace Awesome man! Thanks!

core_lib/managers/layermanager.cpp
```diff
@@ -261,17 +261,40 @@ bool LayerManager::deleteCurrentLayer()
...
     {
-        int frame = pObject->getLayer( i )->getMaxKeyFramePosition();
-        if ( frame > maxFrame )
+        int lastFramePos = pObject->getLayer( i )->getMaxKeyFramePosition();
+        int prevFramePos = pObject->getLayer( i )->getPreviousKeyFramePosition(lastFramePos);
+
+        if (pObject->getLayer(i)->type() == Layer::SOUND) {
+            KeyFrame* keyFrame = pObject->getLayer( i )->getKeyFrameWhichCovers(lastFramePos);
+            KeyFrame* prevKFrame = pObject->getLayer( i )->getKeyFrameWhichCovers(prevFramePos);
+
```
chchwy: Sorry I don't quite get it. What is `prevKFrame` for?

CandyFace: prevKFrame is used to access length() and pos() so we can compare the previous keyframe and the last keyframe against each other. This is to prevent cases where you're overlapping soundtracks and the previous frame + its length is longer than the last frame + its length.

chchwy: If so, you might have to do that in all sound clips, not just on the previous one.
example:
```
          +----------------- frame3
    +---------------------------- frame2
+--------------------------------------frame1
```

CandyFace: This has been tested against and should work no matter how many clips you use, since it's looping through all layers. 
test3

CandyFace:
Two new additions:
fixed: sounds not resetting when clicking on the "Start" arrow while playback was active.
also implemented such tracks that had been paused, will continue playing again when pressing start. Prior to this only the most recent track would play again, but now we're checking against all active keyframes.

CandyFace:
Hmm i've discovered a bug which affects the playback when the fps has been changed. Don't merge yet.

CandyFace:
@Jose-Moreno  I've made some changes to the way frames are processed now, there should be no more time displacement #745. Are you able to compile my branch and test the changes? I've tested with the the lip sync track we got from the forum for a while now and it seems stable, but I would like you to test it too. If you can't compile, i'll see if I can provide a windows build myself for you to test :)

Jose-Moreno:
@CandyFace I'll download the build you provided and test it against the previous file I made. I'll elt you know as soon as I can.

On another note: There was another issue for a guy in the forum. I tagged you there. He had a crashing issue when importing sound files. The catch is, his system is a bit low-end, and he is using the 32bit variant of the latest windows version.

I think the memory leak issues we had before are still present in Pencil2D, but it's more prominent when you have less RAM memory to work with. I'll tag Matt over there as well so maybe he can take a look as well, since that issue was kind of elusive for the longest time.

Give me a few hours to try and test your branch build. Thanks a lot!

CandyFace:
@Jose-Moreno Thanks, though i've already discovered a few problems though so there's already a new build if you want to try :P

https://drive.google.com/open?id=0B3Jkb3m7pLp4eDMwTlkyWFgxRGc

Fixed:
some playback issues that would be caused with multiple tracks were paused and started again.

As for the memory consumption, i'm not sure this is a problem anymore, at least on my macbook the memory consumption is very low, like 40 megabyte for 10 soundtracks running, and the "helloguys" project which we received from the forum takes up like 60 mb, so either it's much worse on windows, or there's something else wrong in that regard.

Thank you for taking the time to checkout the builds still :)

Jose-Moreno:
@CandyFace To be honest, I can't find any prominent bug in the newest build you linked. The sounds are reproducing as expected and now it's even working with the range functionality which makes it easier to lipsync even without having the waveform. I tried to break the build by looping the range and pressing randomly either Start or End buttons. 

I even used the period and comma keys which de-stabilizes the sync for fun. If you keep the period key pressed while playing the range loop, the sound file will continue to play, but after leaving it alone the playback resumes normally on the next cycle. So I think you did an extremely great job with that.

The only issue I saw was related to exporting, but it's not a bad things actually. Right now the problem is that since the sound layers length are being considered, now when you export a video it will export the video until the sound ends. This is actually preferred most of the time, but we might require a way to restrict the amount of frames we can output on export.

Maybe in the export dialogue we can have a **start frame** and **end frame** box so we don't have to export the whole timeline and edit it later, of course this is natural behavior and would be better than changing back and forth the timeline frame count on the preferences before exporting.

This would also work while we manage as well to get a way to "trim" the sound layers length. For now I'll put that idea as a feature request on the tracker for later. awesome job and thanks a ton for letting me test as well!

CandyFace:
@Jose-Moreno Awesome, that means we can soon merge this stuff.

About the export dialog, I agree that it would be useful to be able to select a start and end frame and maybe also a toggle to export to last keyframe. After this has been merged, I'll probably look into that :)

CandyFace:
@chchwy I consider this done. it has been tested by a bunch of people now and there's been no issues related to this, so you can review and merge now.

chchwy:
Well done @CandyFace !

chchwy:
@CandyFace seems like the travis-ci check is not working correctly, please see the log:
https://travis-ci.org/chchwy/pencil2d/jobs/275273009

PR #754, opened 2017-09-11
fixed ignoring files

CandyFace:
I've tested this a couple of times now, and it seems to work.

Unfortunately I had to make another script for windows too, as I couldn't get appveyor's internal skip_commit.files to work.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -25,14 +25,6 @@ environment:
       PYTHON: "C:\\Python35"
       PLATFORM_: "amd64"
 
-skip_commits:
-  files:
-    - '**/*.md'
-    - '**/*.sh'
-    - '**/*.py'
-    - '**/*.ps1'
-    - '**/*.yml'
-
 install:
   - set PYTHON3="%PYTHON%\python.exe"
   - set UseEnv=true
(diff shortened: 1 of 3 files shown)


scribblemaniac:
Even if the only thing changed are .md files, this should still trigger the doxygen build + delploy. For example, the README.md serves as the homepage for the documentation so the documentation will need to be updated when that is changed. I also want to add the files from docs eventually to the doxygen documentation so that everything is in one place so hardcoding is not an option.

Additionally if documentation-deploy.sh is updated then it should also be triggered. In my opinion lots of the code from the travis file should be moved to external scripts to keep it cleaner, and they would also trigger updates.

CandyFace:
I see no reason why the main readme file nor other files not directly related to source code should trigger a build, even if it's for documentation. Imo. the documentation generated from doxygen should only contain source code, other things can be found on the "wiki". Why do you want to have all that information there too?

The reason i'm ignoring these files is to avoid duplicate builds which has no meaning on google drive when the source code itself hasn't changed.

chchwy:
I think it's ok to have this ignore files because we trigger the Travis auto build very often. Even the change of a .md file doesn't trigger an auto build immediately, the docs will be updated in 2-3 days. We can leave it  at the moment and see if you need a change later.

PR #757, opened 2017-09-14
Ability export movie with specified range

CandyFace:
range-test

### New:
* Ability export movie with specified range.
* Made it possible to type higher resolution values.
* checkbox to quickly switch between whole project length or last painted keyframe.

### DIFF
--- a/app/actioncommands.cpp
+++ b/app/actioncommands.cpp
@@ -107,37 +107,13 @@ Status ActionCommands::importSound()
     return st;
 }
 
-void ActionCommands::updateFrameEnd()
-{
-    int layerCount = mEditor->object()->getLayerCount();
-    int projectLength = mEditor->layers()->projectLength();
-    if ( mExportMovieDialog->useKeyFrameLength() )
-    {
-        int maxPosition = 0;
-        Layer* layer;
-        for(int i = 0; i < layerCount; i++)
-        {
-            layer = mEditor->layers()->getLayer( i );
-
-            if ( layer->isLayerPaintable( layer->type() ) )
-            {
-                maxPosition = layer->getMaxKeyFramePosition();
-            }
-        }
-        mExportMovieDialog->setEndFrame( maxPosition );
-    }
-    else
-    {
-        mExportMovieDialog->setEndFrame( projectLength );
-    }
-}
-
 Status ActionCommands::exportMovie()
 {
     mExportMovieDialog = new ExportMovieDialog( mParent );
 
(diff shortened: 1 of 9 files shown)


app/actioncommands.cpp
```diff
@@ -107,9 +107,37 @@ Status ActionCommands::importSound()
...
+            }
+        }
+        mExportMovieDialog->setEndFrame( maxPosition );
+    }
+    else
+    {
+        mExportMovieDialog->setEndFrame( projectLength );
+    }
+}
+
```
chchwy: looks like this function should be in ExportMovieDailog, not in ActionCommands.

chchwy: I would suggest 
1. Move this function to layer manager to count project length without sound layers.
2. Pass one more parameter like this `ExportMovieDialog::setDefaultRange( start, end, endPaintable )`
3. Move the checkbox connection into ExportMovieDialog class.

The code structure will be more clear and reasonable.

CandyFace: Yeah you may be right about that, alright I'll look into that

CandyFace:
I've made some changes @chchwy, it's not quite like how you suggested though, because I wanted to avoid adding references to Editor and Layers in the class and I didn't see how you would connect the checkbox to anything useful in the export dialog either. it should however be more structured now.

chchwy:
No worries I didn't explain very clearly, I'll put some minor changes later.
Thanks for your work! @CandyFace

J5lx:
@CandyFace Any particular reason for making the export dialog a class member? I’m wondering whether I should do that on my lav exporter branch, too.

CandyFace:
@J5lx Hmm not anymore no, if I recall correctly I did it because I was using it to get access to the variable in another method. I refactored that away though, so it's really just a leftover from that. Given that the window is short lived, there's no reason why we should have the pointer stored always.

Jose-Moreno:
@CandyFace I forgot to tell you that the range export is only working for movies and is not implemented for image sequences. Want me to fill out an issue so you guys can keep track of that? it's not something urgent but would be coherent to have the dialog on each export option for av media.

CandyFace:
@Jose-Moreno Yes please do that. I don't have much time for Pencil2D atm. as I'm working on something else, so it's hard to say when i'll get time to fix it.

Jose-Moreno:
@CandyFace No problem man, we all have been busy lately. I'll do that as soon as I can then. Thanks for keeping tabs on the project too!

PR #758, opened 2017-09-14
Implemented duplicate for sound frames.

CandyFace:
duplicate-sound-tracks

### Added
* duplicate for sound frames
* made sound frames slightly transparent to be able to see overlapping sound tracks

### Fixed
* typos? referring to these `if ( !key->fileName().isEmpty() ) { return Status::FAIL; }` which made no sense before... not empty == fail? how does that make sense O_o

### Other
* refactoring.

### Future improvements:
It should be possible to duplicate frames to the current scrub position by selecting a frame and clicking duplicate. Atm. you are required to have the scrubber within the frame you want to duplicate.

### DIFF
--- a/core_lib/managers/soundmanager.cpp
+++ b/core_lib/managers/soundmanager.cpp
@@ -90,7 +90,7 @@ Status SoundManager::loadSound( Layer* soundLayer, int frameNumber, QString strS
         soundLayer->addKeyFrame( frameNumber, key );
     }
 
-    if ( !key->fileName().isEmpty() )
+    if ( key->fileName().isEmpty() )
     {
         return Status::FAIL;
     }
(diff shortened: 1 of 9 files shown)


core_lib/managers/soundmanager.cpp
```diff
@@ -90,7 +90,7 @@ Status SoundManager::loadSound( Layer* soundLayer, int frameNumber, QString strS
         soundLayer->addKeyFrame( frameNumber, key );
     }
 
-    if ( !key->fileName().isEmpty() )
+    if ( key->fileName().isEmpty() )
```
chchwy: Just think of it. Because in this function it's loading a new audio file, so the SoundClip's file path should be empty, we don't want to overwrite an existing sound clip. So..the original code is correct. I'll make a change and put a comment on it.

CandyFace: Aaah riight yes well that make sense.

PR #760, opened 2017-09-19
Scribble and Canvas stuff...

CandyFace:
Changes:
+ Canvas renderer should never apply bilinear filtering
(smoothTransform)

Fixed:
+ Canvas could get painted even if frame didn’t exist (bitmap).
+ Update was accidentally called recursively… cpu throttle.

Other: 
+ Refactoring

### DIFF
--- a/core_lib/canvasrenderer.cpp
+++ b/core_lib/canvasrenderer.cpp
@@ -80,7 +80,6 @@ void CanvasRenderer::paint( Object* object, int layer, int frame, QRect rect )
     QPainter painter( mCanvas );
 
     painter.setWorldTransform( mViewTransform );
-    painter.setRenderHint( QPainter::SmoothPixmapTransform, mOptions.bAntiAlias );
     painter.setRenderHint( QPainter::Antialiasing, true );
 
     // Don't set clip rect, paint whole canvas. `rect` therefore unused.
--- a/core_lib/interface/scribblearea.cpp
+++ b/core_lib/interface/scribblearea.cpp
@@ -474,6 +474,16 @@ bool ScribbleArea::areLayersSane() const
     return true;
 }
 
+bool ScribbleArea::allowSmudging()
+{
+    ToolType toolType = currentTool()->type();
+    if ( toolType == SMUDGE )
+    {
+        return true;
+    }
+    return false;
+}
+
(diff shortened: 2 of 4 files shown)


core_lib/canvasrenderer.cpp
```diff
@@ -80,7 +80,6 @@ void CanvasRenderer::paint( Object* object, int layer, int frame, QRect rect )
     QPainter painter( mCanvas );
 
     painter.setWorldTransform( mViewTransform );
-    painter.setRenderHint( QPainter::SmoothPixmapTransform, mOptions.bAntiAlias );
```
chchwy: Curious why remove this line? It's an option that you can switch it on and off in Preferences

CandyFace: Set renderHint with smoothPixmapTransform serves no purpose here, it creates a similar effect to that of mipmaps for textures, making the output look blurry which is not an effect you would want for pixel drawing application. The anti-aliasing checkbox in preferences should only affect the edges of vector lines (which it does)

mipmap

notice the output becoming blurry as soon as i let go of the mouse button.

chchwy: Alright, you convinced me. 
Going to merge it.

chchwy:
Mmm...the ignore filter doesn't work
https://travis-ci.org/chchwy/pencil2d/jobs/277990045

CandyFace:
Mmm indeed... I wonder if it's related to committing from master rather than branches. You can see mine worked https://travis-ci.org/CandyFace/pencil/jobs/277793235... I'll have a look at both of them later.

chchwy:
It works when I put another commit to it.
Seems like the ignore filter will fail with a merge request.
https://travis-ci.org/chchwy/pencil2d/jobs/278078388

chchwy:
@CandyFace 
I think I found the problem
Please see the log of this build: https://travis-ci.org/chchwy/pencil2d/jobs/278442027

`git diff --name-only origin/master...${TRAVIS_COMMIT}` output nothing.
it means the shell will work on all branches except `master`.

CandyFace:
I think I have a fix very soon, I just need verify the travis build, then i'll make a pr.

Issue #765, opened 2017-10-01
[BUG] Still Image and Img Sequence export not applying camera motion


Jose-Moreno:
### --Actual Results

When Exporting either an Image Still or an Image sequence when there is a camera interpolation / keyframe motion, it will result in a static frame across the exported frames. Even if you move the camera in the first frame, and export a still image, the camera "animation" it will not be applied.
### --Expected Results--

Camera motion, regardless of export method should be applied and represented in the final result.

### --Steps to reproduce--

1) Import an image that's bigger than the camera frame size. Or reduce the camera size.
2) Select camera layer.
2) go to any frame after the first and create a keyframe for the camera.
3) Move the camera so the image is visibly "out of bounds"
4) Go to File > Export > Image (position playback on any frame) or Image Sequence
5) Review result form either method and confirm that the image has the same placement from when the image was first imported.

### --System Information--
+ Pencil2D Version:

commit: 8fd508104389c097a83e7bc9d9a1ed32a85f9f9a
date: 2017-09-23_13:05:14
+ Operating System:

Windows 7 Ultimate
+ RAM Size:

+ Graphics Tablet:

chchwy:
Hi @Jose-Moreno 
This issue should be fixed in 28 Oct version, could you please confirm?

Jose-Moreno:
@chchwy alright I'll give it a try throughout the day. I'll report approximately at 20:00 pm GMT -5 thank you for your work!

Jose-Moreno:
![came_interp_test](https://user-images.githubusercontent.com/4912907/32136059-6de3ad38-bbce-11e7-92e4-2b201c43ba0f.gif)

I had time early, so I gave it a quick try. Exporting image sequence works well even when rotating the camera :smile: 

I exported a still frame from the file with camera interpolation and it worked well.

**However** I have found a consistent crash when exporting still images that is related to the camera layer keyframes, I created an issue here #770

chchwy:
@Jose-Moreno Thanks for your time :) I'm going to close this.

PR #767, opened 2017-10-05
Vector fill and more

CandyFace:
### What's new
Reworked vector fill, vector selection, refactoring and added draft documentation for VectorImage class. Fixed points sticking to each other on overlap. Curves can now be painted separately, and the width of the curve can be changed as well.

### How it works now:
vector fill1

The first time you draw a curve and fill it, both the area and the curve will be filled with the same color. Afterwards you can choose to select the curve or the area or both to fill. You select a curve or area by using the move tool.  

For now filling a vector stroke will not respect color boundaries but merely fill the path of the curve.

### What's removed
+ curve visualization on hover.


Jose-Moreno:
@CandyFace Awesome!

chchwy:
Cool work @CandyFace! 
I just returned today from a 2-week holiday break, I'll review it asap.
And I think we can prepare for the 0.6 release.

CandyFace:
@chchwy Welcome back :D
That sounds good, though it would be nice if we could get it tested some more. @Jose-Moreno would you mind testing this build:
https://drive.google.com/file/d/0B3Jkb3m7pLp4VzJWb1RVY2ZHX2c/view?usp=sharing
While bugs has been fixed here and there, you focus on how vector filling works now, and the way you fill a curve and area and such.

Jose-Moreno:
@CandyFace Sure thing, i'll test it during the day and let you guys know. On the other hand I reported a bug earlier, that although is not app-breaking @chchwy should take a look at it since it involves the camera layer https://github.com/pencil2d/pencil/issues/765

PR #776, opened 2017-11-03
Minor fixes

scribblemaniac:
Clean up the warnings and make some changes to the modified animation warning.

### DIFF
--- a/app/mainwindow2.cpp
+++ b/app/mainwindow2.cpp
@@ -598,12 +598,12 @@ bool MainWindow2::saveObject(QString strSavedFileName)
     return true;
 }
 
-void MainWindow2::saveDocument()
+bool MainWindow2::saveDocument()
 {
     if (!mEditor->object()->filePath().isEmpty())
-        saveObject(mEditor->object()->filePath());
+        return saveObject(mEditor->object()->filePath());
     else
-        saveAsNewDocument();
+        return saveAsNewDocument();
 }
 
 bool MainWindow2::maybeSave()
@@ -617,8 +617,7 @@ bool MainWindow2::maybeSave()
             QMessageBox::Cancel | QMessageBox::Escape);
         if (ret == QMessageBox::Yes)
         {
-            saveDocument();
-            return true;
+            return saveDocument();
         }
         else if (ret == QMessageBox::Cancel)
         {
--- a/app/mainwindow2.cpp
+++ b/app/mainwindow2.cpp
@@ -970,7 +970,7 @@ void MainWindow2::makeConnections(Editor* editor, ScribbleArea* scribbleArea)
 {
     connect(editor->tools(), &ToolManager::toolChanged, scribbleArea, &ScribbleArea::setCurrentTool);
     connect(editor->tools(), &ToolManager::toolPropertyChanged, scribbleArea, &ScribbleArea::updateToolCursor);
-    connect(editor->layers(), &LayerManager::currentLayerChanged, scribbleArea, &ScribbleArea::updateAllFrames);
+    connect(editor->layers(), QOverload<int>::of(&LayerManager::currentLayerChanged), scribbleArea, &ScribbleArea::updateAllFrames);
 
     connect(editor, &Editor::currentFrameChanged, scribbleArea, &ScribbleArea::updateFrame);
     connect(editor, &Editor::selectAll, scribbleArea, &ScribbleArea::selectAll);
@@ -1000,14 +1000,14 @@ void MainWindow2::makeConnections(Editor* pEditor, TimeLine* pTimeline)
 
     connect(pTimeline, &TimeLine::toogleAbsoluteOnionClick, pEditor, &Editor::toogleOnionSkinType);
 
-    connect(pEditor->layers(), &LayerManager::currentLayerChanged, pTimeline, &TimeLine::updateUI);
+    connect(pEditor->layers(), QOverload<int>::of(&LayerManager::currentLayerChanged), pTimeline, &TimeLine::updateUI);
     connect(pEditor->layers(), &LayerManager::layerCountChanged, pTimeline, &TimeLine::updateUI);
     connect(pEditor->sound(), &SoundManager::soundClipDurationChanged, pTimeline, &TimeLine::updateUI);
 
     connect(pEditor, &Editor::objectLoaded, pTimeline, &TimeLine::onObjectLoaded);
     connect(pEditor, &Editor::updateTimeLine, pTimeline, &TimeLine::updateUI);
 
-    connect(pEditor->layers(), &LayerManager::currentLayerChanged, mToolOptions, &ToolOptionWidget::updateUI);
+    connect(pEditor->layers(), QOverload<int>::of(&LayerManager::currentLayerChanged), mToolOptions, &ToolOptionWidget::updateUI);
 }
 
 void MainWindow2::makeConnections(Editor* editor, DisplayOptionWidget* display)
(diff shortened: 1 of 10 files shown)


app/mainwindow2.cpp
```diff
@@ -970,7 +970,7 @@ void MainWindow2::makeConnections(Editor* editor, ScribbleArea* scribbleArea)
 {
     connect(editor->tools(), &ToolManager::toolChanged, scribbleArea, &ScribbleArea::setCurrentTool);
     connect(editor->tools(), &ToolManager::toolPropertyChanged, scribbleArea, &ScribbleArea::updateToolCursor);
-    connect(editor->layers(), &LayerManager::currentLayerChanged, scribbleArea, &ScribbleArea::updateAllFrames);
+    connect(editor->layers(), QOverload<int>::of(&LayerManager::currentLayerChanged), scribbleArea, &ScribbleArea::updateAllFrames);
```
chchwy: Sorry I don't quite understand. Could you tell what `QOverload` is?

scribblemaniac: `BaseManager` has the function `currentLayerChanged(Layer*)`, and its child class `LayerManager` has `currentLayerChanged(int)`, so the `QOverload<int>::of()` is used to specify that you want the `currentLayerChanged` function with an `int` parameter.

It's a relatively new feature, so [this](https://doc.qt.io/qt-5/qtglobal.html#qOverload) is about the best documentation Qt has on it at the moment.

J5lx:
I sent @scribblemaniac a PR with another warning fix.

J5lx:
Besides, it seems that we need to use a newer version of Qt on AppVeyor in order to use QOverload (currently we use 5.6 but we need at least 5.7). I think it would be best to use 5.9 like we do on Travis, but [that version is not available for the MSVC2013/32bit combo on AppVeyor](https://www.appveyor.com/docs/build-environment/#qt). So we could:

- Switch to MSVC2015 upwards, however I’m not a Windows dev and I don’t know about possible implications of switching MSVC versions.
- Use Qt 5.8 instead, however support for 5.8 ends in less than three months afaik.
- Drop 32bit support. We don’t support 32bit for platforms other than Windows anyway, and 32bit usage is so low these days that even several mainstream Linux distributions [have already dropped support for it](https://www.archlinux.org/news/phasing-out-i686-support/) or have made [first steps](http://www.omgubuntu.co.uk/2017/09/ubuntu-17-10-32-bit-builds-dropped) [in that direction](http://www.phoronix.com/scan.php?page=news_item&px=fedora-24-i686-demotion). However, @Jose-Moreno has mentioned several times people in less developed countries using weaker hardware and I’m not sure what that means in regard to system architecture.

Any thoughts?

CandyFace:
Personally I think we should stay on a QT version that allows the best compatibility for all platforms.
It's always nice to play with new features, but for every time we jump to a new version, we disregard a potential userbase. 

I know for example via @Jose-Moreno that there are a great deal of Windows XP users that use Pencil, which is supported on Qt 5.6. Same goes with Mac which support down to 10.8. Whereas with Qt 5.9 min requirements are windows 7, and mac OSX 10.10.

Jose-Moreno:
@J5lx @scribblemaniac @chchwy @CandyFace It's as candyface says, although I'm not as worried for Windows XP users since most of them are starting to migrate to Windows 7 (I know I did despite my lack of enthusiasm), I am far more preoccupied for MAC OSX users.

An example can be seen on this very recent thread https://www.pencil2d.org/forums/topic/mac-osx-10-95-pencil2d-wont-open/

The user has OSX 10.9.5 and yet he couldn't run Pencil2D no matter what, apparently because the QT version being used (according to our research) didn't add support for it. For example, some people in the past have claimed to have versions as early as OSX 10.6, however we can't hope to cater to everyone since Apple will continue to force their users to upgrade no matter what, and thus QT will be forced to raise the bar on their support via Clang, so despite our best intentions, we need to draw the line somewhere as well. I think as long as we are up-front and properly communicate these changes we'll be fine. If there's code that's more efficient on newer iterations of the framework then we must gather enough reasons to use it and say that clearly.

Of course these cases are sparse, but we have to attempt to achieve a balance between modern features and backwards compatibility for our users. I could try to do a series of surveys, but at the end of the day we might have to limit support for windows and mac. Starting at least from Windows 7 and 10.8 seem like decent targets to me, but what do you guys think, is this possible?.

J5lx:
Sounds reasonable, I didn’t think about the requirements of Qt itself. In that case we’d need to find a solution without using QOverload. I wonder how many people in the userbase are still using macOS<10.10, though. Since we’re already using Qt 5.9 on macOS I suppose we would’ve gotten complaints by now, so it might be fine to keep using 5.9 there and on Linux (since Linux requirements don’t seem much different between 5.6 and 5.9 anyway).

J5lx:
@Jose-Moreno Ah apparently there *are* complaints. I don’t think it’d be difficult to switch back to 5.6 for mac, though, so we could do that. Are there complaints about Linux, too, or can we keep using 5.9 there?

Jose-Moreno:
@J5lx Just to make sure i'll begin rolling some simple surveys on all the socal media we have access to.

So far most complaints come from MAC users. Linux is almost null, I think mainly due to the fact that Linux users are mostly power users, so they know how to solve their stuff, although some home users did have issues before, but mainly because problems with administrator rights (sudo and stuff), I recall a specific problem where someone couldn't run a version of Pencil on ubuntu, but it was because it's an ancient version, it's on the forum here: https://www.pencil2d.org/forums/topic/how-can-i-load-0-44b-ver-files-in-ubuntu/

On Windows, there's been windows xp complaints, but those were years ago, conversely some have had issues with Windows 10, but that might have been due to the QT version being lower than the required target.

scribblemaniac:
The fact of the matter is that what we are debating has already happened. QOperatingSystemVersion which is used [here](https://github.com/pencil2d/pencil/blob/6bf33191e153c47933e37a91415968d4867468e4/core_lib/external/macosx/macosx.cpp#L44) is a 5.9+ Qt class, making it a requirement at the moment for the macOS platform. Now I'm not opposed to changing this an any other necessary code to lower our minimum supported Qt version to perhaps 5.6, which will be supported until March 16th 2019. However, 5.6 supports only supports up to and including macOS 10.11; the latest version is 10.13. So this means that we *may* need two separate macOS builds, which could get confusing.

As for Windows, XP and Vista support should be possible with Qt 5.6, however there are also [some changes](http://www.tripleboot.org/?p=423) we would need to make and we would need to test it on that system. We should be sure that there are no Windows 10 issues however before we even think about legacy systems. The implications of upgrading to MSVC2015 is something we need to look into, as we'll eventually have to change it.

For those not on Discord, I am planning on adding a 32-bit Linux version for the v0.6 release since there was someone who requested it. Currently there are no issues compiling an appimage for this platform with the latest Qt version.

QOverload can easily be replaced with a very ugly static cast (`static_cast<void (LayerManager::*)(int)()`) or we can refactor the signal name as @J5lx suggested.

We should see wait and see what the results of this poll are before making any concrete decisions on what platforms to support.

CandyFace:
Indeed we are already using 5.9 for mac platforms and maybe we should just stay there. Alternatively we could as you also mention @scribblemaniac make a secondary builder for older platforms, for example one using Qt 5.6, to keep support for winxp and mac os 10.8 and 10.9 users. Unless a lot of changes are required for this to work, it should be doable to add a new parameter for the pre compiler to look through.

chchwy:
Ah, my mistake. I shouldn't have named a function which already exists. I can simply rename it and then get rid of `QOverload`.

PR #777, opened 2017-11-04
Use combo box for line interpolation

scribblemaniac:
I found the unlabeled radio buttons to be very user-unfriendly. Unless you think to hover over it, there is no way to tell which radio button is for higher or lower interpolation levels. Having a combo box that uses the label text instead is a much better option. It looks like this now:
screen shot 2017-11-03 at 8 08 30 pm

Even better would be to have a icon button set, something similar to this:
interpolationmockup
Or to have a slider control over the specific weighting of the interpolation. But since we are in the middle of the RC phase right now, I didn't want to mess with this any more than I had to.

### DIFF
--- a/app/tooloptionwidget.cpp
+++ b/app/tooloptionwidget.cpp
@@ -17,10 +17,10 @@ GNU General Public License for more details.
 #include <QLabel>
 #include <QToolButton>
 #include <QCheckBox>
+#include <QComboBox>
 #include <QSpinBox>
 #include <QGridLayout>
 #include <QGroupBox>
-#include <QRadioButton>
 #include <QSettings>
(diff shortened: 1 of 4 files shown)


CandyFace:
I agree that it is un-userfriendly but it was also meant as a temporary solution. it was the intention to make a bar of the sort you're describing. Something like this:

screen shot 2017-11-04 at 09 55 31 as my aim but a more indirect with icons would maybe be even better.

However because we started to focus on bug fixing and the sort, I put it on hold. There are plenty UI related things to be done still, for example we also have the new icons which should get implemented too.

A combo box will do fine for now though, at least it's better than unlabeled radio buttons.  👍

PR #781, opened 2017-11-06
Fix nightly AppImages (again)

J5lx:
Travis really gotta update their Ubuntu, I tell you.

Luckily, though, impact on build time is hardly noticeable, since desktop-file-utils is so small.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,6 +64,15 @@ install:
   - pip install -U google-api-python-client
   - python3 -V
   - pip -V
+  - 'if [ "$TRAVIS_OS_NAME" == "linux" ]; then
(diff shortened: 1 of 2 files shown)


scribblemaniac:
Good job coming up with a proper solution! I assumed we would need to do upstream changes; glad to see it wasn't that complicated after all.

J5lx:
Look at [this](https://blog.travis-ci.com/2017-12-01-new-update-schedule-for-linux-build-images) on the Travis Blog:

> On a related note… Xenial is just around the corner! :tada:

CAN YOU BELIEVE THIS? Let’s just hope that “just around the corner” doesn’t mean “Soon™”…

Issue #774, opened 2017-11-03
[BUG] Graphic glitch when using move tool to clear selection


Jose-Moreno:
### --Issue Summary--
Issue No.1
Ok, I was working on a project I was sent and needed to use Pencil2D. So using the marquee tool I began cutting, pasting and moving. But after a few frames I noticed that somehow a portion of the same frame was copied onto the lower right corner of the image.

After investigating thoroughly, I noticed that when you create a marquee selection, and click outside the selection with the move tool (black arrow) it will make the selection marquee disappear, but also it will create a duplicate of the selected artwork on the lower right corner of the selected content boundary box. 

Even if you create a different marquee selection on a different place, even if it doesn't cover the artwork, clicking outside with the move tool will copy and paste the same part on the lower right portion of the artwork boundary box.

Related Issue No.2
Another issue that seems to be related to this is that while moving the test artwork. When I pressed the select all command (CTRL + A) it wouldn't select just the artwork boundary, but the previous "bitmap space" that the artwork occupied before being moved down.

### --Expected Results--
Clicking outside any selection with the black arrow tool should have the same effect as clearing the selection (CTRL + D / Deselect All) and should not copy anything inside the frame canvas, which can potentially damage the artwork.

### --Video or Image Reference--
Issue No.1
![image](https://user-images.githubusercontent.com/4912907/32358166-9db02b58-c011-11e7-9588-a010e59f1511.png)

Issue No.2
![image](https://user-images.githubusercontent.com/4912907/32358161-904eb77c-c011-11e7-9d3f-6d346a3893af.png)

### --Steps to reproduce--
Issue No.1
1) Open a new pencil file.
2) Draw something
3) Use the marquee tool and select any part of it, or even an empty space on the canvas.
4) Select the move tool and click OUTSIDE the selection.
5) The artwork will become copied inside it's boundary box.

Issue No. 2
1) open a new file
2) draw something
3) Press CTRL + A to select all the artwork. Take note of the current boundary size.
4) Move the whole selected artwork to a different position. 
5) Press CTRL+A to select all. Notice the difference between boundaries.

Alternative Steps (for added Whitespace / Transparency):
3) Instead of selecting all the artwork in step,  select a portion of the artwork OR select a bigger area and just the drawing, including a lot of whitespace. Take not of how much whitespace you are selecting.
4) Move the artwork
5) Press CTRL+A to select all. Notice that the whitespace is being ADDED to the boundary of the selection.

[...]

PR #791, opened 2017-11-07
Minimize Select All operation with autocropping

scribblemaniac:
Now the BitmapImage will automatically resize itself to be at the minimum size possible without cropping any data. This fixes the second issue of #774 and goes a step further to update the Select All bounding box when the edge is erased with an eraser, deleted with a selection clear, or brush strokes/flood fill with an alpha of 0 are applied. Since it resizes the image when necessary, this will also reduce the file size and speed up playback slightly whenever part of the minimum bounding box of an image decreases (most notably when moving the whole layer).

### DIFF
--- a/core_lib/graphics/bitmap/bitmapimage.cpp
+++ b/core_lib/graphics/bitmap/bitmapimage.cpp
@@ -93,14 +93,19 @@ BitmapImage BitmapImage::copy(QRect rectangle)
     return result;
 }
 
-void BitmapImage::paste(BitmapImage* bitmapImage)
-{
-    paste( bitmapImage, QPainter::CompositionMode_SourceOver );
-}
-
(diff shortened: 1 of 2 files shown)


chchwy:
At the moment `Bitmap::paste()` is intensively used when a user is drawing something, it will be called multiple times in one stroke. (in `ScribbleArea::drawBrush()`)

As I observed, the `autocrop()` slows down the stroke apparently (takes 20% CPU time from my profiler).
Is it possible to add a flag so the `autocrop()` will not be always triggered?

scribblemaniac:
Yeah, I see what you mean, I'll push some more optimizations this evening.

scribblemaniac:
I'm going to close this for now until I get things sorted out properly. I am sure that this fix will make it into v0.6.1 though.

PR #792, opened 2017-11-07
Fix error report when making a new file in a read-only directory

scribblemaniac:
Add check when saving if parent directory exists and if it is writable for better error messages. Pretty sure I didn't screw it up this time 😥


chchwy:
Cool man.

PR #799, opened 2017-11-14
Minor eraser improvements

scribblemaniac:
Adds support for eraser dabs, and fixes the bug where the eraser strokes were always being feathered.

### DIFF
--- a/core_lib/tool/erasertool.cpp
+++ b/core_lib/tool/erasertool.cpp
@@ -135,6 +135,7 @@ void EraserTool::mousePressEvent( QMouseEvent *event )
 
     startStroke();
     mLastBrushPoint = getCurrentPoint();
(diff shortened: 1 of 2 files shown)


Jose-Moreno:
@scribblemaniac Ok I tested it as per your request. Everything is working fine. Although the Antialias checkbox doesn't seem to work. It appears that AA is on all the time, same with the Pressure checkbox, even if I disable it it both size and opacity will continue to be pressure sensitive. :+1:

scribblemaniac:
> Although the Antialias checkbox doesn't seem to work. It appears that AA is on all the time, same with the Pressure checkbox, even if I disable it it both size and opacity will continue to be pressure sensitive. 

All of these things *should* be fixed when this PR gets merged.

PR #803, opened 2017-11-21
Proper playback update on range controls change

AskePit:
Hello!
I'm a member of the Pinguem.ru competition on finding errors in open source projects.
I found an issue with Pencil playback where you can make playback state in out of sync with the range-playback controls. It is especially noticeable when you work with a sound. To reproduce the bug please follow steps:

- Create new sound layer with `Ctrl+Alt+W` shortcut;
- Add new frame by pushing `Keys: Add Frame (+)` button;
- Choose an audio file;
- Push `Loop` button (just for convinience);
- Push `Play` button;
- While playing try to trigger `Range` box or change `Strat of playback loop`/`End of playback loop` spinbox values — you'll see no difference in playback behaviour, it'll still play with old settings even after end of a loop iteration. But at the same time graphical range indication (yellow selection at the top of time scale) changes immediately, so we'll observe a lost of sync between graphical/controls indication and real behaviour. This pull request is intended to fix that.

Also there are additional errors/warnings found by PVS-Studio analyzer:

- [V591](https://www.viva64.com/en/w/v591/) Non-void function should return a value. actioncommands.cpp 471. This one caused crash on my side when I tried to add a frame. Very dangerous bug!

Don't know how to treat these warnings, but please take a look. There might be potential typos here.
- [V523](https://www.viva64.com/en/w/v523/) The 'then' statement is equivalent to the 'else' statement. vectorimage.cpp 2311
- [V686](https://www.viva64.com/en/w/v686/) A pattern was detected: (frame >= mEndFrame) || ((frame >= mEndFrame) && ...). The expression is excessive or contains a logical error. playbackmanager.cpp 73
- [V686](https://www.viva64.com/en/w/v686/) A pattern was detected: (currentFrame >= mEndFrame) || ((currentFrame >= mEndFrame) && ...). The expression is excessive or contains a logical error. playbackmanager.cpp 259

### DIFF
--- a/app/actioncommands.cpp
+++ b/app/actioncommands.cpp
@@ -467,6 +467,7 @@ Status ActionCommands::addNewKey()
         camLayer->LinearInterpolateTransform(cam);
         mEditor->view()->updateViewTransforms();
     }
+    return Status::OK;
 }
 
 void ActionCommands::removeKey()
--- a/core_lib/interface/timecontrols.cpp
+++ b/core_lib/interface/timecontrols.cpp
@@ -28,6 +28,7 @@ GNU General Public License for more details.
 #include "preferencemanager.h"
 #include "timeline.h"
 
+static auto spinBoxValueChanged = static_cast< void ( QSpinBox::* )( int ) >( &QSpinBox::valueChanged );
 
 TimeControls::TimeControls(TimeLine *parent ) : QToolBar( parent )
 {
@@ -106,7 +107,6 @@ TimeControls::TimeControls(TimeLine *parent ) : QToolBar( parent )
 
     makeConnections();
 
-    auto spinBoxValueChanged = static_cast< void ( QSpinBox::* )( int ) >( &QSpinBox::valueChanged );
     connect( mLoopStartSpinBox, spinBoxValueChanged, this, &TimeControls::preLoopStartClick );
     connect( mLoopEndSpinBox, spinBoxValueChanged, this, &TimeControls::loopEndClick );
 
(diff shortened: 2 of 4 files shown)


app/actioncommands.cpp
```diff
@@ -467,6 +467,7 @@ Status ActionCommands::addNewKey()
         camLayer->LinearInterpolateTransform(cam);
         mEditor->view()->updateViewTransforms();
     }
+    return Status::OK;
```
AskePit: Caused crash on my side!

Jose-Moreno:
@AskePit Thanks a lot Nikolai! Didn't know this competition existed. Good luck with it! If you want to find and fix errors, Pencil2D welcomes you to do so indeed! :smile: The lead dev is on a different timezone, so he will probably notice this PR in 10-12 hours. I hope you get points even if PR is pending. Best of luck with the competition, and thank you for helping out our project, we can use all the help that is needed! :+1:

CandyFace:
Ah yes that's correct, prior to this the timeline wouldn't update the ranged state unless you stopped the playback first.

Thanks for the contribution @AskePit

core_lib/interface/timecontrols.cpp
```diff
@@ -162,6 +162,9 @@ void TimeControls::setCore( Editor* editor )
 {
     Q_ASSERT( editor != nullptr );
     mEditor = editor;
+
+    connect( mLoopStartSpinBox, spinBoxValueChanged, mEditor->playback(), &PlaybackManager::setRangedStartFrame );
+    connect( mLoopEndSpinBox, spinBoxValueChanged, mEditor->playback(), &PlaybackManager::setRangedEndFrame );
```
chchwy: These 2 connections are actually duplicated. But I understand the current signal connections of Timeline is a mess. I think the new connections are more straightforward than the olds so I'll keep this and remove the existing one.

PR #812, opened 2017-11-27
Make our i18n setup play more nicely with Transifex

J5lx:
Changes:
- Remove unused Language.en.ts
  > **J5lx:** BTW, is it right to assume that pencil.ts contains the translation sources and Language.en.ts contains english translations?
  > **chchwy:** atm there is no such english translations
  > **J5lx:** Ah yes, Language.en.ts is not included in TRANSLATIONS in pencil.pro or in translations.qrc
  > **J5lx:** Is Language.en.ts basically useless then?
  > **chchwy:** yeah
- Use underscores rather than dashes to seperate languages from regions (e.g. hu_HU rather than hu-HU) in translation files. This is how both Qt and Transifex do it.
- Add configuration for Transifex client so translations can be pulled by running `tx pull`

While preparing these changes I noticed that I had accidentally added hu rather than hu_HU to the Transifex project. I now added hu_HU and transferred the translations to it but I haven’t been able to delete hu yet due to an error on Transifex’s side. Therefore we have to keep in mind not to commit the (now empty) hu translations to the repository.


chchwy:
Cool @J5lx Thanks

PR #813, opened 2017-11-27
Allow movie export from command line

J5lx:
Additionally, I
- made sure all CLI output goes directly to stdout rather than through qDebug() (since the latter can be disabled at build time)
- Put the logic for ignoring -NSDocumentRevisionsDebugMode into handleArguments()
- Shifted some code around so the program is only fully initialized once it’s actually needed (i.e. no need to have a complete MainWindow2 constructed (which takes time) if the user is just running `Pencil2d -h`)

### DIFF
--- a/app/main.cpp
+++ b/app/main.cpp
@@ -15,8 +15,6 @@ GNU General Public License for more details.
 
 */
 
-#include <iostream>
-#include <cstring>
 #include <QTranslator>
 #include <QLibraryInfo>
 #include <QCommandLineParser>
@@ -54,8 +52,10 @@ void installTranslator( PencilApplication& app )
     qDebug() << "Install translation = " << b;
 }
 
-int handleArguments( MainWindow2 & mainWindow )
+int handleArguments( PencilApplication& app )
 {
+    QTextStream out( stdout );
+    QTextStream err( stderr );
     QStringList args = PencilApplication::arguments();
     QString inputPath;
     QStringList outputPaths;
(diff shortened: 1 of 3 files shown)


app/main.cpp
```diff
@@ -54,26 +52,28 @@ void installTranslator( PencilApplication& app )
...
-    // TODO: Ignore -NSDocumentRevisionsDebugMode
+    args.removeOne("-NSDocumentRevisionsDebugMode");
 
     parser.setApplicationDescription( PencilApplication::tr("Pencil2D is an animation/drawing software for Mac OS X, Windows, and Linux. It lets you create traditional hand-drawn animation (cartoon) using both bitmap and vector graphics.") );
     parser.addHelpOption();
     parser.addVersionOption();
     parser.addPositionalArgument( "input", PencilApplication::tr( "Path to the input pencil file." ) );
 
-    QCommandLineOption exportSeqOption( QStringList() << "o" << "export-sequence",
+    QCommandLineOption exportOutOption( QStringList() << "o" << "export-sequence" << "export-movie",
```
scribblemaniac: Having --export-sequence and --export-movie do the same thing is a bit weird as it allows for things like: `--export-sequence test.mp4` or `--export-movie test.png`. We should either just use --export for the long form, or handle them separately.

core_lib/interface/editor.cpp
```diff
@@ -718,6 +722,37 @@ bool Editor::exportSeqCLI(QString filePath, QString format, int width, int heigh
     return true;
 }
 
+bool Editor::exportMovieCLI(QString filePath, int width, int height)
```
scribblemaniac: Exporting a movie through the GUI allows for range selection or running to the end of the sound layer. Exporting both image sequences and movies allows for camera layer selection. There should be optional arguments for all of these things. If we choose to use one single argument for both movie and image sequence export, then we should probably throw warnings for any options that will be ignored.

app/main.cpp
```diff
@@ -54,26 +52,28 @@ void installTranslator( PencilApplication& app )
     qDebug() << "Install translation = " << b;
 }
 
-int handleArguments( MainWindow2 & mainWindow )
+int handleArguments( PencilApplication& app )
 {
+    QTextStream out( stdout );
+    QTextStream err( stderr );
```
scribblemaniac: Do you think qWarning() and qCritical() would be better than QTextStreams? They exist for the specific function of printing warnings and errors, and can easily be controlled with defines, however we would have to use something something like qCritical().noQuote() or convert the QString to a const char * on each output to prevent it from automatically quoting the translated QStrings, which is quite ugly.

J5lx: To me it seems that all of these q*() functions are mainly for logging, not for cli output that is meant to be seen by the user. For example, take this from the documentation of qWarning() and qCritical():

> Under Windows, the message is sent to the debugger. On QNX the message is sent to slogger2.

Imo cli output is not something that should be sent to debuggers or system logging daemons under any circumstances. And ftr, QCommandLineParser also prints directly to stdout/stderr (using fputs) on all systems except WinRT and possibly Windows.

app/main.cpp
J5lx: Fair point. I’ll make --export the primary long form and turn --export-sequence into a hidden option for bc.

app/main.cpp
scribblemaniac: > Imo cli output is not something that should be sent to debuggers or system logging daemons under any circumstances

I agree. I did not realize it did that for non-qDebug() functions. We can leave it as is then.

core_lib/interface/editor.cpp
J5lx: Yeah, good point here too. I just blindly followed what the image sequence exporter was doing without thinking too much about it, but these are indeed reasonable enhancements. I’ll take care of it asap.

chchwy:
Good work @J5lx!  and thanks for your reviewing @scribblemaniac

Issue #816, opened 2017-11-29
[BUG] Color Wheel HSV Mode input problem for slider values


Jose-Moreno:
### --Issue Summary--
The problem is that when trying to input values on the property boxes for the HSV mode it happens the same thing that it used to with the RGB sliders, where the values get mixed up and changed without notice.

When you select the value and write a new one, the numbers are kind of "lagging" or other times it will input wrong values. For example writing 52% will input 51% always changing the 2 for a 1. Same with the 8, it will always show a 7.

Write 48 and you'll get 37, and when using backspace to delete the 7, the 3 will round down to a 2.

For a UX issue on the RGB mode sliders please review the follow-up comment below the main issue.

### --Expected Results--
It should work in the same way as the RGB boxes at least. Where you can:
a) click on a box and type in a number to replace the current value
b) not have the current numerical value changed without user input.

### --Steps to reproduce--
1) Open Pencil2D
2) Select the HSV color mode radial button on the color wheel
3) Click on any property (hue, saturation, value, alpha)
4) Press backspace to delete the whole number
5) try the following cases:
a) Write the number 100. Notice how the 1 turns into a 0 after the first digit. If you press 1 again it will write itself correctly, then if you press 0 it will get to 100.
b) Write the number 57. Delete the 7. Write the number 58. Notice the no. 8 is consistently changed into a 7.
c) Write the number 48. You will get no. 37. Delete the 7. The 3 is changed to the number 2.
d) Write the number 72. you will get 62. Delete the 2. The 6 will be changed to a 5.

### --System Information--
+ Pencil2D Version:
Version: 0.6.0 (Nov 28th 2017 Nightly Build)

+ Operating System:
Windows 7

Jose-Moreno:
As an **additional issue** which is more related to the **user experience**, on the RGB mode sliders, when you are writing values, the cursor will fall back after writing the second digit, instead of going forward. So if you try to write a 3 digit number like 210, you end up with the number 201. 

In such case you have to either notice what you are writing or are forced to delete the number and try again.

CandyFace:
Should be fixed in #819

chchwy:
@Jose-Moreno Please check this in nightly build 3 Dec 2017.

Jose-Moreno:
@CandyFace @chchwy thanks, will do!

Jose-Moreno:
I can confirm it works on windows so far. Thanks a lot!

PR #819, opened 2017-12-01
Fixed colorinspector spinboxes behaving unusually

CandyFace:
#816 
Additionally I've also added a degree symbol to the hue spinbox.

### DIFF
--- a/app/colorbox.cpp
+++ b/app/colorbox.cpp
@@ -69,7 +69,7 @@ QColor ColorBox::color()
 
 void ColorBox::setColor(const QColor& newColor)
 {
-    if ( newColor != mColorWheel->color() )
+    if ( newColor.toHsv() != mColorWheel->color() )
     {
         mColorWheel->setColor(newColor);
         mColorInspector->setColor(newColor);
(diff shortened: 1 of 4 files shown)


chchwy:
Thanks @CandyFace  it did resolve the issue.

app/colorbox.cpp
```diff
@@ -69,7 +69,7 @@ QColor ColorBox::color()
 
 void ColorBox::setColor(const QColor& newColor)
 {
-    if ( newColor != mColorWheel->color() )
+    if ( newColor.toHsv() != mColorWheel->color() )
```
chchwy: Just a question. What is the difference between of this colour comparison with and without `toHsv()`?

CandyFace: When you choose RGB in the Pencil2D UI, newColor will be based on RGB colorspace while mColorWheel will always be in HSV colorspace. This means that it will always be false whenever you compare against two different color spaces. Therefore to avoid that I've added the .ToHsv, so we can be sure that it's always checking correctly.

PR #820, opened 2017-12-02
Sync master with release

scribblemaniac:
Merge extra changes in the release branch into master and use conditional statements to change the build settings on the release branch instead of always having differing branches.  This should also be merged into the release branch eventually so they are the same. Theoretically my changes will not affect anything on either the master or release branches.


chchwy:
Thanks @scribblemaniac

Issue #801, opened 2017-11-16
Add support for custom themes


ElectricVersion:
Pretty self explanatory - not a huge priority but allowing some customization of the UI would probably be an appealing feature for some users. If everyone is on board with this idea, I'll assign myself and begin working on it :)

I'm thinking we could make it so people can add CSS stylesheets or something, but if anyone has a better idea for how we could implement it I'm open to that of course.

CandyFace:
Related to #550 
I'm not against UI customization but Imo. just having a themes with light and dark or custom color would be enough.

When that's said though, if you want to implement it, feel free to do it ;)
Qt has its own styling syntax which you can use, it's very similar to standard css.

J5lx:
Just make sure that native themes (e.g. GTK+ themes picked up by Qt’s GTK+ style) keep working. For me it’s more important for a program to fit in with the rest of the system than having its very own unique look and feel (And for reference, custom CSS in Pencil2D was my primary reason for opening #550). However as long as that remains possible, go for it :+1:

PR #667, opened 2017-05-08
New toolbar icons

CandyFace:
I consider #603 complete.

Showcase:
screen shot 2017-05-08 at 15 35 44

The png icons has been placed in icons/new and svg in icons/new/svg.
This does not make them appear in the application yet, it's only a PR for the resource files.

All icons except (arrow, select and polyline) come in two variations, a flat and a detailed one.

Issue #603, opened 2017-02-12
Toolbar icon design


CandyFace:
Hi

Looking at issue #550 made me think that instead of fixing the icons, it would be better to give the toolbar a makeover. There are lots of decisions to be made when designing icons, so instead of me just bashing pixels without any consideration in what the rest of the pencil community thinks, I've decided to share my thoughts and progress.

I've made a bunch of **sketches** already, mainly to figure out the placement and how the icons should be illustrated. Nothing is finalized though, so try not to think too much about the style, color or shading. 

Rules i've come up with so far:
- Style should be consistent across all icons, if possible
- icons should be designed in size 24x24.
- Icons should be designed in a vector program or be convertible from raster to vector.

### Iterations:
![vector](https://cloud.githubusercontent.com/assets/1045397/22863977/3844f006-f149-11e6-895b-4048ba03dcc4.png)

These represent the look as you would see them in the program ie. in pixels
![pixel](https://cloud.githubusercontent.com/assets/1045397/22863978/386231a2-f149-11e6-9aa6-5ad0a261fefc.png)

100% scale:
![100%](https://cloud.githubusercontent.com/assets/1045397/22864003/7fe5417c-f149-11e6-80bf-49df843a510b.png)

I prefer the top right icon, as it is displayed in a way similar to the old icon but with more emphasis on tip.

### Mockup:
![toolbar](https://cloud.githubusercontent.com/assets/1045397/22863548/96c583d6-f142-11e6-82e8-3fa4c6cb46b1.png)

I'll keep updating the thread, showing more icons as I get them done. If no one disagree or comment, then I'll just go with the design I've mentioned already (the top right).

Jose-Moreno:
@CandyFace due to this issue #597 I was also thinking of moving to a stage where we can begin changing Pencil2D's overall graphic design at least regarding it's icons. 

As you can read on the latest response by scribblemaniac: https://github.com/pencil2d/pencil/issues/597#issuecomment-279237490

he says there shouldn't be a problem with having SVG icons displayed. I honestly believe it would be for the best to have the icons as vector files to improve readability and help with HiDPI screen users.

Regarding the icons themselves I think you've taken a great first step. I agree with your rules as well, however considering SVG icons must also become a factor while designing them since the size limit shouldn't be an issue. 

[...]

PR #821, opened 2017-12-02
Update Toolbar Icons

scribblemaniac:
Now that we've got v0.6.0 out of the way, we can use the new icon set made by @CandyFace in #667.

Related issues: #603, #550, and #801.


J5lx:
Besides, while not related to this PR, the tip of the pencil seems to vanish on the dark background of my theme:

screenshot from 2017-12-02 13-41-57

@CandyFace

CandyFace:
Looking good @scribblemaniac As far as I can see though, this will require manual tweaking. The SVG conversion to SVG tiny 1.2 has made some crude changes here and there. For a small icon, 32x, 64x, it might work but if we are to have bigger icons, the files will have to be tweaked so they look as close to the original SVG files.

@J5lx That's going to be difficult to fix, given that it's an issue caused by the color choice of your theme. I'll see what I can without making too drastic changes.

scribblemaniac:
@J5lx Fixed. Mac really hates files with the same case.

For reference purposes, I had to convert to SVG Tiny 1.2 because otherwise they look like this:
screen shot 2017-12-02 at 9 56 38 am

chchwy:
@scribblemaniac Are you going to update the icons in Tools menu as well?

CandyFace:
ehm.. whoops O_o Why did it even let me commit directly to your branch. Oh well XD @scribblemaniac  I've committed some changes to your branch, adding the icons to tool menu and tweaked some icons as I mentioned

J5lx:
Thanks, it’s a little better now:
screenshot from 2017-12-04 20-10-09
Still not optimal, but at least it doesn’t feel completely invisible now. Let’s keep it like that for now and maybe we can talk about theme-centered icon improvements some more in #550 eventually.

Also, good thing you added them to the menu, I had already forgotten completely about that myself :p

Other than that, I noticed that I now get a bunch of these in the terminal:

    qt.svg: link _Image4 hasn't been detected!
    qt.svg: link _Image4 hasn't been detected!
    …
    qt.svg: link _Image4 hasn't been detected!
    qt.svg: link _Image4 hasn't been detected!
    …
    qt.svg: link _Image4 hasn't been detected!
    …

I have absolutely no idea what that means and the interface seems fine but it doesn’t sound positive :p Do you two have any idea about this?

CandyFace:
I found some image4 reference in one of the svg files, which I've now removed. I didn't notice any changes in the svg itself, so I think it's fine. The warnings are gone now at least.

J5lx:
Actually I wasn’t even referring to theme-specific icons but yes, that would also be an interesting thing of course :). Most headache-free route might be to have some generic monochromatic icon sets (light and dark) available so there’s less potential colour conflicts to worry about.

Issue #811, opened 2017-11-27
[Documentation Request] Create instructions for building Linux AppImages


scribblemaniac:
We have instructions for building on Linux in `docs/build_linux.md` that still work as far as I know, however we need to add a section for building AppImages, which has some tricky steps. Additionally, the process is more complex for non-x86_64 architectures, as some of the tools have to be built from source.

probonopd:
(fyi) [Open Build Service](https://github.com/AppImage/AppImageKit/wiki/Using-Open-Build-Service) has the tools for 32-bit and 64-bit, Intel and ARM.

J5lx:
We’ll have to see whether maintaining a setup for yet another service (in addition to AppVeyor and Travis) is worth the effort for a small project like ours, but from a quick look some of the features seem really interesting for builds used by end-users (dependencies staying the same and such). Thanks for the tip!

PR #823, opened 2017-12-02
Make it pass desktop-file-validate on Ubuntu trusty

probonopd:
Reference:
https://github.com/AppImage/appimage.github.io/pull/194

### DIFF
--- a/pencil2d.desktop
+++ b/pencil2d.desktop
@@ -1,5 +1,5 @@
 [Desktop Entry]
-Version=1.1
+Version=1.0
 Type=Application
 Name=Pencil2D
 GenericName=Animation Software


J5lx:
How does the changed version affect newer DEs/WMs? If it makes them ignore the Keywords key or something, I’m in favour of keeping version 1.1 in the file and adjusting it on-the-fly during the AppImage build process (or maybe offering a pre-made alternative desktop file for AppImages). (Especially since the desktop entry is also installed during `make install` and thus not exclusive to AppImages.) Thoughts, @scribblemaniac, @probonopd?

probonopd:
Since AppImages are generally supposed to also run on the oldest still-supported LTS distributions, the most recent version of the XDG desktop file spec you can use at this time is 1.0. This does not affect the functionality in any way on newer DEs/WMs, which are fully backward compatible.

scribblemaniac:
If this was all that needed to be changed, I might be in favor of it, however I know that the old desktop-file-validate is not going to pass this just by changing the version number. The "Keyword" key is invalid for 1.0 desktop files and would have to be removed completely, obviously breaking the keyword feature used for things like searches. So, I suppose it depends on exactly what happens when 14.04 users try to use our 1.1 desktop file. If it works fine and throws some warning, then leave it as is, if it crashes the program then maybe change it.

chchwy:
@J5lx how will you do the on-the-fly desktop file adjusting?

J5lx:
Using a `sed` line or two in .travis.yml :) Come to think about it, maybe we could also abstract the AppImage build process into its own shell script considering #811.

J5lx:
Alright then, I’ll come up with a proper solution asap.

J5lx:
@probonopd An alternative solution to this problem was recently merged in #871. Could you verify whether it is sufficient?

probonopd:
> Could you verify whether it is sufficient?

I can do that as soon there is a release available that contains this fix.

J5lx:
@probonopd We just released 0.6.1, containing the fix: https://github.com/pencil2d/pencil/releases/download/v0.6.1/pencil2d-linux-amd64-0.6.1.AppImage

probonopd:
Thank you.

PR #827, opened 2017-12-04
Re-added old zoom levels

CandyFace:
extending the implementation of fixed zoom levels in https://github.com/pencil2d/pencil/commit/0cadf70d2b97590185e570cedf3b330dba6c7c0b for zoom shortcuts.

### DIFF
--- a/core_lib/managers/viewmanager.h
+++ b/core_lib/managers/viewmanager.h
@@ -95,9 +95,14 @@ private:
     bool mIsFlipHorizontal = false;
     bool mIsFlipVertical = false;
 
-    const float mZoomLevels[8] = { 0.25, 0.5, 0.75, 1.0,
-                                   1.25, 1.5, 1.75, 2.0 };
-    const int mZoomLevelCount = 8;
+    const QVector<float> mZoomLevels = QVector<float>() << .01f << .02f << .04f
+                                                        << .06f << .08f << .12f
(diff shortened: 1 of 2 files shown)


core_lib/managers/viewmanager.h
```diff
@@ -95,9 +95,14 @@ class ViewManager : public BaseManager
...
-                                   1.25, 1.5, 1.75, 2.0 };
-    const int mZoomLevelCount = 8;
+    const QVector<float> mZoomLevels = QVector<float>() << .01f << .02f << .04f
+                                                        << .06f << .08f << .12f
+                                                        << .16f << .25f << .33f
+                                                        << .5f << .75f << 1.f
+                                                        << 1.5f << 2.f << 3.f
+                                                        << 4.f << 5.f << 6.f
+                                                        << 8.f << 16.f << 32.f
+                                                        << 48.f << 64.f << 96.f;
```
chchwy: Personally, I will suggest using std::vector or std::array instead of QVector. Usually has better support from compiler & debugger.

chchwy:
Thanks for bringing it back @CandyFace

core_lib/managers/viewmanager.h
CandyFace: I did actually use std::vector but then thought about doing it the Qt way and therefore used QVector but if you think it's better to use std, you're welcome to change it :)

Issue #681, opened 2017-05-31
Movie Exporter Rewrite


J5lx:
This came up in #671 a few weeks ago and I figured it deserved its own issue by now. I’m going to go into details, so this will be long; the tl;dr is that the current movie exporter based on the FFmpeg CLI is inelegant and inefficient and it would be better to use something based on an API such as the FFmpeg API.

### Current state of affairs

The current exporter employs a mixed bag of Qt APIs, custom multimedia code and calls to the FFmpeg CLI to export a project in up to 5 (five!) passes, with the exact number depending mainly on the file format:

- if MP4: (re-)encode and save all sound clips as WAV files
- if MP4: mix sound clips and save as WAV file
- encode and save all frames as PNG files
- encode as H.264 and AAC and mux both into an MP4 file
- if MP4: re-encode as H.264 and AAC and mux both into the final MP4 output file
- if GIF: Generate palette
- if GIF: Encode and save as final GIF file

Needless to say, this procedure is far from ideal. It relies on heaps of temporary files (consuming a good amount of memory or disk space, depending on OS and setup), degrades quality due to using lossy codecs for intermediate results (using lossless codecs instead would make resource consumption much worse) and provides more surface area for bugs due to the custom multimedia code that is not even relevant to Pencil’s core functionality. Instead of relying on a CLI which was not conceived for programmatic use, it would be better to use an API. The most obvious choice is the one provided by FFmpeg since it is the de-facto standard for Open Source multimedia handling, it is available on all platforms targeted by Pencil and we are already using it indirectly right now.

### New exporter based on FFmpeg API

As mentioned before I have already started experimenting with a new exporter using the FFmpeg API. My current progress can be found on [my export-lav branch](https://github.com/J5lx/pencil/tree/export-lav/) and nightly builds are available from <https://j5lx.de/pencil-builds-j5lx>. To enable the exporter, CONFIG+=EXPORT_LAV must be passed to qmake and it must know where to find the libraries.

- On Linux it looks for the pkg-config packages, so those need to be available. On Arch Linux they are in the ffmpeg package, on a modern Ubuntu they are in libavformat-dev, libavcodec-dev, libavutil-dev, libswresample-dev and libswscale-dev.
- On macOS it also looks for the pkg-config package. An easy way to obtain it is via Homebrew (i.e. `brew install ffmpeg`). When building with the homebrew package, the PKG_CONFIG_PATH environment variable must contain /usr/local/lib/pkgconfig.
[...]

PR #835, opened 2017-12-13
Improve GIF export quality by removing first pass conversion to mp4

### LINKED ISSUE #818: GIF Export results in video-like artifacts
I've noticed that exporting a gif seems to convert the image to an mp4, and THEN to a gif. This is probably the cause of the image artifacts. I'm not sure if it's possible to export images directly to a gif with FFMPEG. If not, perhaps we could use MagickWand, GIFLIB, or something else for gif export.

scribblemaniac:
I was content to wait for #681, however since there were already attempts to improve the current exporting lately (ac5e19f, 17bff13), I thought I would take a moment to finish this up. GIF export is now greatly improved, no more video artifacts. Based on my tests, it is visually identical to ImageMagick's GIF export and looks about as good as GIFs possibly can.

Closes #818

### DIFF
--- a/core_lib/movieexporter.cpp
+++ b/core_lib/movieexporter.cpp
@@ -172,7 +172,14 @@ Status MovieExporter::run(const Object* obj,
     STATUS_CHECK(generateImageSequence(obj, progress));
     progress(0.9f);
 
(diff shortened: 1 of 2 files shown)


chchwy:
Well done @scribblemaniac !

PR #836, opened 2017-12-14
properly downscale canvas when zooming out

CandyFace:
This will properly downscale the canvas when the view is below 100%
### before:
wip

### after:
wip2

I've been trying to get it to work when zooming in too, but the way the renderer currently works makes it impossible as the canvas has to be upscaled a lot and qimage is not made for that.

When we migrate to myPaintLib, I will look into this again, because its renderer splits the canvas up into smaller sections and therefore should provide the solution we need to make that work.
This will additionally also make it possible for us to prescale the brush stroke, so the representation will have the same look.

### DIFF
--- a/core_lib/canvasrenderer.cpp
+++ b/core_lib/canvasrenderer.cpp
@@ -23,6 +23,7 @@ GNU General Public License for more details.
 #include "layercamera.h"
 #include "vectorimage.h"
 #include "util.h"
+#include "viewmanager.h"
 
 
 
@@ -69,7 +70,7 @@ void CanvasRenderer::ignoreTransformedSelection()
     mRenderTransform = false;
 }
 
-void CanvasRenderer::paint( Object* object, int layer, int frame, QRect rect )
+void CanvasRenderer::paint(Object* object, int layer, int frame, QRect rect )
 {
     Q_ASSERT( object );
     mObject = object;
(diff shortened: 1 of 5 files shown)


scribblemaniac:
Have you done any tests on the performance impact of prescale? Also, can you squash your commits please.

CandyFace:
I have made some ElapsedTimer estimations but nothing beyond that
average:
prescale: 314237.438931 nsencs
normal:   41230.8219178 nsecs

so yes it's indeed a bit slower but it's a necessary evil imo. just like it will be to prescale the brush stroke when painting and when zooming in, to get a better and more accurate result of the image.

chchwy:
Overall it's good. Thanks @CandyFace 

For better architecture, I will remove view manger in CanvasRenderer later. The CanvasRender is supposed to work without knowing anything except Object. And in the future, the Preview windows will also use CanvasRenderer and definitely has another scale value from main canvas. I'll do that, just let you know. Good job!

CandyFace:
Ah yes you're right about that... I always forget that we have certain way of doing things and what should have acces to what. Where can I find the schematic for the structure? This is the closest one I've been able to find but it doesn't have canvasRenderer and other things. It would be nice with a updated and make it easy to find. Ideally we should probably have a "HOW WE DO THINGS" readme file in the same way as Krita's "HACKING" https://github.com/KDE/krita/blob/master/HACKING

chchwy:
There is a noticeable performance degradation when the zoom is about 15% or less. Probably worth an investigation.

CandyFace:
Hmm I never noticed any of this D: perhaps because I have enough ram to allow it. Maybe we should remove these changes again then, given that both down scaling and up scaling requires this much with the current painter implementation. I'm not sure we can do anything else for now, but I'll take a look at it again when I get home.

Issue #578, opened 2016-12-01
[Proposal] Replace Bitmap Layers


scribblemaniac:
I had this idea a long time ago, and I was reminded of it recently upon experimenting with the mypaint branch (excellent work btw @feeef). One of the current issues with it is that the bitmap layers are loosing quality when drawing on a transformed canvas. Imagine this simple scenario:
1. The drawing canvas is scaled to half size.
2. A line is drawn.
3. The image is rendered.

The result of this will be a line that has half the resolution that it should. This can cause issues with antialiasing, as well as with the quality of the new brushes. This situation is made much worse by the fact that everything is drawn onto the transformed bitmap, which means that to save the frame, you have to transform it back to a zoom and rotation of zero, and subsequently retransform it back to the canvas transformation. What this all means from an animators perspective is, if they want to scale, the quality of their drawings is reduced, and if they want rotation, the quality of past strokes degrades with each new stroke you add. Now I can probably think of three or four different solutions to this issue, but I believe this solution provides the most benefit (but admittedly a non-trivial amount of work).

Okay, so my proposal is simple in principle. We disable painting on bitmap layers (but keep the layer type for backwards-compatibility and image import purposes). Next we add a 'stroked' feature to the vector layer (or as it's own layer type). When it is enabled, instead of painting vector paths to the screen it draws brush strokes along the vectors to a bitmap image, which is then cached and drawn to the screen. When a new brush stroke is started, it can be handled as bitmap manipulations with the vector data being stored in the background. When the stroke is completed then the vector data can be added. To a user, this would appear nearly identically to painting on a bitmap layer.

Advantages:
- Solves mypaint issue mentioned in the beginning. Overall this change should integrate well with the qmypaint.
- All brush strokes (including with mypaint brushes) made after this change can be upscaled to any resolution without losing quality
- Individual brush strokes can be selected, moved, scaled, rotated, duplicated, and deleted. Brush widths, colors, feathering, stroke smoothness and brush types can also be changed after the fact.
- Virtually no impact on drawing speed since active brush strokes will remain a bitmap operation
- Unless you have a massive amount of brush strokes or an extremely small canvas, file sizes can be drastically reduced
- Generating old pcl format files is simple
- We don't have to support two separate types of drawing

Disadvantages:
- It takes effort to make these changes, which is a finite resource for this project
- Additional code complexity
[...]

PR #837, opened 2017-12-14
[WIP] Modify the bitmap bucket flood fill algorithm

scribblemaniac:
This makes some significant changes to the flood fill algorithm of the Bucket tool. Here's my best summary of the changes and why they're needed.

### Color difference
Instead of checking if any one channel has a difference in color greater than the threshold, we check if the Euclidean distance between the colors (excluding the alpha channel) is less than the threshold. This means if you have two colors that are 14 values apart on the red, blue, and green channels, then with a threshold of 15 it will be over the threshold because the Euclidean distance is ~26. Differences in transparency are immediately assumed to be different colors, primarily because a transparency of 0 is the same color regardless of the other channels. There is one exception to this which I will talk about shortly.

### Speed optimization
Inspired by how Krita does flood filling, I have added a hash map to remember if colors are within the threshold for the duration of the operation (threshold and oldColor will remain constant, so only newColor is needed for lookup). Since the vast majority of the pixels that will be encountered are usually from a small subset of colors, this should have significant speed benefits. The Euclidean difference operation is probably a bit slower than the old difference operation, so this is even more important.

### Limit fill boundaries
There is currently an issue in Pencil2D where filling outside of the camera area creates some weird behavior. To fix this, here is the behavior I propose (and implemented already 😝 ). If the place you click is outside the union of the the current camera view bounds and active layer keyframe bounds, then do nothing. Otherwise, expand the image size to the union bounds, and then apply the flood fill at the location pressed to the entire layer keyframe image.

### Different behavior for filling fully transparent areas
This is the only change I made that could really be contentious. I am not aware of any applications that implement bucket fill this way, but I think this is the best way for us to do it, specifically for the workflows that Pencil2D animators will typically be following.

Basically, when you try to fill a transparent area, it will assume that all transparent *and translucent* pixels are below the threshold. In addition to this change, instead of composting the filled color pixmap over top of the original image, it is composited *underneath*. There are two main user-noticeable differences as a result of these changes:
1. Antialiased lines are filled smoothly. This is the desired effect and the whole reason for making this change. Instead of the fill stopping at the antialiased lines, it goes under them. If for instance you drew an aliased or feathered black line and then flood filled the background with black, an outline would remain filled with translucent black pixels normally. With this new technique it would just be a fully black canvas. Here's a demonstration the difference:

| Unfilled | Old Fill Algorithm | New Fill Algorithm|
|---|---|---|
| example unfilled | example old filled | example new filled |

2. Now for the catch (maybe). If there are solid, translucent lines, they will not stop the flood fill. However I don't think this is a significant issue since translucent lines are hardly every used in animation. Far more frequently you seen antialiased/feathered opaque lines than translucent lines. Slightly more likely, two lines may only be connected by feathering, in which case the will not form a "closed" segment as far as the flood fill is concerned. Depending on the situation, I could see this as either a feature or a bug.

I think that the benefits of this new transparent fill algorithm outweighs its drawbacks, especially for traditional animation and the typical "cartoon style".

### DIFF
--- a/core_lib/graphics/bitmap/bitmapimage.cpp
+++ b/core_lib/graphics/bitmap/bitmapimage.cpp
@@ -18,6 +18,8 @@ GNU General Public License for more details.
 
 #include <cmath>
 #include <QDebug>
+#include <QtMath>
+#include <QScopedPointer>
 #include "util.h"
 
 BitmapImage::BitmapImage()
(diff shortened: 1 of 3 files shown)


CandyFace:
I like what I'm seeing, I however don't think it should be a fixed value that you can't fill inside translucent lines because that would make the fill tool useless for anything except pen tool. unless you intend to 

If anything I think it should be an option that you can skip translucent lines and we should look into a anti-aliased fill solution for filling feathered or heavily sampled strokes.

scribblemaniac:
> because that would make the fill tool useless for anything except pen tool.

I think you mean it would make the fill tool useless only for the pen tool with a translucent color. This is the best solution I can think of to fill feathered/aliased areas. A tool option to enable or disable this behavior might be okay. There may be some way to detect translucent lines and not go over them, however I think that would be far more work than it's worth.

CandyFace:
Yes you're right, I meant for translucent lines.

I disagree though, the flood fill tool is a vital feature that should be taken with utmost care to make it work in the best possible way and with all kinds of scenarios.

> This is the best solution I can think of to fill feathered/aliased areas

but this will only fill properly if the background is transparent, when the user tries to fill an anti-aliased stroke on a colored background, the result will still be aliased.

scribblemaniac:
> but this will only fill properly if the background is transparent, when the user tries to fill a feathered stroke on a colored background, the result will still be aliased.

True, but that's still one situation more than before where the output doesn't look like 💩  

I won't claim that this implementation is perfect, the fact of the matter is that few if any of our tools are. I'm open to suggestions on how to improve the algorithm to tackle some of these other issues. The composition under technique used here will obviously not work for non-transparent areas without some other preprocessing steps.

scribblemaniac:
Oh, it's also worth noting that if we go through with #578, that will basically eliminate the need for any of this code to handle flood filling of anti-aliased/feathered edges since flood fill will be based on the path edge, not the rendered edge.

feeef:
Thanks for the great improvement @scribblemaniac !

Also, I think that a good way to colorize bitmap animations is to use a layer dedicated to colors under the one dedicated to line art. This way, you can tweak you colors without affecting the lines and vice-versa. Pencil2D used to have that feature where the flood fill tool would paint on the bitmap layer above (if present). This was kind of confusing for new users but it is how most digital painters handle their coloring by isolating their line art. I think that it is good practice. TVPaint works this way too.

This feature seems to have disappeared now and the color is directly applied on the line art layer. It is a shame as isolating colors from lines is now pretty difficult to do. Unless I missed something?

Jose-Moreno:
@feeef No, you're right, that behaviour was removed. I argued at first that it should stay, but in hindsight it's better that it was removed for now, so we can have the timeline rewrite + mypaint integration in place to re-implement it "properly".

Besides someone else asked for a similar feature like what toonboom uses, which is the concept of sub-layers. This of course should be thoroughly discussed but I believe we first have to improve all the fundamental systems before trying to get advanced functionality like that working properly not only related to it's function but also to the user experience and how it's easier for the user to understand and use said feature.

CandyFace:
I agree that colouring and lineart should be separated but not automatically. The user should have full control over where and how to fill something. Instead of trying to make a "one size fits all" solution where we try to justify the need to do it this or that way, we should try to make the tools work in various situations. 

If the user want to make their life a bit easier by filling what they can first and then manually paint the rest, then the tool should be able to do that. You also have to think about the future. 
For example a lot of the brushes in Krita and mypaint rely of translucent lines in one way or another. Do we really want to abandon the ability to fill within boundaries in all those cases?

Another thing: if we do it this way, then the result should at least be consistent. What is the purpose of being able to fill beneath translucent lines if it only works for cases with no background? 
The problem with this approach is that it will most likely be encountered as a bug because it doesn't happen in any other case.

scribblemaniac:
An option to sample the whole image rather than just the active layer would be a fairly simple way to provide the option of separating the bucket filling on a lower layer. That could however add potential confusion if you're trying to fill something that is drawn on a layer above the active layer. These changes were focused on transparency specifically with the idea of eventually being able to add the fill on a separate layer, so areas could be filled and refilled very easily.

The purpose of being able to fill only in the case where there is no background is because it is much easier to program 😁 . Seriously though, with opaque boundaries there is much less data to work with so at best the program can perform a "best guess" which could also have seemingly inconsistent behavior that the average user could consider a bug.

I have noticed that the pencil tool doesn't always draw lines with an opaque center either, so I am reconsidering the importance of dealing with that case. I will put some thought into a better algorithm.

feeef:
@CandyFace I agree that the user should be able to have control over where to paint. It is why he should also be able to paint on separate layer.

What about the ability to select the layer you want to paint to? It could be a bucket toggle button on each bitmap layer, that would be grayed out by default. While no layer has been selected as bucket target, the bucket tool paints on the active layer (just like currently). If a layer is selected as color layer, the bucket tool, paints on the first active color layer under it (or itself if active color layers are above).

The bucket tool would always apply its fill algorithm on a merge image from the visible layers.

Just a proposal...

scribblemaniac:
I'm currently in the process of refactoring the useful code out of the code from this PR, breaking down commits by feature (as I should have done to begin with). The `Limit Fill Boundaries` section has been merged with ccb18639875964b60470925cfa6c3448eb076400. `Color Difference` and `Speed Optimization` will hopefully be arriving as new separate PRs soon. Discussion for an improved fill algorithm will come eventually.

PR #839, opened 2017-12-19
Fix ram consumption and lag when zooming out 

CandyFace:
The problem occurred here https://github.com/pencil2d/pencil/commit/fac2c762c8ecaf4246369fc851c9a93bdce21271
when my transform was accidentally reversed :P 

So basically what caused the lag here is that the canvas got upscaled A LOT and then downscaled again (by the painter) to get a good subsampled picture. This is not a problem when the image is small but when it becomes 50x larger then it gets hairy because it's trying to allocate a huge image in memory (in case of 96x scale) that has to be made for each time it's being asked to zoom the canvas.


chchwy:
yeah that's my mistake. Thank you @CandyFace to find it out.

PR #848, opened 2018-01-06
[WIP] Create NSIS installer script for Windows

J5lx:
Now that I'm going to get busier again, I'd like to get some of the working stuff out that I've been working on lately, starting with this. As I already mentioned on IRC, I dug up some of my old [NSIS](http://nsis.sourceforge.net/) stuff and adapted it to create a Windows installer for Pencil2D. Right now, there are some things that could possibly be added (checking for an existing installation, for instance), but the actual (un)installation routine seems to be working splendidly at this point. The only thing that definitely needs improvement before this can reasonably be released is the graphics; I'm by no means a professional designer and I'd say it shows :) While I'm at it, time for some screenshots:

- Language selection:
  muilangdll
- Installer welcome page:
  inwelcome
- Installer license confirmation:
  inlicense
- Installer mode selection:
  inmode
- Installer destination selection:
  indest
- Installer start menu folder selection:
  instartmenu - copy
- Installer progress page:
  inprogress
- Installer finish page:
  infinish
- Look, we even got ourselves an uninstall entry, complete with meta information:
  unentry
- Uninstaller welcome page:
  unwelcome
- Uninstaller confirmation page:
  unconfirm
- Uninstaller progress page:
  unprogress
- Uninstaller finish page:
  unfinish

If anyone wants to try it themselves, look here: <https://drive.google.com/open?id=1vR04yL4LUrbvPwRFPi1ajS7Una0CN01s>. Please note that the contents of these "demo" installers are slightly different from the current releases, most notably I was too lazy to download the FFmpeg binaries.

To create the installer, installer.nsi needs to be processed by makensis like this:

- Windows:

      makensis /DBITS=32|64 /DVERSION=1.2.3 /DINSTALLFILES=path\to\directory\to\be\installed\relative\to\installer.nsi path\to\util\installer\installer.nsi
- *nix (for cross-compilation etc.)

      makensis -DBITS=32|64 -DVERSION=1.2.3 -DINSTALLFILES=path/to/directory/to/be/installed/relative/to/installer.nsi path/to/util/installer/installer.nsi

Due to the NSIS compiler treating all paths as relative to the main script, the resulting installer can then be found in the same directory as installer.nsi.

As I've previously said, I consider the installer itself pretty much complete, but the two thinks I'd like to get ready before merging are:
- [ ] Build it automatically with AppVeyor (NSIS comes pre-installed on AppVeyor, so this should basically just be some manual labour)
- [ ] Make the icon a little cleaner. It currently looks kind of not-so-pretty at small sizes, this is probably solvable by embedding more pre-rendered sizes in the .ico file, I'll just have to recall how to do that again.
- [ ] Make some proper images for the welcome and finish pages. In the simplest case we could just make those nothing but the logo itself, but to give them their own individual touch I could use a little help. I'd appreciate opinions on this.

Let me know what you think, be it about the whole thing overall or about some of the individual aspects!


Kaikogf:
Great !

Jose-Moreno:
@J5lx Well that's cool! thank you for the hard work :smile: , but I'd like to ask, what are exactly the advantages of having and installer as opposed to have a "portable" version?

J5lx:
Maybe it’s just me, but over time I got the impression that a) there are a lot more non-technical people that it often seems and b) many of these non-technical people have no idea how to deal with any sort of download that is not itself a document or an executable ­– including ZIP files – because they usually don’t come in contact with them. The fact that Windows calls ZIP files compressed folders doesn’t make it any better since it leads people into thinking there’s absolutely no need to extract them entirely, even when Windows tells them otherwise. Actually it seems that an awful lot of the comments on our [download page](http://pencil2d.github.io/download/) are somehow related to ZIP file extraction (people extracting only some of the files or not knowing how to deal with the archive altogether, etc). Speaking of it, there are also some comments about missing MSVC runtime libraries, maybe I can find a way to detect whether the needed MSVC runtime is already installed and otherwise ask the users whether they want to install it too.

tl;dr; I’m pretty confident that an installer is a good idea, but please let me know if you have any concerns. Of course this PR doesn’t mean we should stop providing the portable ZIP files altogether, it’s more about offering an option that is more straightforward for less technical users.

Jose-Moreno:
@J5lx You're right, in that regard an installer would be great for our Windows non-technical users (mostly kids), what could be done similarly for Mac though? and is it worth to do try that for Linux? (I like the app image, but the only installer-like thingy I know is those RPM packages and I don't know if it's FOSS-y enough lol)

In my case I love using the portable, because I hate installing things on windows since well ...it's windows and you get all the bloated goodies for installing on it, most of my FOSS apps are portable unless I have to specifically install it for some sort of obscure reason, but even then all of those packages sure give the option for most users to pick between an installer and a portable version, so I agree it would be a good thing for P2D.

As a related question, would it be possible to make this installer have a VCredist check? that is actually the number 2 problem with anyone installing on Windows, and the solution is far more technical for the average consumer than extracting a compressed folder.

chchwy:
@J5lx good news to have an installer!

@Jose-Moreno Just as @J5lx said, I've been answering the "missing files" questions many times. A bit surprised that some of the people (not many but some) have no idea about zip & unzip. Also, I believe that the installer is able to handle vc redistributable installation, just need to find out how to do it.

But yeah, personally I prefer a portable version if I have choices. We could probably provide both.

Another good idea is giving Windows Store a go.
For example, `Krita` and `Paint.net` are available on Windows Store now
https://blog.getpaint.net/2017/09/29/paint-net-is-now-available-on-the-windows-store/
https://krita.org/en/item/krita-available-from-the-windows-store/

In the blog posts, they both mentioned one-click-installation and auto-update, which is good for those non-technical people.

scribblemaniac:
Very nice! As mentioned, a significant number of people were having difficulties with the zipped executable, so I definitely second the usefulness of this, even if I would opt for the portable version personally. Another step to increasing usability would be static building for Windows, which if I understand what it is correctly, would allow for just the exe file to be distributed. Windows Store is also good.

@Jose-Moreno Mac/Linux should not require a similar thing. Zip files are extracted when opened, rather than "peeked-into" as in Windows. Also for both Mac's .app and Linux's AppImage formats, all of the libraries and resources are self-enclosed, so there's no issues with people not copying some important part of it. RPM packages are fairly FOSS-y imho (although some package managers do one better by compiling from the source).

J5lx:
I don’t know about Mac, but for Linux our current AppImages should suffice, really. Distribution packages are nice, of course, but they are sometimes tricky to get right and are better taken care of by people who actually know their stuff, aka package maintainers of the respective distribution. That’s why mapreri as a Debian person is doing the deb package and I as an Arch user am doing the Arch one.

Regarding static binaries, I can see why they would be nice but I personally don’t really like fat binaries. I’m also not sure what to do with Qt’s plugin architecture (esp. system plugin) and the FFmpeg binary in that case.

As for runtime distribution, it is 100% possible (there are complete solutions available) and it’s totally on my todo list at this point. Getting it into the Microsoft store would probably be a good idea, too, but naturally we’ll have to see how exactly that works and all and whether they take fees or anything. IIRC there was also some talk about getting Pencil into the Mac store some time ago on IRC, we should probably check that out as well.

Also, I actually prefer installation over portable, mainly because it’s so goddamn quick and easy to get just about anything installed on Arch with just the press of one button (kinda), but also because it feels cleaner than having loose app folders floating around.

chchwy:
At the moment, publishing an app on Windows Store costs 19 USD, it's a one-time fee and no renewal charge. But for Mac Store, it costs 99 USD per year, which might be considerable.

Kaikogf:
[The green line and the cross red line over Pencil2D's icon tells a little bit of the story here: it is made by Pencil2D, no mistake! 👍 I like it ]

The Installer program is Much more intuitive. Great

scribblemaniac:
How about these for the install/uninstall icons?
pencil_vector_icon_install
pencil_vector_icon_uninstall

SVG versions [here](https://github.com/pencil2d/pencil/files/1628096/InstallerIcons.zip)

Kaikogf:
The green lines are not intuitive. My thoughts are that you can make an OK mark (V) within the same style as this last red cross uninstaller icon (I like it more than the previous one)

J5lx:
@scribblemaniac I really like those! Personally I find both the check mark ✓ and your arrow-pointing-onto-tray (or whatever one might call it) equally comprehensible, so I’m not sure if the the check mark is better as Kaíko suggested. Also @Kaikogf, I’m afraid I have to admit I used GIMP for the original icons for its guides and scaling features. Shame on me! :)

Kaikogf:
@J5lx 
Now I can see what the green mark meant. So knowing that it started to make sense using it, but I would guess that some people would not grasp it at first sight.

scribblemaniac:
Check
pencil_vector_icon_check

SVG is [here](https://github.com/pencil2d/pencil/files/1628943/pencil_vector_icon_check.svg.zip)

J5lx:
I have now replaced the images with those by @scribblemaniac. Looks good IMO:
inwelcome2
unwelcome2

Kaikogf:
Looks good really!

I have tested the DEMO Installer and it is working just fine! 👍 
I have though two suggestions here: after the succeeded installation of Pencil2D it was not created a shortcut on the desktop (win8.1x64bits). 
I know It is fairly easy to create one, but who knows ... just wanted to say something!

After the uninstallation of Pencil2D there was left its main folder with subfolders.. A very few programs that I know are capable of deleting its own folder after its uninstallation and I found it more clean

;)

best regards

Kaikogf:
I am waiting to see Portuguese (Portugal) to be implemented here... just to check, once again, the translation

J5lx:
That the installer does not create a desktop shortcut is 100% correct intended behaviour. If you ask me, desktop symbols are [an annoying pest](https://imgur.com/LZ2WAmV) and only an excuse for not having an efficient launcher, and they should be purged from this world :) But of course not everyone shares that opinion, so I’ll see what I can do about that. As for the leftover folder, that should not happen if you went through the uninstallation routine properly (i.e. quit Pencil2D before uninstallation etc.). If it does, I’d consider it a bug, but it never happened to myself, so I’ll need some more details from you:

- In which mode did you run the installer (current user / all users)?
- Where did you install the program (C:\…)
- Which directories and files *exactly* are still present after uninstallation? E.g.
  ```
  C:\path\to\instdir\
  C:\path\to\instdir\Pencil2D.exe
  C:\path\to\instdir\audio\
  C:\path\to\instdir\audio\qtaudio_wasapi.dll
  C:\path\to\instdir\audio\qtaudio_windows.dll
  ⋮
  ```

Regarding the missing Portuguese (Portugal), that’s because Pencil2D itself [currently doesn’t include it](https://github.com/pencil2d/pencil/tree/cd399fabf97a69be8e3995816995d0d5472b18f2/translations), but I noticed it’s already fully translated on Transifex, so we can easily add it to the installer as well (just add `!insertmacro MUI_LANGUAGE "Portuguese"` in installer.nsi).

Kaikogf:
"The annoying pest" has no more room for another shortcut jejeje... what a desktop

I run both methods, All Users and Current User
Installed the program to its default path (kept the one of the launcher installation for both methods)

I tried the demo again for checking... this is what was left after uninstalling Pencil2D [**_All Users_**] 
**_Pencil2D path: C:\Programs\Pencil2D_**
nova-imagem

When installing it for **_Current User_**, Pencil2D will be hidden on the AppData folders, so this made me wonder if a beginner (non-technical) tries it he/she will have a hard time getting Pencil2D installation main folder, mainly if he wants to find it for adding a shortcut else where (Windows here) since Pencil2D's launcher doesn't create one
**_Pencil2D path: C:\Users\Carlos\AppData\Local\Pencil2D_**
The left folders are the same as the previous image, nothing inside of them

While installing Pencil2D I did not create a shortcut on the Main Menu (check mark the box)

One thing I would like to point out is that after installing Pencil2D and opening it for the first time, my HUION pen doesn't work, meaning its tip moves all over the UI but nothing reacts to it, no selection no drawing, no moving, etc. After closing and restarting it again it works fine

To be honest I am afraid you would you say all, from top to bottom, is normal

J5lx:
Okay, thanks for the info, I’ll see what I can do about it.

Also, valid point about Pencil not being easily accessible in the AppData folder when no shortcuts were created. I’ll definitely add options for additional shortcuts.

J5lx:
FTR, I have now added pre-scaled versions in 16x16, 32x32 and 48x48 (as per [recommendation from Microsoft](https://msdn.microsoft.com/en-us/library/windows/desktop/dn742485(v=vs.85).aspx)) to the icon file, so the icon now looks much, much better at smaller sizes.

J5lx:
@Kaikogf I finally managed to obtain a legal copy of Win8.1 (which was a little tricky since MS apparently doesn't want people to create new installations of it) to have a look at your issue, but I was not able to reproduce it. I tried installing and uninstalling both per-user and system-wide, both with and without shortcuts, and in all of the cases all files and folders were completely removed without any leftovers. I also looked for hints on the internet but I didn't find any, so I'll have to ask you a few more questions and hope that this gets us any closer to the solution.

1. What locale (i.e. display language, number format etc.) are you using? (Portuguese (Portugal) maybe?) Are there any characters in your file system paths that are not part of the English alphabet, e.g. letters with diacritics or maybe even characters from scripts other than the Latin script?
2. You said that very few programs you know of are capable of cleanly deleting their own folder. Could you list some of the programs that you found to be capable of this? Is it possible that there were no problems originally and that the problems only started occurring at a certain point in time?
3. Do you have any other programs installed that you suspect might interfere with the uninstallation, particularly the kind that integrates deeply with the system (such as anti-virus solutions)?

Kaikogf:
@J5lx 

I will be back to you later with details. Sorry for the delay. Time to rest now
Thank you

One of the programs that I recall right now to be capable of deleting its own directory/folder is _Iobit Unistaller_

CandyFace:
@J5lx I have tested the uninstaller via wine and i'm experiencing the same... there's a bunch of folders left.
screen shot 2018-04-01 at 15 44 40


This is being removed currently:
https://pastebin.com/6WG2bXY2

J5lx:
After some more experimentation, I think I got the issue with the leftover folders on uninstallation fixed... a rather stupid oversight that didn't affect all builds. Could you try installing and uninstalling using the latest installer test build ([32bit][], [64bit][]) and let me know how it goes? Meanwhile, I'll see about the other todo items...

[32bit]: https://drive.google.com/open?id=1uvvw-h9ml-47BTZn2jOgw01SCJQr1Oc2
[64bit]: https://drive.google.com/open?id=1w5DO_3duXoj9zsfLQT-xlIDhMX75ucIt

Kaikogf:
> After some more experimentation, I think I got the issue with the leftover folders on uninstallation fixed... a rather stupid oversight that didn't affect all builds. Could you try installing and uninstalling using the latest installer test build ([32bit](https://drive.google.com/open?id=1uvvw-h9ml-47BTZn2jOgw01SCJQr1Oc2), [64bit](https://drive.google.com/open?id=1w5DO_3duXoj9zsfLQT-xlIDhMX75ucIt)) and let me know how it goes? Meanwhile, I'll see about the other todo items...

Hi @J5lx 
I have tested and have some issues. I have a little free time so I tested and installed it (one user/multiple users, with shortcut/without shortcut):

The shortcut is working fine. I have win8.1 and I know where new items are left so I have to lookup for its shortcut since new items are not set on my simulated desktop environment of win 7

Pencil2D wasn't able to work with my pen, it stopped right after pencil2D opens. Using the mouse worked fine (HUION GT 185)
When uninstalling, it deleted everything from Programs. On the Local Appdata, nothing was deleted and I could still opens pencil from there

Kaikogf:
NOTE:

Using the uninstaller provided on Pencil2D folder, it deleted everything but its main folder (Pencil2D) and a Log folder

J5lx:
Thanks for the quick response! Glad to see that uninstallation from Program Files is working now. As for the leftovers of the AppData uninstallation, it's pretty much expected: The installer doesn't install any log files, so the uninstaller doesn't delete them; instead it assumes that those unknown files were created by the user and leaves them there. That said, this is the first time I'm hearing about Pencil2D creating any logs, and if it interferes with the uninstallation we should figure out where they come from. Could you post the exact name and location of that log folder (relative to the installation directory) as well as its contents, if any?

As for the other stuff, adding extra icons is still on my todo list so right now it's still just the Start Menu entries. Regarding the pen issue, I'm pretty sure it is unrelated to the installer. It might be a good idea to try the "normal" (aka ZIP archive / portable) version of that build ([32bit], [64bit]) and check whether the issue occurs there, too. If that's the case, it's definitely unrelated to the installer and you should consider reporting it as a bug.

[32bit]: https://drive.google.com/open?id=1aXWc_-Q93ABZx-JKMN3lxG7X5FJoapet
[64bit]: https://drive.google.com/open?id=1Y-iChcNF3VDdSNaFBfRF3V_CSRPV8zwW

J5lx:
@Kaikogf Thanks for figuring that pen issue out! With that it's making much more sense to me. I still can't say for sure why it happens (though I've seen Windows device drivers do some interesting things with helper programs and stuff, so I imagine it might be some sort of a permissions problem since we're dealing with elevation and such), but now I should be able to experiment with it a little. Adding this to the todo list! And if it's the worst case and I absolutely can't figure it out, we could still simply consider removing the "Run Pencil2D" checkbox altogether.

@ everyone I'm also pretty much done with the runtime installation part, I think. I'm now getting into those areas where I need additional UI strings that NSIS doesn't provide, but I managed to put those in ini files so once this gets merged I should be able to add them on Transifex just fine. We'll just have to remember to update those translations alongside the others (+ add any new languages to Languages.nsh). But I'll elaborate some more on the details of all this later on.

chchwy:
Good to know that. Well done @J5lx!

Kaikogf:
@J5lx 
Glad to help you all

By the way, is it safe to have the installer only tested with a HUION tablet, mine, in order to be accurate related to this issue of the pen stopping working? 

NOTE:
Today I uninstalled and installed Pencil2D again. After the installation was completed, Pencil2D was launched (checkbox active) and the pen-cursor now was showing a "slightly" different issue (I have experienced this a while back, but it was not consistent), in which the pen-cursor was moving OK all over the canvas, but it was not able to click or select any function/action. Moving the cursor over a tool or function it would be highlighted but after clicking on it nothing would happen.  

Those are the two issues

Thanks

MrStevns:
@J5lx What's the status of this PR?

J5lx:
There’s still some rough edges and some things that are missing entirely (e.g. more shortcuts, as noted by Kaíko). I did some more work locally, now if only I had the time to actually finish it up…

MrStevns:
Excellent, great job finishing the remaining work!

As for future changes and continued contributions to this setup, how does one edit such an installation setup? Is everything written by hand or do you have a tool to generate most of it?

MrStevns (approved):
I've looked through the code and it looks fine to me.
From what I can read though, Wix can be a pretty sizable beast to maintain, and I can imagine that from looking at all the code that you've written already.

As I recall you decided on Wix because it's what you've used before, right?

With all the effort put into this PR, I wouldn't want to see it go to waste but i can't help but ask.

Did you consider an installer that might have required less work to maintain?

Both Nullsoft Scriptable Install System and Inno Setup should be easier to maintain afaik but of course might not be as flexible is writing everything yourself.

Overall good job completing this PR, it's been in a draft state for quite some time and it's a much asked feature, so I think our users will definitely appreciate this!

MrStevns:
> In which way? Code size? ....

Code size is one thing, but that's not my concern. No it's a combination of different things that could increase the friction when having to update this in the future.
- The steps involved to setup the installer
- That almost everything is written by hand, 11 files excluding icons and images that potentially requires changes between releases.
- WiX not being well documented, although it seems to still be somewhat preferred amongst windows developers still, so that might not be a _real_ concern.

Maybe my expectations are just too high as for what an installer framework might do for you. If Windows developers are used to this kind of work, then I'm not one to argue against it, since I'm not one myself 😅 

> Of course, if you don’t see value in that it’s also possible to do a super barebones installer that does the absolute minimum to get the program installed – even in WiX. But personally, I’d rather put a little effort in to make it a decent experience. And even then, what I did here is nothing compared to the complexity of other real world installers such as those of GIMP and Firefox (which use InnoSetup and NSIS, respectively).

It's not so much that I don't see the value in making the installation process decent, because I do (I'm all in for it actually 😀 ), I'm just surprised by the work required to get a decent windows installation experience. When I read about windows installers, it feels like this field hasn't seen much if any innovation in decades.

> Actually, I think this is the most important thing I’d like to say with this comment – at the end of the day you can do simple installers in WiX and very complex ones in InnoSetup and NSIS. It depends more on what you’re trying to do than the tool you use.

Fair point, I have no further comments on the subject, we can merge the PR 🎉

PR #850, opened 2018-01-09
Using miniz as zip library

chchwy:
Hello all,

A heads up. I've changed the Pencil2D's internal zip library for saving/loading pclx from QuaZip+zlib to [miniz](https://github.com/richgel999/miniz), a super lightweight zip library.

With this change, we can get a simpler project structure, and a small performance boost (approx 10% faster zip/unzip). To users, they will not notice any difference because this is an internal code change. All functions are working as before. 

Since miniz has only 1 header and 1 cpp file, it can be easily embedded into the current core_lib project. And all Quazip and zlib source files have been deleted from the repo. #530

The key changes are in filemanager.cpp:
`JlCompress::compressDir()` becomes `MiniZ::compressFolder()`
`JlCompress::extractDir()` becomes `MiniZ::uncompressFolder()`

Another thing worth mentioning is, I made a small change in the `mz_fopen()` function, at miniz.cpp line 2942, to handle Unicode filename on Windows. So it's slightly different from the official version of miniz.

I've tested the new miniz Pencil2D on Windows and Mac, loading/saving is working fine. Haven't tested on Linux yet but I suppose the filesystem is similar to Mac.


J5lx:
Cool! Unicode support sounds like it could be useful to others, have you considered contributing your changes to upstream? I'll also give it a try on Linux asap.

chchwy:
Yeah, I will speak to the miniz guys with the Unicode support, not sure they will accept it because It's a little bit tricky.

To open a file with Unicode filename, it's straightforward on Mac/Linux because filenames are all in UTF-8. But on Windows it requires the filename encoded in **UTF-16**, so I did a UTF-8 to UTF-16 conversion right before opening the file. And it requires the filename are encoded in UTF-8 from the application side (which is not always true on Windows).

Luckily QString can be converted to UTF-8 string just in one function call `QString::toUtf8()` no matter what platform is, so this trick works for us.

Issue #747, opened 2017-08-31
[Enhancement] A way to turn Off/On the sound track


Kaikogf:
### --Issue Summary--
When playing back a sound track, it turns out that the _Hidden Layer Icon_ on the timeline is not used for Off/On the sound track but to hide its layer instead.

### --System Information--
+ Pencil2D Version:
Build 31August2017

+ Operating System:
WIN8.1 X 64bits

+ RAM Size:
32GB

+ Graphics Tablet:
HUION GT185 HD

Kaikogf:
@chchwy 

This comment can be closed since the last build (31Dec2017) brings the on/off sound solution 👍 

By the way, I think, not sure though, that @CandyFace added a special mark for detecting sound tracks on the timeline. I would like to add something about it, but firstly does anyone knows where was it commented? Thanks

CandyFace:
Hmm I don't recall having implemented anything like that, could you give some more details about the "special mark" you recall having seen or heard?

Kaikogf:
My bad. I wrongly thought to have seen marks at the beginning of the sound tracks. 

What I saw was the Scrub of the "timeline" and the "Onion Skin of the Keyframes"

Kaikogf:
Closing this: the sound track has already a feature for ON/OFF the sound.
Thank you

Jose-Moreno:
@Kaikogf Please re-open this. I think this idea is helpful. While the "speaker" turns off all the sounds on the timeline, hiding the layer should, intuitevely only silence the sound(s) on said layer. Think of it as having a BGM and a few sfx, and you just want to focus on a particular sound effect, you hide the other layers and work only on the one that remains visible.

chchwy:
I'm reopening this issue.
Agree with @Jose-Moreno, it's good to have the ability to set every single sound layer on or off.

CandyFace:
This has now been implemented and integrated into the layer visibility system. When you hide a layer, the sound on that said layer will also be muted.

Awaiting confirmation that it works from other contributors, so we can close this.

chchwy:
the feature is available in the 12 Jan nightly build

Kaikogf:
@CandyFace 
This is for you since you was the one to implemented it:
Testing the sound layers on/off implementation.
Two issues:
A)- Testing with 2 sound tracks (on different layers): 
If we want to silence one of them it is not possible doing so while on playback mode. We have to stop playback and them select the specific track.
[...]

PR #853, opened 2018-01-12
enhancement: don't play sound when layer is hidden.

CandyFace:
feature request: #747 

Disabling a layer will now turn off the sound playing on that layer.


chchwy:
OMG you're quick @CandyFace  Thanks.

Issue #855, opened 2018-01-13
[ISSUE] with the automatic resize panels


Kaikogf:
Open Pencil2D with the default UI

When stretching up the _**Tools Panel**_ to its minimum height, it will eventually place all its icons on an horizontal line: two issues here: 
A)- the far right icons will fall out of the main windows boundaries and will be placed over the canvas while all panels below will not be lined up to the last icon. A way to solve this is by selecting a drawing tool here and all panels will be lined-up to the _Tools panel width_. 
B)- After doing the above you will notice that it is not possible to resize (reduce) the width of the main panel. The solution is to resize (grab down) the _**Tools Panel**_ height till it forms two lines of icons. 

Last build 12JAN2018) win8.1x64bits

CandyFace:
Should be fixed in upcoming PR ;)

chchwy:
Please check this in nightly build 15 Jan 2018

Kaikogf:
@chchwy 
Well done here. It is working without any of the issues above. 

But I noticed something new (Issue) related to the same **_Tools Panel_**: it is the only panel that is very hard to attach to other panels/windows. _The more I tried the less responsive it becomes._

While playing around with all other panels I found out that after some attempts with the Tools Panels alone to place it on some other window/panel it will stop attaching for good. While the other panels shows _every-time_ a preview of the new spot with a blue background, that only happens at first attempts with the Tools Panel. After a while it will not show it.

Another subject:
This is a picking thing and has to be with the arrangement of the icons on the Tools Panels. Would it be better to keep the three main drawing tools together? I think so. I mean to put them one next to the other. Even when resizing the tools panel they will not be put apart from the others
Right now they behave like this: 
1 raw of icons - all of them are separated
2 row - **_Pen & Pencil_** are next to each other
3 row - **_Brush & Pen_** together

CandyFace:
@Kaikogf 
> But I noticed something new (Issue) related to the same Tools Panel: it is the only panel that is very hard to attach to other panels/windows. The more I tried the less responsive it becomes.

This is intentional to avoid some situations where the icons doesn't fit the line. So if you for example attach the panel to the top, then you can't just re-attach it to the left or right, because it doesn't fit. You therefore have to detach it and resize it manually so the width is smaller than the height. then you can attach it to the left or right dock. Same rules applies when you've attached the dock to the left or right.

It's not the most ideal solution but I haven't been able to figure a way that would let me correct such error cases without screwing up the layout.

[...]

PR #859, opened 2018-01-14
fixed resize bug when layout was either one row or column

CandyFace:
Check #855 for more details

### DIFF
--- a/app/toolbox.h
+++ b/app/toolbox.h
@@ -35,6 +35,7 @@ class ToolBoxWidget : public BaseDockWidget
 
 public:
     ToolBoxWidget(QWidget* parent);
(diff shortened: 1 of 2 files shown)


app/toolbox.h
```diff
@@ -76,6 +79,14 @@ public slots:
     QToolButton* mEyedropperButton = nullptr;
     QToolButton* mSmudgeButton = nullptr;
     QToolButton* mClearButton = nullptr;
+
+    Qt::DockWidgetArea mAreaLocation;
```
chchwy: Better to have a default value just in case.

chchwy:
Good fixes @CandyFace

PR #866, opened 2018-01-19
Add Code of Conduct

scribblemaniac:
This adds a modified version of the [Contributor Covenant](https://www.contributor-covenant.org) as our Code of Conduct. A lot of it should be common sense, but it's best to have it written out explicitly *before* anything happens. It is important to note that I have widened the scope from that of the initial document, this now applies to all participants on official Pencil2D communication channels, not just those that are representing the project/community. I'm still not particularly impressed with the detailing of this document, but it is a good first step and will suffice for now at least.

@chchwy You will need to modify the email address in the Enforcement section to wherever it is best to have those sent. There should also be some indication of who will read the messages sent to that email address if it is not immediately obvious.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,75 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as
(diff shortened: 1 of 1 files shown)


chchwy:
Thanks @scribblemaniac

PR #871, opened 2018-01-25
Patch desktop entry to be version 1.0 in AppImages

J5lx:
This should allow us to use linuxdeployqt without any modifications. Replaces #823.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,6 @@ addons:
       - doxygen-latex
       - doxygen-gui
       - graphviz
-      - squashfs-tools
       - python3
       - python3-pip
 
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -97,21 +96,11 @@ after_success:
       make INSTALL_ROOT="${PWD}/Pencil2D" install;
       rm -rf Pencil2D/usr/lib;
       echo "Creating AppImage...";
+      sed -i "/^Keywords\(\[[a-zA-Z_.@]\+\]\)\?=/d;/^Version=/cVersion=1.0" Pencil2D/usr/share/applications/pencil2d.desktop;
```
scribblemaniac: Can you add a bit of documentation for what this line does please?

J5lx: It replaces the value of the version key with 1.0 and deletes all lines with a keyword key in any language.

chchwy:
@J5lx Are we able to close #823 after this PR?

PR #872, opened 2018-01-27
[WIP] Migrate UI code to .ui files where possible

J5lx:
As chchwy might remember, this is something that I’ve started working on some time ago. To keep UI and logic separated as much as possible, I’m trying to transition to .ui files wherever possible. Currently a few places are still missing, but what’s there should work. However, given the amount of code involved, there might be some things that I missed, so please let me know if you discover anything that seems odd or broken. As part of this refactoring I also unified some parts of the UI, such as dialog buttons (now always QDialogButtonBox, previously sometimes custom layouts with pushbuttons). I’ll post here again once I’m done with the remaining parts.


app/toolbox.cpp
```diff
@@ -35,130 +36,102 @@ QString GetToolTips(QString strCommandName)
...
-    connect(mSelectButton, &QToolButton::clicked, this, &ToolBoxWidget::selectOn);
-    connect(mMoveButton, &QToolButton::clicked, this, &ToolBoxWidget::moveOn);
-    connect(mPenButton, &QToolButton::clicked, this, &ToolBoxWidget::penOn);
-    connect(mHandButton, &QToolButton::clicked, this, &ToolBoxWidget::handOn);
-    connect(mPolylineButton, &QToolButton::clicked, this, &ToolBoxWidget::polylineOn);
-    connect(mBucketButton, &QToolButton::clicked, this, &ToolBoxWidget::bucketOn);
-    connect(mEyedropperButton, &QToolButton::clicked, this, &ToolBoxWidget::eyedropperOn);
-    connect(mBrushButton, &QToolButton::clicked, this, &ToolBoxWidget::brushOn);
-    connect(mSmudgeButton, &QToolButton::clicked, this, &ToolBoxWidget::smudgeOn);
-    connect(mClearButton, &QToolButton::clicked, this, &ToolBoxWidget::clearButtonClicked);
```
chchwy: I won't recommend doing connections with ui files. UI file is good at positioning GUI things, but it could be annoying that it deals with logic as well. For example, if I want to know the code flow of `pencilOn()` function and search the function name in the entire project, I will finally find it in an XML file like this 
```xml
<connection>
   <sender>pencilButton</sender>
   <signal>clicked()</signal>
   <receiver>ToolBoxWidget</receiver>
   <slot>pencilOn()</slot>
   <hints>
    <hint type="sourcelabel">
     <x>381</x>
     <y>111</y>
    </hint>
    <hint type="destinationlabel">
     <x>199</x>
     <y>99</y>
    </hint>
   </hints>
</connection>
```
The readability is quite bad comparing to a single line connection in a .cpp file
```cpp
connect(mPencilButton, &QToolButton::clicked, this, &ToolBoxWidget::pencilOn);
```

Also I couldn't see any benefits to do connections in ui files.

app/toolbox.cpp
```diff
@@ -35,130 +36,102 @@ QString GetToolTips(QString strCommandName)
...
+    ui->moveButton->setStyleSheet(sStyle);
+    ui->handButton->setStyleSheet(sStyle);
+    ui->penButton->setStyleSheet(sStyle);
+    ui->eraserButton->setStyleSheet(sStyle);
+    ui->polylineButton->setStyleSheet(sStyle);
+    ui->bucketButton->setStyleSheet(sStyle);
+    ui->brushButton->setStyleSheet(sStyle);
+    ui->eyedropperButton->setStyleSheet(sStyle);
+    ui->clearButton->setStyleSheet(sStyle);
+    ui->smudgeButton->setStyleSheet(sStyle);
```
chchwy: Setting the stylesheet in the parent widget will work, seems it doesn't need to repeat so many times.

app/colorbox.cpp
```diff
@@ -14,42 +14,24 @@ GNU General Public License for more details.
...
-#include "colorinspector.h"
 #include "colorbox.h"
+#include "ui_colorbox.h"
 
-ColorBox::ColorBox( QWidget* parent ) : BaseDockWidget( parent )
+ColorBox::ColorBox( QWidget* parent ) :
+    BaseDockWidget( parent ),
+    ui( new Ui::ColorBox )
 {
-	setWindowTitle( tr( "Color Wheel", "Color Wheel's window title" ) );
```
chchwy: Colorbox is a very simple widget, it seems, at least at this point, no benefits to migrating to a ui file since the complicate UIs is done in ColorWheel and ColorInspector.

app/toolbox.cpp
J5lx: Yes, but the designer can’t set it only on OS X I believe.

app/toolbox.cpp
J5lx: Ah yes my reason for this was separation but readability is certainly an important aspect to keep track of. Probably going to revert that then, but I’m going to think about it properly tomorrow when I’m not tired.

chchwy:
Excellent work @J5lx!
I think generally the PR is good except for some concerns in the comments.

This PR will probably not be merged at once because of the considerable amount of code changes, I will cherry-pick the commits into master branch one by one. Hope you understand.

J5lx:
Sure, no problem!

app/colorbox.cpp
chchwy: The reason for separating UI and logic is making code simple, clean and easier to understand. It is a means to an end, not an end in itself.

In this case, ColorBox is a thin layer on top of ColorWheel and ColorInspector that give a simple interface to deal with color sutff.  So introducing `colorbox.ui` doesn't help a lot.

It might take a code reader longer to figure out what happens here, because we got another new file but it doesn't do much. And the task could be done in less 10 lines cpp code originally.

J5lx:
Alright, I think I’m done now except for the points addressed above (which I’ll take care of asap). Technically, TimeLine and TimeControls are still left, but both of those involve QToolBars, and support for those in Qt Designer is half-baked *at best*. Also, I’d like to note that there were a few widgets in TimelinePage that were created but never added to anything (e.g. font size), so I left them out in the UI file.

app/toolbox.cpp
J5lx: Thought about it some more, and it seems that in regard to readability there really is no benefit in keeping the connections in the UI file even know that it’s already done, so I’m going to move the connections back into the code once I find some time for it.

app/colorbox.cpp
J5lx: Thanks for writing this the way you did, it made me rethink the separation of UI and logic, and I believe I have a better grasp now of where it’s worth lies. I took a look again at how some of the other frameworks solve this problem and it occurred to me that it’s not just about the what, but also about the how. Gtk Builder code, [like this one, for instance](https://github.com/GNOME/gnome-chat/blob/a1073ee7e936df04f2bbd4e171809ae25f8a3dca/src/chat-main-window.ui), is rather concise, and even though the example given is also pretty small, I’d still say it is an improvement over hand-written C UI setup code since the hierarchy is clearer (aka a belongs to b and not the other way around) and rather than always thinking about what a particular function/method does, it’s very easy to discern properties from children etc. When I made GTK+ applications back in the day I used to write this code by hand, and *it felt natural to work with*. Qt’s UI code on the other hand is a verbose mess. It feels more like an internal save format of the Designer than a developer tool in itself, and I only ever edit it by hand [when I’m afraid that the Designer would otherwise mess it up more badly than I could](https://github.com/pencil2d/pencil/pull/634#issuecomment-292660879). It doesn’t simplify anything for developers because it isn’t made for developers – if anything, it is made for designers. Just look at those sourceLabel and destinationLabel hints in your comment about the connections, how is that even relevant to anything but the Designer?

tl;dr; I still believe that there can be merit in moving UI setup of this length into dedicated files, but Qt’s solution fails to exploit that potential benefit. As it stands there really is no reason to move ColorBox to a UI file, and while I don’t feel strongly enough about this to completely abandon the work that has already been done, I’m entirely fine with it if you or the others prefer keeping the UI setup in the C++ code. Since you already said you’d like to cherry-pick the commits from this PR individually, feel free to simply skip the one for ColorBox.

chchwy:
Merged. 
Thank you again for this tedious work. @J5lx

J5lx:
Nice! I see you left the connections in the UI files, so should I still migrate them back into the C++ code?

J5lx:
Ah nevermind, just noticed you did that in a separate commit.

PR #875, opened 2018-01-28
Improve movie export dialog

scribblemaniac:
In addition to the main progress bar, a second progress bar and a
label have been added to display subtasks of the exporting. Not only
does this better show progress, but it also makes reports about
freezing/crashing during export more useful (ex. "it says Generating
Palette and then crashes").

The ffmpeg execution code was modified to detect the progress
of the operation. This also fixes a hidden serious bug with movie
exporting: waitForFinished has a timeout of 30 seconds by default,
so any movie that takes longer than that to export would timeout
and fail. This should hopeful fix some of the exporting errors
we've been seeing. waitForReadyRead also has a timeout (default 30s
I think), but this isn't an issue because FFmpeg will write updates
and flush every second or so, returning waitForReadyRead well
before the timeout.

This commit also fixes another issue, where it was impossible to
cancel during the movie export operation because processEvents was
only every being called with QEventLoop::ExcludeUserInputEvents,
which prevented the cancel button event from registering until
after the export is done! The dialog is modal so it still prevents other user input events.

Related: #681

### DIFF
--- a/app/actioncommands.cpp
+++ b/app/actioncommands.cpp
@@ -218,8 +218,8 @@ Status ActionCommands::exportMovie()
 
             QApplication::processEvents();
         },
-        [&progressDlg](const char *s) {
-            progressDlg.setStatus(tr(s));
+        [&progressDlg](QString s) {
+            progressDlg.setStatus(s);
             QApplication::processEvents();
         }
     );
--- a/app/actioncommands.cpp
+++ b/app/actioncommands.cpp
@@ -46,6 +46,7 @@ GNU General Public License for more details.
 #include "exportmoviedialog.h"
 #include "exportimagedialog.h"
 #include "aboutdialog.h"
+#include "doubleprogressdialog.h"
 
 
 ActionCommands::ActionCommands(QWidget* parent) : QObject(parent)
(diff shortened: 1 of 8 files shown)


app/doubleprogressdialog.h
```diff
@@ -0,0 +1,56 @@
...
+    Q_OBJECT
+
+public:
+    explicit DoubleProgressDialog(QWidget *parent = 0);
+    ~DoubleProgressDialog();
+
+    QString getStatus();
+    void setStatus(QString msg);
+
+    class ProgressBarControl {
```
chchwy: Any reason to make this inner-class? 

Most of the functions of `ProgressBarControl` can be done with a plain `QProgressBar`.
setMin() => QProgressBar::setMinimum()
setMax() => QProgressBar::setMaximum() 
setRange() => QProgressBar::setRange() 
setPresision() => why need this if there is already a range?

update progress: `float_progress_value * (max - min)`

core_lib/movieexporter.cpp
```diff
@@ -161,27 +165,47 @@ Status MovieExporter::run(const Object* obj,
...
-        STATUS_CHECK(assembleAudio(obj, ffmpegPath, progress));
+        progressMessage("Assembling audio...");
+        minorProgress(0.f);
+        STATUS_CHECK(assembleAudio(obj, ffmpegPath, minorProgress));
+        minorProgress(1.f);
     }
-    progress(0.10f);
+    majorProgress(0.1f, 0.5f);
+    qDebug() << "Time:" << clock();
+    progressMessage("Generating frames...");
```
chchwy: tr() this

core_lib/movieexporter.cpp
```diff
@@ -161,27 +165,47 @@ Status MovieExporter::run(const Object* obj,
...
-    progress(0.9f);
+    STATUS_CHECK(generateImageSequence(obj, minorProgress));
+    qDebug() << "Time:" << clock();
+    minorProgress(1.f);
 
+    minorProgress(0.f);
     if (desc.strFileName.endsWith("gif", Qt::CaseInsensitive))
     {
-        STATUS_CHECK(convertToGif(ffmpegPath, desc.strFileName));
+        progressMessage("Generating palette...");
```
chchwy: tr() this and so on.

app/doubleprogressdialog.h
scribblemaniac: This was to enable float values while obscuring the implementation details. I also wanted the `progressDialog.major` and `progressDialog.minor` format without modifying the UI directly from other classes (imo, the UI implementation should be separate from the widget's api, if you decide to change how the user interface is displayed, you should not have to change code outside of widget's class). Putting stuff like `float_progress_value * (max - min)` makes it less readable in my opinion. The way I have it, you just give it the percentage complete by default. `setPrecision` sets the number of decimal places that the progress bar shows. If the range is 0-1 and the precision is 2, then there are 100 steps. All of the corresponding QProgressBar functions that you pointed out are int only, which is not practical when you are thinking in terms of percentages.

app/actioncommands.cpp
```diff
@@ -183,25 +184,45 @@ Status ActionCommands::exportMovie()
...
+        },
+        [&progressDlg, &minorStart, &minorLength](float f) {
+            progressDlg.minor->setValue(f);
+
+            progressDlg.major->setValue(minorStart + f * minorLength);
+
+            QApplication::processEvents();
+        },
+        [&progressDlg](const char *s) {
+            progressDlg.setStatus(tr(s));
```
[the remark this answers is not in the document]
scribblemaniac: Right, forgot about that. I will fix this right away.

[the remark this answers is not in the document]
scribblemaniac: Done. Had to use `QApplication::tr` because `MovieExporter` is not a `Q_OBJECT`

app/doubleprogressdialog.h
chchwy: If I set range(1, 2) and setValue(1.5), things screw up. 

Either use max-min or use precision with a uniform 0~1 progress value. It doesn't make sense to have both at the same time. And I thought `ratio * (end - begin)` is a very common way to do distance calculation?

app/actioncommands.cpp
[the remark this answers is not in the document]
chchwy: QObject::tr() is good as well.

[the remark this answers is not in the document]
scribblemaniac: Good to know.

app/doubleprogressdialog.h
scribblemaniac: > If I set range(1, 2) and setValue(1.5), things screw up.

What exactly screws up? That should still work. Now that I think about it, there's an issue when you change the precision. Regardless of that, doing what you describe should put the progress bar exactly half way as one would expect.

I'll think about this a bit more and continue the discussion tomorrow.

chchwy: Ah, my bad, I didn't notice that you do `convertUnits()` in setMin()/setMax() as well. Apologise.

chchwy: Finally understood what you did. It's mapping an arbitrary range of floating numbers to a range of integers for the progress bar. It's more delicate than I thought.

scribblemaniac: That may be the first time my code has been described as "delicate". Well thank you, glad you like it 😁

PR #876, opened 2018-01-29
Speed up movie export

scribblemaniac:
There have been so many good PRs lately, I've got to step up my game! 💪 

While making #875, I noticed some significant bottlenecks with the current movie exporter that could easily be addressed. First off is the `QImage.save`, which by default renders a PNG with some compression, which is normally what you want, such as for image sequence export, but when they are just temporary images, uncompressed images are going to be faster to make and read. After some testing, my suspicions were confirmed, uncompressed PNGs were both written faster and processed faster by FFmpeg than compressed ones. Turns out that BMPs performed even better than uncompressed PNGs, and oddly enough used less space, so that's what I went with.

The second bottleneck was that the temporary images were taking up a significant amount of space on the hard drive (even more so now that they are uncompressed) and they obviously waste a lot of time writing to the hard drive just to load it back into memory in FFmpeg. #681 will no doubt fix this issue, but until then we can pipe the image data into FFmpeg through stdin! The frames are processed as soon as they are generated, resulting in no storage space requirements, minimal memory usage, and faster exporting.

Unfortunately I can't think of a way to implement the pipes solution with the two-step GIF export, so that will be a bit slower (the uncompressed intermediary images will still help considerably). Also another side effect of these changes is that the combineVideoAndAudio function is a complete mess now. Considering that the image sequence data has to be passed to ffmpeg, the ffmpeg results determine when to render the next frame, and everything's got to be synchronous, there is no easy way to keep the functions neatly separated. The only good way I can think of to do this is to rewrite the whole thing to be asychronous using signals for generating frames, announcing ffmpeg updates, and so on. Given that this exporter isn't even here to stay, I don't think refactoring the entire class is worth it.

**TLDR;** Here are a few changes to make movie exporting faster (a tenfold time decrease in my tests), mainly just while we're waiting for #681.


scribblemaniac:
And here's the benchmark data I obtained from single trials with a very large project (first 2 minutes of Sintel. Stats: 3001 frames, 1920x818 export resolution, 17 GB of memory 😮, infinite saving time ).

#### PNG with Default Compression
Image generation duration: 1365.33s
FFmpeg duration: 1.02s
Total duration: 1366.35s

Secondary storage: 3.2GB

#### PNG with Minimum Compression
Image generation duration: 329.81s
FFmpeg duration: 0.12s
Total duration: 329.93s

Secondary storage: 18GB

#### BMP with Minimum Compression
Image generation duration: 88.72s
FFmpeg duration: 0.48s
Total duration: 89.20s

Secondary storage: 13GB

#### Code from this PR
Total duration: 115.75s
Note: This is a bit slower that the BMP method because there is some CPU time wasted when FFmpeg is waiting for frames to be generated, but the memory advantages greatly outweigh this slight slowdown.

Secondary storage: 0GB

Timing was done with clock(), which isn't super high-res I know, but more than good enough when we are talking about a 20 minute difference!

chchwy:
Very exciting achievement @scribblemaniac !

The only thing I concern is running out of disk space when exporting gif from a bunch of bmp files.
Probably have to check the remaining free disk space at the beginning (can use `QStorageInfo`), since it is very likely user don't have 10G free spaces.

scribblemaniac:
@chchwy First off, my example was very extreme, other aspects of the program prevent a 3000 HD frame project from being even feasible to create (for example, I can't seem to even save the project, so I have to reimport the frames every time I want to test it 😖 ). A 10GB intermediate export would be very unusual.

Even more important to consider is that all of the frames are stored uncompressed in virtual memory. In other words, if the export is 10GB, Pencil2D is taking up more than 10GB (in the case of my test example, it was taking up 17GB). I feel it's a fairly safe assumption that disk space > RAM size for any computer, high or low end. If they don't have enough room to export, they won't have enough room to load the animation in the first place.

chchwy:
As far as I know, a user on our forum made a 7000 frames animation, so yes, it's extreme, but someone was doing it. And a 7000-frame animation doesn't mean there are 7000 keyframes, maybe 500 keyframes loosely spread through the whole timeline.

Also, there are a lot of laptops equip 120G SSD today, so it could be very common that the free space is less than 5GB.

chchwy:
Yeah, but maybe you're right. Exporting a 3000-frame gif is crazy.

scribblemaniac:
It may be possible to use `QImage.colorTable` to generate the palette beforehand, allowing us to use the FFmpeg pipe trick with GIFs. This is not trivial however as you would have to add code for color quantization (which probably won't be as good as whatever ffmpeg does) and export it into whatever format ffmpeg expects. And this could slow things down quite a bit because you have to go through all the layers on all the frames and at the very least render the vector layers (which would in total be done twice instead of once). Or you could just generate all the frames twice, once for the palette and once for the movie, it would eliminate the disk space requirement, but double the time obviously.

A 7000 frame project at the default resolution is 600*800*7000 = ~1 billion pixels. A 3000 frame project at the resolutions I had is 1920*818*3000 = ~14 billion pixels. So that 7000 frame project would assumably be under 1GB when using BMPs.

chchwy:
The app froze when I tried to export a video.

After some debugging, the program got stuck in the movieexporter.cpp line 460
```
ffmpeg.waitForReadyRead();
```

The steps before the app frozen:

1. Go Into `MovieExporter::combineVideoAndAudio()`
1. First time run into the outer while loop `while(ffmpeg.state() == QProcess::Running)`
2. run into the inner while loop `while(currentFrame - frameStart <= framesProcessed + 10)`
3. Everything ok. Left the inner while loop
4. Second time run into the outer while loop.
5. Get stuck at line 460 `ffmpeg.waitForReadyRead();` 

Wait for more than 1 minute but the app was still not able to respond.

It's a 8-frame project and ffmpeg version is 3.4.1.

scribblemaniac:
Very strange, I can't replicate this even with very small projects. I would expect that the worst case scenario when something goes wrong is that waitForReadyRead times out after 30 seconds. Could you share the last few lines of your application output before it freezes? I'm particularly looking for lines from FFmpeg like this:
```
"frame=    5 fps=0.0 q=-0.0 Lsize=     102kB time=00:00:00.34 bitrate=2463.1kbits/s speed=3.74x    \n"
```
Or empty lines which suggests that no output was available. Also what format are you exporting to?

Thanks for the debugging steps, as a developer I'm sure you understand how much I appreciate that 😉. Unfortunately I'm still not sure what's causing the issue.

chchwy:
Seems the project length doesn't matter. it gets stuck at the same location when I exported a 1300 frame project. And it happens on both my laptop and PC.

Here is the log 
```
"D:/Github/pencil/app/../bin/plugins/ffmpeg.exe"
OutFile:  "C:/Users/Matt/Desktop/hey.mp4"
Audio Length =  110.083  seconds
TempAudio= "C:/Users/Matt/AppData/Local/Temp/Pencil2D-BAIX6a/tmpaudio0.wav"
"\"D:/Github/pencil/app/../bin/plugins/ffmpeg.exe\" -f image2pipe -framerate 12 -start_number 1 -i - -threads 16 -s 640x480 -pix_fmt yuv420p -y \"C:/Users/Matt/Desktop/hey.mp4\""
[stdout] "ffmpeg version 3.4.1 Copyright (c) 2000-2017 the FFmpeg developers"
[stdout] "  built with gcc 7.2.0 (GCC)"
[stdout] "  configuration: --enable-gpl --enable-version3 --enable-sdl2 --enable-bzlib --enable-fontconfig --enable-gnutls --enable-iconv --enable-libass --enable-libbluray --enable-libfreetype --enable-libmp3lame --enable-libopenjpeg --enable-libopus --enable-libshine --enable-libsnappy --enable-libsoxr --enable-libtheora --enable-libtwolame --enable-libvpx --enable-libwavpack --enable-libwebp --enable-libx264 --enable-libx265 --enable-libxml2 --enable-libzimg --enable-lzma --enable-zlib --enable-gmp --enable-libvidstab --enable-libvorbis --enable-cuda --enable-cuvid --enable-d3d11va --enable-nvenc --enable-dxva2 --enable-avisynth --enable-libmfx"
[stdout] "  libavutil      55. 78.100 / 55. 78.100"
[stdout] "  libavcodec     57.107.100 / 57.107.100"
[stdout] "  libavformat    57. 83.100 / 57. 83.100"
[stdout] "  libavdevice    57. 10.100 / 57. 10.100"
[stdout] "  libavfilter     6.107.100 /  6.107.100"
[stdout] "  libswscale      4.  8.100 /  4.  8.100"
[stdout] "  libswresample   2.  9.100 /  2.  9.100"
[stdout] "  libpostproc    54.  7.100 / 54.  7.100"
[stdout] "Input #0, image2pipe, from 'pipe:':"
[stdout] "  Duration: N/A, bitrate: N/A"
[stdout] "    Stream #0:0: Video: bmp, bgr24, 640x480, 12 fps, 12 tbr, 12 tbn, 12 tbc"
[stdout] "Stream mapping:"
[stdout] "  Stream #0:0 -> #0:0 (bmp (native) -> h264 (libx264))"
[stdout] "[libx264 @ 00000268cc8328c0] using cpu capabilities: MMX2 SSE2Fast SSSE3 SSE4.2 AVX FMA3 BMI2 AVX2"
[stdout] "[libx264 @ 00000268cc8328c0] profile High, level 2.2"
[stdout] "[libx264 @ 00000268cc8328c0] 264 - core 152 r2851 ba24899 - H.264/MPEG-4 AVC codec - Copyleft 2003-2017 - http://www.videolan.org/x264.html - options: cabac=1 ref=3 deblock=1:0:0 analyse=0x3:0x113 me=hex subme=7 psy=1 psy_rd=1.00:0.00 mixed_ref=1 me_range=16 chroma_me=1 trellis=1 8x8dct=1 cqm=0 deadzone=21,11 fast_pskip=1 chroma_qp_offset=-2 threads=16 lookahead_threads=2 sliced_threads=0 nr=0 decimate=1 interlaced=0 bluray_compat=0 constrained_intra=0 bframes=3 b_pyramid=2 b_adapt=1 b_bias=0 direct=1 weightb=1 open_gop=0 weightp=2 keyint=250 keyint_min=12 scenecut=40 intra_refresh=0 rc_lookahead=40 rc=crf mbtree=1 crf=23.0 qcomp=0.60 qpmin=0 qpmax=69 qpstep=4 ip_ratio=1.40 aq=1:1.00"
[stdout] "Output #0, mp4, to 'C:/Users/Matt/Desktop/hey.mp4':"
[stdout] "  Metadata:"
[stdout] "    encoder         : Lavf57.83.100"
[stdout] "    Stream #0:0: Video: h264 (libx264) (avc1 / 0x31637661), yuv420p, 640x480, q=-1--1, 12 fps, 12288 tbn, 12 tbc"
[stdout] "    Metadata:"
[stdout] "      encoder         : Lavc57.107.100 libx264"
[stdout] "    Side data:"
[stdout] "      cpb: bitrate max/min/avg: 0/0/0 buffer size: 0 vbv_delay: -1"
```

scribblemaniac:
The issue was that 10 frames were sent and processed by FFmpeg, but it did not output a progress update right away as I assumed it would. Instead it was waiting for more frames, and Pencil2D was waiting for an update before sending more frames, resulting in a deadlock. I have fixed this by sending a new frame after a second with no output from FFmpeg and no new frames.

scribblemaniac:
Actually it may be possible to combine the palette and export into one command using the filter_complex argument. I don't use that very often, but I will look into it to see if I can make it work.

scribblemaniac:
Okay, I've found a way to combine the GIF generation into a single FFmpeg command, allowing me to use the same pipe technique to eliminate writing frames to the disk. I have also found a way to refactor it to make it reasonably tidy. There is a drawback with this approach which is that the palettegen filter requires all of the frames to be written to the pipe before the encoding can begin. This means that all the uncompressed frames have to be held in virtual memory by FFmpeg. In the currently expected use cases for Pencil2D, this will not be much of an issue or even noticable, but for animations larger than the user's free memory, then it will be about as slow as writing to the disk, and probably more disruptive to the computer if the users are trying to use other applications while the animation is exporting.

Here are updated benchmarks for the 842cc98 commit:
MP4 export duration: 95.20s (yes I slipped a few more optimizations into this 😉 )
GIF export duration: 95.99s. I actually only exported 1/3 of the test animation and then multiplied the duration by 3 due to memory constraints. You can expect that if the animation is large enough to run out of primary memory, it will be much slower.

Jose-Moreno:
@scribblemaniac Some pro animators from waht is called "the web gen" have become known for and fancy creating longer GIF animations (1500+) than what is normal. Considering this case would there be an issue with your proposed implementation for such cases? Could there be an optional performance vs stability toggle?

scribblemaniac:
@Jose-Moreno Long GIF animations are still possible and will in fact be more likely to work if limited disk space, not limited memory, is actually the issue as @chchwy was previously suggesting. Lower resolutions (ex. 480p) will be much easier to export long GIFs than high res ones obviously.

There could be some toggle, but I would worry about clutter and user-friendliness. Maybe some automatic detection and switch based on computer specs and the estimated animation size could be done instead. Going even more extreme, instead of just writing to the disk, we could fallback to the poor quality export we have in v0.6.0 which doesn't have a palette and therefore does not need to write the entire animation at one time.

However I'm still not convinced any of this is needed. The fact of the matter is that long animations are impossible to make in Pencil2D right now for other reasons: Pencil2D is horrible with memory management and it takes a *very* long time to save animations large enough to cause issues with GIF export (in fact I'm not even sure the save works at such sizes because we don't have a progress bar and I don't have the patience to wait 1+ hours (this is not an exaggeration)). When we fix these issues, then we can revisit such currently theoretical issues.

chchwy:
Did a quick test this morning. The results of exporting time are not ideal.

Gif exporting is a lot better.

| Gif export master branch | GIF export with this PR |
| -----| ----|
| 53.8 secs| 17.6 secs|

Mp4 exporting is 10 times slower.

| Mp4 export master branch | Mp4 export with this PR |
| -----| ----|
| 42.7 secs| 463.5 secs |

Here is the log output:
https://gist.github.com/chchwy/05aea23838b8b18f27f5ab87488718cb
https://gist.github.com/chchwy/a04a0b8cc2605d3cd1cc7b53760b2b58

This project has 1 bitmap layer, 1 camera layer, 640x480, 1321 frames.
ffmpge 3.4.1 on windows 10 64bit

scribblemaniac:
@chchwy Sorry for the delay, I've been a bit too busy to look into this recently. I've made some changes that will either fix the slow MP4 export, or give me a better idea of why it is happening. Please pull the latest update and test when you get a chance.

chchwy:
@scribblemaniac 
the same project, new export time:
GIF Export time = 17.141 sec
MP4 Export time = 8.540 sec
Output log: https://gist.github.com/chchwy/ba61bf819333e07ab06bd9262926e5a0

It's pretty amazing.

scribblemaniac:
Okay. I gave up on the idea of making FFmpeg work the way I want it to, it seems to have a (stubborn) mind of its own. I've cleanup the debug messages, updated it with the changes from master, and added some documentation. It is ready to go now!

J5lx:
@scribblemaniac The reason why the QImage data is BGRA rather than ARGB is because it depends on endianness, while FFmpeg’s pixel formats do not. So to pick the right format for FFmpeg you could do something like this (taken from my exporter rewrite):

```cpp
#if Q_BYTE_ORDER == Q_LITTLE_ENDIAN
#define PIX_FMT_QIMAGE_ARGB32 AV_PIX_FMT_BGRA
#else
#define PIX_FMT_QIMAGE_ARGB32 AV_PIX_FMT_ARGB
#endif // Q_BYTE_ORDER == Q_LITTLE_ENDIAN
```

PR #886, opened 2018-02-06
Remove hardcoded fonts

### LINKED ISSUE #874: Inconsistent use of Helvetica
While working on #872 I noticed that in some places the font is explicitly set to Helvetica (most notably in various timeline-related widgets: timeline, timecontrols, etc.), while most other code simply uses the default font. It might be a good idea to decide on one solution here: Use Helvetica everywhere, or nowhere. If we use it everywhere, though, we should probably at least think about the cases where it isn’t available. My system, for instance, does not have it and uses DejaVu Sans instead, which is (indirectly) based on Verdana rather than Helvetica, so it kinda defeats the purpose. Not like it’s super important (in the end a font is a font), but I just wanted to drop it here for thought.

J5lx:
The html property in errordialog.ui was overridden from the code anyway, so I removed it completely. The other changes should be quite self-explanatory. Resolves #874.

### DIFF
--- a/app/ui/errordialog.ui
+++ b/app/ui/errordialog.ui
@@ -81,20 +81,8 @@
    </item>
    <item>
     <widget class="QTextEdit" name="details">
-     <property name="font">
-      <font>
-       <family>.SF NS Text</family>
-      </font>
-     </property>
      <property name="readOnly">
       <bool>true</bool>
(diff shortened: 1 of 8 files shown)


chchwy:
Lovely @J5lx

Issue #890, opened 2018-02-13
Git master won't compile on debian stretch.


romeojulietthotel:
### --Issue Summary--

Git master won't compile on debian stretch.

### --Actual Results--

Compile fails.

### --Expected Results--

Compile  successfully succeeds.

### --Video or Image Reference--

No video available.

### --Steps to reproduce--

1. install debian stretch
2. git clone https://github.com/pencil2d/pencil
3. cd pencil
4. qmake -r pencil.pro CONFIG+=debug
5. make

### --System Information--
+ Pencil2D Version:

git master

+ Operating System:

debian stretch

+ RAM Size:

+ Graphics Tablet:

---

**Here are the first two errors emitted:**

```
In file included from interface/editor.cpp:18:0:
interface/editor.h:48:1: error: expected class-name before { token
 {
 ^
interface/editor.h:49:5: error: Q_OBJECT does not name a type
     Q_OBJECT
     ^~~~~~~~
```

Jose-Moreno:
@romeojulietthotel This is rather weird.
If you want hit our devs on IRC: http://webchat.freenode.net/?channels=#pencil2d
There should be someone online at the moment.

scribblemaniac:
This is an issue which we have also noticed. @J5lx said he would make a PR to fix this soon.

chchwy:
Please check the latest master branch, should be fixed.

J5lx:
@chchwy @scribblemaniac Yeah I'm working on it but I didn't have time until now. Actually editor.h isn't the only file where headers are missing, so a PR is still coming.

romeojulietthotel:
Yeah there are a bunch of missing includes.  QObject, QMouseEvent, QCloseEvent and I think that's it.

J5lx:
PR is up #891.

romeojulietthotel:
Interesting, I added `#include <QCloseEvent>` to mainwindow2.cpp and I also added `#include <QObject>` to editor.h.  I'm not saying yours are wrong. But I was able to build and pencil is working.

J5lx:
QObject was already added to editor.h by @chchwy. Strange about QCloseEvent though… Maybe QTabletEvent includes it? Though I’m not sure whether I saw anything about QCloseEvent at all… Maybe I’ll investigate some more tomorrow.

romeojulietthotel:
I saw a specific error about QCloseEvent. I looked at the Qt docs and maybe the best include is `#include <QInputEvent>`.

PR #891, opened 2018-02-14
Various fixes

J5lx:
See #890, also compiler warnings. First commit was tested on Ubuntu Xenial only, but since the symptoms were the same I suppose the fix should also work on Debian Stretch.


chchwy:
Thanks @J5lx

PR #892, opened 2018-02-14
Revise Linux compilation guide

J5lx:
Mainly added instructions for Arch Linux and changed the instructions to use out-of-source builds since those are generally recommended these days (for good reason). Also some minor fixes. I removed some of the qt packages from the apt install line for Ubuntu since I believe they are unnecessary (and I was able to compile the program without them) but if there’s indeed a reason for installing those as well that I’m not aware of please let me know.

### DIFF
--- a/docs/build_linux.md
+++ b/docs/build_linux.md
@@ -2,46 +2,74 @@
 
 These are instructions for building Pencil2D on Linux. If you are using Windows go [here](build_win.md), and if you are using macOS go [here](build_mac.md). This guide is primarily targeted towards developers. If you just want to use the latest version you can just download one of our [nightly builds](https://drive.google.com/drive/folders/0BxdcdOiOmg-CcWhLazdKR1oydHM).
 
-This tutorial was made with Ubuntu Xenial Xerus (16.04) in mind, however you should be able to adapt this guide to other versions or distributions if necessary.
+This tutorial was made with Ubuntu Xenial Xerus (16.04) and Arch Linux in mind, however you should be able to adapt this guide to other versions or distributions if necessary.
 
 ## Installing Dependencies
 
-There are a few things that must be installed in order to build Pencil2D. In this section we will go over the installation of each of these components in detail. We describe both graphical and command-line methods of installation, choose whichever one you feel more comfortable with (and if you don't know the difference, choose graphical).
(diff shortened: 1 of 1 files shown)


docs/build_linux.md
```diff
@@ -2,46 +2,74 @@
...
-- Next you can select the components you wish to install. At the very least you should have Desktop gcc selected under the latest Qt version. Also make sure Qt Creator under the Tools section is being installed (at the time of writing there is no option to uncheck this, but it's worth double checking!)
+- Next you can select the components you wish to install. At the very least you should have Desktop GCC selected under the latest Qt version. Also make sure Qt Creator under the Tools section is being installed (at the time of writing there is no option to uncheck this, but it's worth double checking!)
 - Agree to the license and begin the installation. It will take a long time to download all of the files, so be patient. When the installation is complete, press Done and it will launch Qt Creator for you.
 
-#### Command-line method
+##### Command-line method
 
 Pencil2D must be built with the Qt 5 framework. To install Qt 5, run this command:
 
-    sudo apt install qt5-default qt5-qmake qtdeclarative5-dev qttools5-dev libqt5xmlpatterns5-dev libqt5svg5-dev qtmultimedia5-dev
```
scribblemaniac: I think qttools5-dev is needed for handling the translation files. qtdeclarative5-dev is/was probably needed for something too, but I don't remember what.

J5lx: Hmm are you sure? qttools5-dev only contains some includes and libraries. Maybe qttools5-dev-tools instead, which contains linguist and such? Also qtdeclarative5-dev only contains stuff for Qml and Qt Quick which I’m pretty sure we don’t use. See https://packages.ubuntu.com/xenial/amd64/qtdeclarative5-dev/filelist, https://packages.ubuntu.com/xenial/amd64/qttools5-dev/filelist, https://packages.ubuntu.com/xenial/amd64/qttools5-dev-tools/filelist

scribblemaniac: Oh sorry, I was looking at the docs for qttools5-dev-tools. You're right, I'll double check, but neither of these look necessary.

scribblemaniac: Just tested this. I can confirm that the build worked fine without either of these. 👍

PR #893, opened 2018-02-14
Link API documentation to Qt documentation

J5lx:
As the title says. For the tag file location I used util/ and the system documentation locations for Ubuntu and Arch, Doxygen will skip the ones that do not exist and the first one found will take precedence. Tag files in util/ are assumed to belong to Qt’s online documentation (downloaded when documentation is deployed, for instance), the other ones will be linked to local documentation. If Qt also has a standard documentation path that can be relied on on Windows and/or OS X, adding those would probably be a good idea, too.


J5lx:
Gah, just noticed Travis has it’s own Doxyfile. I’ll make a followup for that.

PR #896, opened 2018-02-17
Rework the project structure to be more concise

J5lx:
As discussed in chat. Mainly I rearranged the directories but there are also some smaller changes.

- Directories: The split nature of the project with its three sub-projects app, core_lib and tests is now reflected in the directory structure. Each subproject comes with its own data, sources and UI, however I chose to leave the translations at the top-level directory at least until the split between app and core_lib is consolidated, since moving translations around seems to be rather tricky. The documentation and util directories also continue to live at the top level.
- app and core_lib now come with their own resources. Since the separation of the two projects is still not perfect, I chose to move all resources that are only used in core_lib to core_lib, while resources that are also or only used in app live in app. In the process I also moved the icons that have fallen out of use into app/data/icons/unused, which in the future will help us do a little housekeeping and determine which icons we might want to replace (related: #550, #603). Furthermore I removed some of the icon files since they were duplicated.
- I also removed resources/about.html which is no longer used.
- In addition, I removed all entries from the qrc files that weren’t used in the program, which will reduce the size of the program ever so slightly.
- In file names, the project is now consistently named pencil2d (rather than Pencil2D or just pencil). Most notably this affects the main project file (pencil.pro, now pencil2d.pro).
- DEPENDPATH has been removed from the QMake files since according to [this](https://stackoverflow.com/a/16769542/3162530) it is no longer relevant or necessary in Qt 5.

The best way to get a feel for the new structure is probably by [browsing the tree](https://github.com/J5lx/pencil/tree/rearrange-everything).


J5lx:
If you feel that something should be done differently, feel free to let me know! After all, this is quite the major change.

J5lx:
Please don’t merge just yet, I just noticed the locations of the qrc files need to be adjusted in the ui files as well.

J5lx:
All fixed, can be merged now.

PR #898, opened 2018-02-22
Fix #851. Make sure the FPS spin-box works on Linux.

### LINKED ISSUE #851: Animation wont play on more than 12 FPS
okay so it wont play on 24 speed

even if i set it to 24, it still plays on 12
please fix  this issue

martinvanzijl:
Fixes #851. This works for me on Ubuntu 14.04.

## Technical Details ##
The connection from the FPS spin-box is set up here:

`connect(mFpsBox, spinBoxValueChanged, this, &TimeControls::fpsClick);`

The Application Output tab showed a warning that the connection was not valid. I added a breakpoint to "fpsClick" and changed the FPS value, but "fpsClick" was never called. 

I tried a couple of other things (from https://stackoverflow.com/questions/16794695/connecting-overloaded-signals-and-slots-in-qt-5/16795664) but couldn't get the valueChanged signal to work. 

So finally I changed it so that if the "spinBoxValueChanged" connection is invalid, we use the "editingFinished" signal instead.

### DIFF
--- a/core_lib/src/interface/timecontrols.cpp
+++ b/core_lib/src/interface/timecontrols.cpp
@@ -166,9 +166,12 @@ void TimeControls::makeConnections()
 
     connect(mSoundButton, &QPushButton::clicked, this, &TimeControls::soundClick);
     connect(mSoundButton, &QPushButton::clicked, this, &TimeControls::updateSoundIcon);
-    connect(mFpsBox, spinBoxValueChanged, this, &TimeControls::fpsClick);
-
-
+    auto connection = connect(mFpsBox, spinBoxValueChanged, this, &TimeControls::fpsClick);
+    if(!connection)
+    {
+        // Use "editingFinished" if the "spinBoxValueChanged" signal doesn't work...
+        connect(mFpsBox, &QSpinBox::editingFinished, this, &TimeControls::onFpsEditingFinished);
+    }
 }
 
 void TimeControls::playButtonClicked()
@@ -263,6 +266,11 @@ void TimeControls::updateSoundIcon(bool soundEnabled)
     {
         mSoundButton->setIcon(QIcon(":icons/controls/sound-disabled.png"));
     }
+}
+
+void TimeControls::onFpsEditingFinished()
+{
+    emit fpsClick(mFpsBox->value());
 }
 
 void TimeControls::updateLength(int frameLength)
--- a/core_lib/src/interface/timecontrols.h
+++ b/core_lib/src/interface/timecontrols.h
@@ -52,6 +52,9 @@ public slots:
     void toggleLoop(bool);
     void toggleLoopControl(bool);
 
+    /// Work-around in case the FPS spin-box "valueChanged" signal doesn't work.
+    void onFpsEditingFinished();
+
 private:
     void makeConnections();
 


chchwy:
Thank you @martinvanzijl ! 

Have you tried the old style connection like this:
`connect(mFpsBox, SIGNAL(valueChanged(int)), this, SLOT(fpsClick(int));`

Does it work?

martinvanzijl:
Hi @chchwy , I just tried the old style connection you suggested, but it doesn't work either. It gives the following error in the Application Output:

`QObject::connect: No such slot TimeControls::fpsClick(int) in ../../pencil/core_lib/src/interface/timecontrols.cpp:172`

Also, I noticed the two "range" spin-boxes also have the same problem. How about I raise a seperate PR to fix those too?

J5lx:
@martinvanzijl That error probably occurs because fpsClick(int) is a signal rather than a slot. Try `connect(mFpsBox, SIGNAL(valueChanged(int)), this, SIGNAL(fpsClick(int));` instead.

martinvanzijl:
@J5lx Thanks, that works!

Maybe we should use that instead of falling back on "fpsEdit". Is there any downside to using this 'old' style of connection?

scribblemaniac:
I think we should fall back to the old style if that works. You can see the advantages and disadvantages of the new style [here](https://wiki.qt.io/New_Signal_Slot_Syntax#Pros). Using two different styles is also a disadvantage specific to this project, but that's better than it not working at all for some people. valueChanged is better than editingFinished because a "live preview" can be shown rather than it only updating when it looses focus. Could you also try this: `connect(mFpsBox, spinBoxValueChanged, [this] (int i) { emit fpsClick(i); });`

CandyFace:
### Pros
+ Compile time check of the existence of the signals and slot, of the types, or if the Q_OBJECT is missing.
+ Argument can be by typedefs or with different namespace specifier, and it works.
+ Possibility to automatically cast the types if there is implicit conversion (e.g. from QString to QVariant)
+ It is possible to connect to any member function of QObject, not only slots.
### Cons
+ More complicated syntax? (you need to specify the type of your object)
+ Very complicated syntax in cases of overloads? (see below)
+ Default arguments in slot is not supported anymore.

Generally I would prefer if we continued using Qt5 syntax.

chchwy:
A major issue of old style connection is the program still compiles when a typo occurs in signal/slot names. Developers will only know that until the connection fails in the runtime, not in the compile time. Other than that, old style connection is good.

We use QSpinBox::valueChanged a lot, so I think we have to pay more attention on this and make sure it  is working perfectly. 

@martinvanzijl What Qt version are you using? Is it Qt 5.6 and gcc 4.8?

Could you please test the following things?

1. use the syntax @CandyFace just mentioned
```
void (QSpinBox::*spinBoxValueChanged)(int) = &QSpinBox::valueChanged;
connect(mFpsBox, spinBoxValueChanged, this, &TimeControls::fpsClick);
```

2. Don't do signal to signal connection. Let's do signal to slot, and emit the singal in the slot.
```
void (QSpinBox::*spinBoxValueChanged)(int) = &QSpinBox::valueChanged;
connect(mFpsBox, spinBoxValueChanged, this, &TimeControls::fpsValueChanged);
...
void TimeControls::fpsValueChanged(int fps)
{
    emit fpsClick(fps);
}
```

If none of them work, probably we have to fall back to Qt4 style connection.

martinvanzijl:
@chchwy I have Qt 5.9 and gcc 4.8.

I'm afraid neither of the two suggestions above work for me. Both give the following output:

`QObject::connect: signal not found in QSpinBox`

chchwy:
Thanks @martinvanzijl 

Alright, I think we have to do Qt4 style connections.

Could you please search `QSpinBox::valueChanged` in the entire project and make sure all connection are working? The signal is used in `colorinspector.cpp`, `importimageseqdialog.cpp`, `preferencesdialog.cpp` and `tooloptionwidget.cpp`.

And we use `QDoubleSpinBox::valueChanged` in `tooloptionwidget.cpp`. Could you check them as well?

CandyFace:
We should investigate why this happens instead of jumping to conclusions that it's the connect syntax, @chchwy can't you make a small sample and let @martinvanzijl test it? 

again if this is a bug, then we should report it.

chchwy:
Mmm...currently, it seems to be a Qt's bug. So let's report it. But before Qt fix it, we need to do something, not just wait for Qt to fix it. 

It worth a try with the lambda syntax, but since the error message said "signal not found in QSpinBox", not slot not found, I'm afraid that will not work. @martinvanzijl could you test this?

The issue we faced is new syntax doesn't work. so no matter how many advantages it has, we won't accept the code which is beautiful but doesn't work.

For the compile-time check, we can put a "assert()" after the connection, like this:
```
auto conn = connect(mFpsBox, SIGNAL(valueChanged(int)), this, SIGNAL(fpsClick(int));
Q_ASSERT(conn == true);
```
So we will know that connection failure in the first run, it can greatly reduce the bug with signal/slot name typo.

CandyFace:
Have you been able to replicate it in a sample project though? if not, they'll likely not accept the report because they can't replicate it.

The weird thing here is that it works on windows, mac and from my understanding also some linux distributions so the problem may lie somewhere else. If it's is a bug in our codebase, then it's likely we're just working around the problem rather than fixing it properly and thus can hit this wall at some other point in which we'll really have to come up with a fix.

I agree that we should fix the problem and we shouldn't wait on a fix (if it's a qt bug) but I'm not sure that refactoring back to Qt4 style is the proper solution yet. 

@martinvanzijl does this happen with all overloaded signals or is it only some?

martinvanzijl:
The problem seems to happen on all QSpinBox value-changed instances. I tested the following: 

```
ColorInspector::ColorInspector(QWidget *parent) :
...
    connect(ui->RedspinBox, spinBoxChanged, this, &ColorInspector::onColorChanged);
    connect(ui->GreenspinBox, spinBoxChanged, this, &ColorInspector::onColorChanged);
    connect(ui->BluespinBox, spinBoxChanged, this, &ColorInspector::onColorChanged);
    connect(ui->AlphaspinBox, spinBoxChanged, this, &ColorInspector::onColorChanged);


void TimeControls::makeConnections()
...
    connect(mLoopStartSpinBox, spinBoxValueChanged, this, &TimeControls::loopStartValueChanged);
    connect(mLoopEndSpinBox, spinBoxValueChanged, this, &TimeControls::loopEndValueChanged);


GeneralPage::GeneralPage(QWidget* parent) :
...
    connect(ui->gridSizeInput, spinValueChanged, this, &GeneralPage::gridSizeChange);


FilesPage::FilesPage(QWidget* parent) :
...
    connect(ui->autosaveNumberBox, spinBoxValueChange, this, &FilesPage::autosaveNumberChange);


TimelinePage::TimelinePage(QWidget* parent) :
...
    connect(ui->timelineLength, spinBoxValueChange, this, &TimelinePage::timelineLengthChanged);


ToolsPage::ToolsPage(QWidget* parent) :
...
    connect(ui->onionMaxOpacityBox, spinBoxChanged, this, &ToolsPage::onionMaxOpacityChange);
    connect(ui->onionMinOpacityBox, spinBoxChanged, this, &ToolsPage::onionMinOpacityChange);
    connect(ui->onionPrevFramesNumBox, spinBoxChanged, this, &ToolsPage::onionPrevFramesNumChange);
    connect(ui->onionNextFramesNumBox, spinBoxChanged, this, &ToolsPage::onionNextFramesNumChange);
```

All gave the warning "QObject::connect: signal not found in QSpinBox", and changing the spin-box value didn't trigger the slot.

However the Q**Double**SpinBox connections in tooloptionwidget.cpp work OK:

```
void ToolOptionWidget::makeConnectionToEditor(Editor* editor)
...
    auto spinboxValueChanged = static_cast<void (QDoubleSpinBox::*)(double)>(&QDoubleSpinBox::valueChanged);
    connect(ui->brushSpinBox, spinboxValueChanged, toolManager, &ToolManager::setWidth);
    connect(ui->featherSpinBox, spinboxValueChanged, toolManager, &ToolManager::setFeather);
```

So only the _integer_ QSpinBox signals are affected. QDoubleSpinBox is OK.

chchwy:
Sounds like it's a special case just for QSpinbox. Because `QDoubleSpinbox::valueChanged()` is overloaded, but it doesn't have any issue. What do you think @CandyFace?

CandyFace:
Having confirmed that it seemingly happens to all QSpinboxes, I think that now we can suspect that's it's a bug in Qt and it should be reported. However before we do that we'll still need to verify our suspicion by making a small sample project that can reproduce the problem on the platform.

As for the fix, let's go back to Qt4 connect syntax for all **overloaded** functions for now.

chchwy:
@CandyFace Wow, you mean all overloaded signals? not only QSpinBox?

chchwy:
Hey guys, I installed Ubuntu 14.04 64bit with Qt 5.9.4 and g++ 4.8.4
The spinbox signal is working, either in pencil2d or other small projects.
How could you replicate it?

CandyFace:
@chchwy yeah to make it consistent we should do it for all overloaded signals if we intend to go back to qt4 syntax. It's odd though that you can't seem to replicate the problem... 
@martinvanzijl have you tried reinstalling qt completely?

scribblemaniac:
@martinvanzijl I have also seen some suggestions that a clean build may help, so try that if you get a chance.

martinvanzijl:
Hi guys, I checked out the latest master - the spin-boxes work OK for me now. I also made a small sample project, and the spin-box signals work fine there, too. So I did some digging and found the following...

In the old master before my patch (i.e. 45d71fedbb2d7c7e4ab2e2c0727afed668c57cc9 on 20 Feb), `spinBoxValueChanged` was declared as a static global variable in 'timecontrols.cpp':

```
static auto spinBoxValueChanged = static_cast<void (QSpinBox::*)(int)>(&QSpinBox::valueChanged);
```

In the current master (373b31bd9ea53d1e9a93591c1ad4d05336188187 on 02 March) this has been moved to a (non-static) local variable inside `TimeControls::makeConnections()`:

```
void TimeControls::makeConnections()
{
...
    auto spinBoxValueChanged = static_cast<void (QSpinBox::*)(int)>(&QSpinBox::valueChanged); <-- HERE
    connect(mLoopStartSpinBox, spinBoxValueChanged, this, &TimeControls::loopStartValueChanged);
    connect(mLoopEndSpinBox, spinBoxValueChanged, this, &TimeControls::loopEndValueChanged);
```

...and this works OK.

If I check out the old master (45d71fedbb2d7c7e4ab2e2c0727afed668c57cc9) and simply make this one change, all the spin-boxes work fine. Even e.g. the ColorInspector ones which are in a completely separate CPP file. It seems that global in 'timecontrols.cpp' messed up the signal everywhere.

I also tried some combinations in my small sample project and found:

1. Global does not work (either static or non-static). e.g.

```
static auto spinBoxChangedGlobal = static_cast<void(QSpinBox::*)(int)>(&QSpinBox::valueChanged);

MainWindow::MainWindow(...
{
...
    connect(ui->spinBoxFrom, spinBoxChangedGlobal, this, &MainWindow::updateLabel);
```

2.  Local static does not work. e.g.

```
MainWindow::MainWindow(...
{
...
    static auto spinBoxChangedLocal = static_cast<void(QSpinBox::*)(int)>(&QSpinBox::valueChanged);
    connect(ui->spinBoxFrom, spinBoxChangedLocal, this, &MainWindow::updateLabel);
```

3.  Local (non-static) works OK. e.g.

```
MainWindow::MainWindow(...
...
    auto spinBoxChangedLocal = static_cast<void(QSpinBox::*)(int)>(&QSpinBox::valueChanged);
    connect(ui->spinBoxFrom, spinBoxChangedLocal, this, &MainWindow::updateLabel);
```

4.  And, strangely, if I have a global declared, using a separate local variable won't work! e.g.

```
static auto spinBoxChanged_GLOBAL = static_cast<void(QSpinBox::*)(int)>(&QSpinBox::valueChanged); // declared but not used

MainWindow::MainWindow(...
{
...
    auto spinBoxChanged_LOCAL = static_cast<void(QSpinBox::*)(int)>(&QSpinBox::valueChanged);
    connect(ui->spinBoxFrom, spinBoxChanged_LOCAL, this, &MainWindow::updateLabel); // should work but doesn't!
```

That probably explains why the suggestions I tried earlier failed - I had kept this global intact and just used local variables with different names...

@chchwy Could you please try those four cases above on your Ubuntu and see whether you get the same results as me? I can send you my small sample project if you like.

chchwy:
Good detective work! @martinvanzijl 
Now it sounds more like a compiler's bug rather than Qt's bug.

Anyway, it's good to know it is working. I'll do the test tonight.

chchwy:
Confirmed I got extactly the same results from case 1 to case 4 as yours. Case 3 is really weird but happened in front of my eyes. 

@martinvanzijl  so does it mean the `QSpinBox::editingFinished` is no lonager needed?

martinvanzijl:
@chchwy Yes, we can remove the `QSpinBox::editingFinished` - it is no longer needed.

Issue #857, opened 2018-01-13
[Enhancement] Implement a Progress Bar for Importing Sequence Images


Kaikogf:
Implement a **_Progress Bar_** for Importing any length of Sequence Images, the same as the Progress Bar already implemented for Exporting Image Sequence (I noticed it was missing when I imported a long animation and as there were no sign of action I thought Pencil2D had stopped working !
;)

Last build 12JAN2018) win8.1x64bits

Kaikogf:
Thank you. It is working ok!

[Pencil build 23Feb2018]

PR #899, opened 2018-02-22
Progress bar for importing image sequences

martinvanzijl:
Fix for #857. Adds a progress dialog when importing image sequences, just like the one for exporting.

Note: The auto-save prompt is disabled while doing the import.

import-dialog-demo
.

### DIFF
--- a/app/src/mainwindow2.cpp
+++ b/app/src/mainwindow2.cpp
@@ -715,8 +715,8 @@ void MainWindow2::importImageSequence()
     progress.show();
 
     int totalImagesToImport = files.count();
+    progress.setMaximum(totalImagesToImport);
     int imagesImportedSoFar = 0;
-    int progressMax = 100;
 
     for (QString strImgFile : files)
     {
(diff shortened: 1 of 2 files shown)


app/src/mainwindow2.cpp
```diff
@@ -699,8 +702,22 @@ void MainWindow2::importImageSequence()
...
     }
 
+    // Flag this so we don't prompt the user about auto-save in the middle of the import.
+    mIsImportingImageSequence = true;
+
     QStringList files = imageSeqDialog->getFilePaths();
     int number = imageSeqDialog->getSpace();
+
+    // Show a progress dialog, as this can take a while if you have lots of images.
+    QProgressDialog progress(tr("Importing image sequence..."), tr("Abort"), 0, 100, this);
```
chchwy: You can use the `totalImagesToImport` as the maximum value of progress bar instead of 100, so the Qt progress bar will do the % calculation for you. And updating the progress by `progress.setValue((imagesImportedSoFar);`, no need to do `progress.setValue((imagesImportedSoFar + 1)*progressMax / totalImagesToImport);`

Not a big issue, just tell a simpler way.

martinvanzijl:
@scribblemaniac @chchwy Thanks for the reviews. I've made those changes now as suggested.

Sorry if this is a stupid question; but how do I add my updated code to the PR?

scribblemaniac:
You just commit the changes and push it to the same branch as the rest of your changed. The PR will be updated automatically.

martinvanzijl:
Thanks, I've done that now.

PR #901, opened 2018-02-22
Add non-API documentation to Doxygen input

J5lx:
As I previously mentioned on IRC, it is possible to integrate our existing non-API developer documentation with Doxygen as well, which allows us to have all documentation aimed at developers under one roof. This is what this PR does. It also includes the page about navigating the code from the wiki. As with #900, take this as a suggestion, any opinions are welcome.

Some notes:

- Unfortunately Doxygen requries links to be relative to the project root, so they will no longer work on GitHub with this
- For Doxygen to generate links to class members, these class members must already have documentation or EXTRACT_ALL must be turned on
- Currently we have autolinking enabled in our Doxygen configuration, which leads to some… interesting results:

  screenshot from 2018-02-22 18-38-54

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -790,7 +790,7 @@ WARN_LOGFILE           =
 # spaces. See also FILE_PATTERNS and EXTENSION_MAPPING
 # Note: If this tag is empty the current directory is searched.
 
-INPUT                  = app core_lib tests README.md
+INPUT                  = app core_lib tests docs README.md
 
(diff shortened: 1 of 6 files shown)


chchwy:
Dunno why github said there is a conflicting file, merged by cherry-pick.

J5lx:
Hmm maybe because Doxyfile-Travis got moved to util/docs by #900.

PR #904, opened 2018-02-24
Improve Movie Export (again)

scribblemaniac:
This builds on top of #876. It uses rawvideo format instead of BMP, which is faster. Also this adds support for APNG export and a looping option (for GIF and APNG formats only). There was also a minor issue with the import/export dialog where it would load the most recently used path, but would open the filebrowser with the default extension (rather than the most recently used extension); this has been fixed.


core_lib/src/movieexporter.cpp
```diff
@@ -479,12 +485,10 @@ Status MovieExporter::generateMovie(
...
-            QByteArray imgData;
-            QBuffer buffer(&imgData);
-            // BMPs are used because QT encodes them faster than other formats (less compression)
-            bool bSave = imageToExport.save(&buffer, "BMP", 100);
-            Q_ASSERT(bSave);
-            ffmpeg.write(imgData);
+            // Should use sizeInBytes instead of byteCount to support large images,
+            // but this is only supported in QT 5.10+
+            bytesWritten = ffmpeg.write(reinterpret_cast<const char*>(imageToExport.constBits()), imageToExport.byteCount());
+            Q_ASSERT(bytesWritten == imageToExport.byteCount());
```
J5lx: According to docs, QIODevice::write can return -1 when an error occurs. I suppose if that happened (maybe when FFmpeg dies while being written to or something?) it would trigger the assertion and potentially crash the program. I believe it would be better to account for such potential errors.

core_lib/src/movieexporter.cpp
```diff
@@ -598,12 +606,8 @@ Status MovieExporter::generateGif(
...
         obj->paintImage(painter, currentFrame, false, true);
 
-        QByteArray imgData;
-        QBuffer buffer(&imgData);
-        // BMPs are used because QT encodes them faster than other formats (less compression)
-        bool bSave = imageToExport.save(&buffer, "BMP", 100);
-        Q_ASSERT(bSave);
-        ffmpeg.write(imgData);
+        bytesWritten = ffmpeg.write(reinterpret_cast<const char*>(imageToExport.constBits()), imageToExport.byteCount());
+        Q_ASSERT(bytesWritten == imageToExport.byteCount());
```
J5lx: Same here.

core_lib/src/movieexporter.cpp
scribblemaniac: How exactly should it be handled? If it fails, retrying would be fairly pointless as it's not likely to suddenly work the second time. Throwing an error to the user would be ideal, but at the moment there is no error handling/dialog for any export errors, so to handle all errors would require more work beyond these lines. With the asserts, at least we know that part is failing. During releases, it will not crash the program. Adding full error handling to the export is on my todo list, but I think could be left for separate PR later.

J5lx: Hmm, fair points. Maybe leave a TODO though or make an issue to avoid forgetting about it.

scribblemaniac:
@chchwy I would say so.

Issue #800, opened 2017-11-15
Roadmap of 0.6.x and follow-up tasks


chchwy:
Hey guys,
@Jose-Moreno @CandyFace  @scribblemaniac  @J5lx  @feeef  @Kaikogf @gordielachance

In the days of waiting for 0.6 official release, it's time to discuss what we can do next in 0.6.x patches.

This roadmap will more focus on stability, robustness and performance issues. I would like to put new features into 0.7 rather than a 0.6.x patch. Still, any suggestion/discussion are welcomed.

# 0.6.1 Reduce memory usage

Currently, Pencil2d keeps all frames in RAM while loading a project. The upside is obvious, it's simple, efficient and intuitive to program. But on the other hand, the downside is obvious, too. It consumes a lot of memory. For example, a 200-frame animation may take up to 1.6GB ram if every frame is in full size of 1920x1080 (8MB ram each). The total memory consumption is considerable.

We've heard some cases said Pencil2D became very unstable after creating more than 700 frames. I suspect that's because Pencil2D has used up all the available RAM. Especially on a 32bit system, the limit of RAM is more restricted (on a 32bit Windows it's only 1.5GB per process as I know).

So the plan is to use the disk as much as possible. Pencil2D will keep maybe only 10 or 20 active frames beside the current frame and save all other frames to disk. I might use methods like [LRU](https://en.wikipedia.org/wiki/Cache_replacement_policies#LRU) to decide which frames Pencil2D should keep, but the main idea is to reduce the number of active frames in RAM. It may cause a sudden short pause in a long frame jump (e.g. jump from frame 1 to frame 100) due to the disk loading time of images, but we'll see.

Related issues: #253 #520 #574

# 0.6.2 Crash recovery

Another advantage of writing frames to disk is, it will be easier to implement a decent crash recovery since almost all frames will not get lost when pencil2d accidentally crashes.

At the moment Pencil2D creates a temporary workspace in system's temp folder while loading a project. (The temporary workspace can be thought of as an uncompressed pclx) The temporary workspace will be wipe out when Pencil2D exits normally. So if there is something left, it means Pencil2D was not properly closed last time. And then we can try to restore the project in the next start.

Related issues: #782

# 0.6.3 Stroke performance

The issue has been [long recognized](https://github.com/pencil2d/pencil/issues/486) but hasn't been resolved yet. There are some inefficient operations when Pencil2d handles pen strokes.

[...]

PR #905, opened 2018-02-27
Lowering memory usage

chchwy:
As I said in #800, improving the memory management is one of the goals in 0.6.1.

The test project has 1321 frames, 1 bitmap layer, 1 vector layer and 1 camera layer, camera resolution is 800x600.

Please see the results:
This is the memory usage graph with this PR, captured by Visual Studio's diagnostic tool: 
Imgur

Comparing to the official 0.6 build:
Imgur

The amount of memory used during loading a project and playing animation are both greatly improved. Also large projects load/save is significantly faster. I'm gonna to explain more technical details later.


core_lib/src/graphics/bitmap/bitmapimage.cpp
```diff
@@ -85,6 +85,21 @@ BitmapImage* BitmapImage::clone()
     return new BitmapImage(*this);
 }
 
+void BitmapImage::loadFile()
+{
+    mImage = std::make_shared<QImage>(fileName());
```
scribblemaniac: Maybe you should add a check here to make sure that `mImage == nullptr || !isModified()`. If this is accidentally called from somewhere else when this is not a nullptr, then changes to the image could be overwritten which is almost certainly not the desired behavior.

core_lib/src/interface/editor.cpp
```diff
@@ -73,12 +65,15 @@ Editor::Editor(QObject* parent) : QObject(parent)
     clipboardBitmapOk = false;
     clipboardVectorOk = false;
     clipboardSoundClipOk = false;
+
+    mActiveFramePool.reset(new ActiveFramePool(200));
```
scribblemaniac: Not sure if I like the constant size here. People with better/worse machines may want this adjusted, although at the same time I don't want to add unnecessary complexity to the preferences UI so maybe this is fine. How did you arrive at the number 200?

core_lib/src/graphics/bitmap/bitmapimage.cpp
```diff
@@ -85,6 +85,21 @@ BitmapImage* BitmapImage::clone()
...
+    mBounds.setSize(mImage->size());
+    qDebug() << "Load file=" << fileName();
+}
+
+void BitmapImage::unloadFile()
+{
+    if (isModified() == false)
```
scribblemaniac: Could we do setModified(false) when the image is saved and have a check in setModified that would reset mImage if the argument is false and the activeframepool does not contain the image? As it stands now, memory consumption is going to continue to increase for the lifetime of the program while making modifications, but if we did this, then all the old modified frames would removed from memory when saved.

core_lib/src/interface/editor.cpp
```diff
@@ -494,6 +489,21 @@ void Editor::updateAutoSaveCounter()
     }
 }
 
+void Editor::updateActiveFrames(int frame)
+{
+    int beginFrame = std::max(frame - 3, 1);
```
scribblemaniac: Setting this to frame during playback might help speed things up a little bit since you only need to load frames in front of the current frame. On a similar note, have you considered loading the frames at the beginning of a range when you are close to the end of the range so that looping is smooth?

core_lib/src/activeframepool.cpp
```diff
@@ -0,0 +1,73 @@
...
+        return;
+
+    Q_ASSERT(key->pos() > 0);
+
+    auto it = mCacheFramesMap.find(key);
+    mCacheFramesList.push_front(key);
+    if (it != mCacheFramesMap.end())
+    {
+        mCacheFramesList.erase(it->second);
+        mCacheFramesMap.erase(it);
```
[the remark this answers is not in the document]
chchwy: You're right. it's redundant.

core_lib/src/graphics/bitmap/bitmapimage.cpp
chchwy: Good advice. will do that.

core_lib/src/interface/editor.cpp
chchwy: An HD image take about 7.91MB (1920x1080x32bit = 7.91MB) 
200 frames x 7.91MB = 1582MB ~= 1.5G

In 32bit Windows, one process can manage up to 2GB memory, 500MB reserved by kernel space, rest part is about 1.5GB in user space.

It's a conservative approach to this issue. A better idea is to detect the os architecture and memory size of user's PC, for example, the user has 16GB ram and 64bit OS, and then set a higher value. 

Since there is no way to know the actual memory size in Qt, we have to do it in a native way for each platform.

core_lib/src/interface/editor.cpp
chchwy: Currently, we load frames in `canvaspainter.cpp` line 208, just before the frame gets drawn. ActiveFramePool only does unloading. A frame in ActiveFramePool doesn't mean it will get loaded.

I'm still thinking how to do it in a better way.

core_lib/src/activeframepool.cpp
```diff
@@ -0,0 +1,73 @@
+#include "activeframepool.h"
+#include "keyframe.h"
+#include <QDebug>
+
+
+ActiveFramePool::ActiveFramePool(int maxSize)
+{
+    Q_ASSERT(maxSize > 10);
```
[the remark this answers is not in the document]
chchwy: It's a random number tbh, just make sure the size is not too small or the app will be busy in loading/unloading things all the time. :p

core_lib/src/graphics/bitmap/bitmapimage.cpp
chchwy: We do `setModified(false)` in `layerbitmap.cpp` line 78, in the `LayerBitmap::saveKeyFrameFile()` function. 
All modified frames will be written back to disk when auto-save is triggered. And we only do that with modified frames, so the auto-save will be a lot faster now.

scribblemaniac: That's good then we don't have to change that. What I was mainly trying to say is that those frames that are written to disk should be unloaded if they are not in the active pool. BitmapImage::unloadFile is only called when it's removed from the activethreadpool, but it won't be removed from memory if it is in a modified state at that time. If after that, a save or an autosave writes those changes to disk and removes the modified state, the frame will remain loaded unnecessarily.

core_lib/src/interface/editor.cpp
scribblemaniac: Ah I see. Well how about loading the frames in ActiveFramePool::put? That class already handles the unloading anyway.

core_lib/src/graphics/bitmap/bitmapimage.cpp
chchwy: right, any good suggestion for that?

core_lib/src/interface/editor.cpp
chchwy: Right, we can do that. And probably it's a good idea to preload the frames.

core_lib/src/graphics/bitmap/bitmapimage.cpp
scribblemaniac: Well off the top of my head we could do something like this. Emit a modified signal when a keyframe is modified (from setModified/modification). Then:
```
void ActiveFramePool::unloadFrame(KeyFrame* key) {
    ...
    if(key->isModified()) {
        connect(key, &KeyFrame::modified, this, &ActiveFramePool::handler);
    }
    key->unloadFile();
}
```
The handler function would check if the new modified state is false and if the map does not contain the sender keyframe, and if these conditions are met it would call unloadFile again.

chchwy:
This is an overview of lowering memory usage work for anyone interested in technical details. Please note, the work has two parts. The first part is in commit b24f9ce6, and this PR is the second part.

### Lazy loading

Instead of loading all bitmap images into memory on opening a project, lazy loading was introduced into pencil2d. With lazy loading, a bitmap image will only be loaded and occupy memory space until the point which it is needed. If you take a look at `bitmapimage.cpp` line 53, the BitmapImage holds an image's file path but doesn't load it. The actual place of loading an image is in `canvaspainter.cpp` line 208, right before the image gets drawn.

So the app can run in minimal memory footprint when a user opens a large project, usually only current frame and onion skins are needed. It also speeds up the file loading time.

### Lazy saving

In version 0.6, pencil2d writes all bitmap/vector frames to disk when users hit "save" button. It takes considerable time to write thousands of png/xml files with a large project. So instead of writing all frames, we can write only those modified frames.

In order to do that, we use an `IsModified` flag in class `KeyFrame`. The flag will be raised whenever a keyframe is modified,  (e.g., `BitmapImage::paste()`). And in the `LayerBitmap::needSaveFrame()`, it checks the flag to determine whether a keyframe needs to be written to disk during the saving process. This will save a lot of time from not writing unmodified frames, and make auto-save less annoying.

### Active Frame Pool

And the last part is introducing Active Frame Pool. It implements a [LRU cache](https://en.wikipedia.org/wiki/Cache_replacement_policies#Least_Recently_Used_(LRU)) to keep tracking the most recent accessed keyframe and unloads the keyframes which weren't accessed for a while.

Curretnly, pencil2d only keeps at most 200 keyframes in memory. If pencil2d try to load more frames, the active frame pool discards the least recently used keyframes to keep memory usage low.

Generally I focused on bitmap layers first and don't do much on vector layers because I noticed vector layers use much less memory than bitmap layers. But these concepts can apply to vectors layers as well.

PR #914, opened 2018-03-05
Developer documentation additions

J5lx:
This is basically a followup to #901. Most notably I added a proper main page (which is heavily based on the current wiki main page), but there’s also information on build options now and I added some words regarding #896. With this I’d say we can consider pointing potential new contributors straight to the generated Doxygen documentation and reserve the wiki for meta stuff that is not directly related to the development of the program.


chchwy:
Cool @J5lx

PR #917, opened 2018-03-10
Fix build with Qt 5.10

J5lx:
Now necessary because of 7550fff4bc37ea596f5edd8cf53ebf59fa9d480a.

### DIFF
--- a/core_lib/src/movieexporter.cpp
+++ b/core_lib/src/movieexporter.cpp
@@ -460,6 +460,8 @@ Status MovieExporter::generateMovie(
 
     // Run FFmpeg command
 
+    // Attention: Lots of unnecessary code duplication ahead because Microsoft’s preprocessor is stupid
+#if QT_VERSION < QT_VERSION_CHECK(5, 10, 0)
     STATUS_CHECK(executeFFMpegPipe(strCmd, progress, [&](QProcess& ffmpeg, int framesProcessed)
     {
         if(framesProcessed < 0)
@@ -495,6 +497,41 @@ Status MovieExporter::generateMovie(
 
         return false;
     }));
+#else
+    STATUS_CHECK(executeFFMpegPipe(strCmd, progress, [&](QProcess& ffmpeg, int framesProcessed)
+    {
+        if(framesProcessed < 0)
+        {
+            failCounter++;
+        }
+
+        if(currentFrame > frameEnd)
+        {
+            ffmpeg.closeWriteChannel();
+            return false;
+        }
+
+        if((currentFrame - frameStart <= framesProcessed + frameWindow || failCounter > 10) && currentFrame <= frameEnd)
+        {
+            QImage imageToExport = imageToExportBase.copy();
+            QPainter painter(&imageToExport);
+
+            painter.setWorldTransform(view * centralizeCamera);
+            painter.setWindow(QRect(0, 0, camSize.width(), camSize.height()));
+
+            obj->paintImage(painter, currentFrame, false, true);
+
+            bytesWritten = ffmpeg.write(reinterpret_cast<const char*>(imageToExport.constBits()), imageToExport.sizeInBytes());
+            Q_ASSERT(bytesWritten == imageToExport.sizeInBytes());
+
+            currentFrame++;
+            failCounter = 0;
+            return true;
+        }
+
+        return false;
+    }));
+#endif
 
     return Status::OK;
 }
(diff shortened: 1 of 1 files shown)


J5lx:
Actually, don’t merge this yet, seems like Microsoft’s preprocessor has problems with that syntax.

CandyFace:
Is this simply to avoid more warnings or is there an actual reason to use sizeInBytes over byteCount? I can't imagine we'll ever have an image larger than 2 gb, so is this necessary at all?

J5lx:
byteCount was deprecated in 5.10, and because @chchwy disabled deprecated symbols in 7550fff4bc37ea596f5edd8cf53ebf59fa9d480a, movieexporter.cpp can no longer compile when using byteCount. However sizeInBytes was only introduced in 5.10, so we still need to use byteCount when compiling with earlier versions of Qt.

J5lx:
FWIW, here are the errors generated without this change (as reported by clang):

```
../../core_lib/src/movieexporter.cpp:489:113: error: no member named 'byteCount' in 'QImage'
            bytesWritten = ffmpeg.write(reinterpret_cast<const char*>(imageToExport.constBits()), imageToExport.byteCount());
                                                                                                  ~~~~~~~~~~~~~ ^
../../core_lib/src/util/pencilerror.h:85:17: note: expanded from macro 'STATUS_CHECK'
        { Status st = (x); if ( !st.ok() ) { return st; } }
                       ^
../../core_lib/src/movieexporter.cpp:490:52: error: no member named 'byteCount' in 'QImage'
            Q_ASSERT(bytesWritten == imageToExport.byteCount());
                                     ~~~~~~~~~~~~~ ^
/usr/include/qt/QtCore/qglobal.h:732:30: note: expanded from macro 'Q_ASSERT'
#    define Q_ASSERT(cond) ((cond) ? static_cast<void>(0) : qt_assert(#cond, __FILE__, __LINE__))
                             ^~~~
../../core_lib/src/util/pencilerror.h:85:17: note: expanded from macro 'STATUS_CHECK'
        { Status st = (x); if ( !st.ok() ) { return st; } }
                       ^
../../core_lib/src/movieexporter.cpp:608:109: error: no member named 'byteCount' in 'QImage'
        bytesWritten = ffmpeg.write(reinterpret_cast<const char*>(imageToExport.constBits()), imageToExport.byteCount());
                                                                                              ~~~~~~~~~~~~~ ^
../../core_lib/src/util/pencilerror.h:85:17: note: expanded from macro 'STATUS_CHECK'
        { Status st = (x); if ( !st.ok() ) { return st; } }
                       ^
../../core_lib/src/movieexporter.cpp:609:48: error: no member named 'byteCount' in 'QImage'
        Q_ASSERT(bytesWritten == imageToExport.byteCount());
                                 ~~~~~~~~~~~~~ ^
/usr/include/qt/QtCore/qglobal.h:732:30: note: expanded from macro 'Q_ASSERT'
#    define Q_ASSERT(cond) ((cond) ? static_cast<void>(0) : qt_assert(#cond, __FILE__, __LINE__))
                             ^~~~
../../core_lib/src/util/pencilerror.h:85:17: note: expanded from macro 'STATUS_CHECK'
        { Status st = (x); if ( !st.ok() ) { return st; } }
                       ^
4 errors generated.
```

J5lx:
Alright I updated the PR so it works with all compilers now. Turns out Microsoft’s preprocessor is just plain stupid, and the only way to work around that here is apparently by introducing lots of otherwise unnecessary code duplication. If anyone has a better solution, please let me know.

CandyFace:
@chchwy Wouldn't it be better to revert the deprecated symbol defines instead? I don't think we should use the latest Qt unless necessary. The build server is also fixed at Qt 5.9 so I don't see why we should introduce 5.10 stuff anyway. 

Given that we support down to Qt 5.6, this could get really messy fast if they suddenly start deprecating things.

J5lx:
Sure, that’d be fine by me, though we should hear what @chchwy thinks about that, since he introduced that policy. Maybe we could also simply change it to `QT_DISABLE_DEPRECATED_BEFORE=0x051000`, which would allow compiling with 5.10 but still generate errors with stuff deprecated in earlier versions.

The reason why I use 5.10 is because I’m on Arch, which is a rolling release distro and thus always provides the latest versions of everything. For me the only way of developing againtst 5.9 or earlier would be compiling those releases from source, and you know yourself how annoying that is.

chchwy:
Ah sorry guys, didn't expect the change screws things up.

@J5lx I don't have Qt 5.10 installed. 
Do you think it's better to have an `imageByteCount()` to avoid duplicated code?
```
uint64_t imageByteCount(const QImage& img)
{
#if QT_VERSION < QT_VERSION_CHECK(5, 10, 0)
    return img.byteCount();
#else
    return img.sizeInBytes();
#endif
}
``` 

@CandyFace I agreed with you that we can remove the deprecated symbol. Since we stick with Qt 5.6 & 5.9, it is not necessary to make it fail to compile. 

So probably let's keep `QT_DEPRECATED_WARNINGS`, and change the second DEFINE to  `QT_DISABLE_DEPRECATED_BEFORE=0x050600` (Qt 5.6). What do you think?

CandyFace:
@chchwy sounds good to me 👍

J5lx:
As long as it compiles with 5.10, I’m fine with either solution.

chchwy:
Disabled functions deprecated in Qt 5.6 in commit cb521ec

PR #922, opened 2018-03-13
A temporary solution for high dpi displays

chchwy:
A temporary solution for high DPI displays by setting `Qt::AA_EnableHighDpiScaling`.

I test it on an 15-inch MacBook Pro with Win 10 installed, resolution 2800x1800, 200% display scaling.

1. `Qt::AA_EnableHighDpiScaling` On:

Generally it's fine, but take a closer look at the play and add-key button, you will see pixelated icons.

Imgur

2. `Qt::AA_EnableHighDpiScaling` Off (as 0.6 official build)

Very tiny icons, and overlapped font on Timeline

Imgur

It's not a perfect solution but will make the Pencil2D UI *usable* on a 4k monitor.

### DIFF
--- a/app/src/main.cpp
+++ b/app/src/main.cpp
@@ -305,7 +305,7 @@ int handleArguments( PencilApplication& app )
 int main(int argc, char* argv[])
 {
     Q_INIT_RESOURCE(core_lib);
-
+    PencilApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
     PencilApplication app( argc, argv );
 
     installTranslator( app );
--- a/app/src/main.cpp
+++ b/app/src/main.cpp
@@ -305,7 +305,16 @@ int handleArguments( PencilApplication& app )
 int main(int argc, char* argv[])
 {
     Q_INIT_RESOURCE(core_lib);
-    PencilApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
+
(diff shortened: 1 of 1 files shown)


app/src/main.cpp
```diff
@@ -305,7 +305,16 @@ int handleArguments( PencilApplication& app )
 int main(int argc, char* argv[])
 {
     Q_INIT_RESOURCE(core_lib);
-    PencilApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
+
+    QSettings settings(PENCIL2D, PENCIL2D);
+    if (settings.value("EnableHighDpiScaling", "true").toBool())
```
CandyFace: Why do you add the EnableHighDpiScaling as a setting and check it if it's always true? Is it not enough to just set the attribute always?

chchwy: Just in case we want to disable it without changing the code as it's a temporary solution.

scribblemaniac:
Good work, just got around to testing this on a Linux machine with 4k:
high_dpi_linux
Pretty much everything except for the label texts are pixelated and blurry, but at least they are a reasonable size now. It also appears to scale the canvas, which is not really desirable. Even Qt icons such as the close button on QDockWidgets are scaled poorly. This is a good temporary fix though for 0.6.1 and is much more usable than 0.6.0 :+1:

PR #946, opened 2018-04-15
Install the icon in the right path

mapreri:
The icon currently in app/data/pencil2d.png is actually 256x256 px, so
it should be installed in the 256x256 dir, not the 128x128 one!

### DIFF
--- a/app/app.pro
+++ b/app/app.pro
@@ -144,7 +144,7 @@ linux {
     desktopentry.path = $${PREFIX}/share/applications
 
     icon.files = data/pencil2d.png
-    icon.path = $${PREFIX}/share/icons/hicolor/128x128/apps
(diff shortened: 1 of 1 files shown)


J5lx:
Nice catch! I even remember thinking about the implications of [the size change in 794def99551ff3b0ec64127d19860acc7c5888b8](https://github.com/pencil2d/pencil/commit/794def99551ff3b0ec64127d19860acc7c5888b8?short_path=02c74a4#diff-02c74a40c1fd043da32affc80f9bb32c) but then I just thought to myself it wouldn’t hurt because the new file was larger rather than smaller. So much for that!

mapreri:
The thanks go to https://lintian.debian.org/tags/icon-size-and-directory-name-mismatch.html that caught it while I was preparing the 0.6.1-1 upload to Debian ;)

PR #948, opened 2018-04-15
Remove .tiff (doesn't work) and add warning if file is not imported

CandyFace:
strImgFileLower was only used as a mean to check the file extension but never used with importImage(QString). The if statement is unnecessary because we filter the allowed file formats to be imported via pencildef.h

Prior to this, if the file format was read as invalid from the if statement, the for loop would simply continue. This could lead to files not being imported correctly.

I've also removed .tiff format because it doesn't appear to be working.

### DIFF
--- a/app/src/mainwindow2.cpp
+++ b/app/src/mainwindow2.cpp
@@ -721,21 +721,18 @@ void MainWindow2::importImageSequence()
     progress.setMaximum(totalImagesToImport);
     int imagesImportedSoFar = 0;
 
+    QString failedFiles;
+    bool failedImport = false;
     for (QString strImgFile : files)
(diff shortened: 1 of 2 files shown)


scribblemaniac:
> strImgFileLower was only used as a mean to check the file extension but never used with importImage(QString)

I'm pretty sure this was intentional. If the user want's to import `Image.png`, you don't want to import the potentially different file `image.png` on case-sensitive file systems.

scribblemaniac:
You should still remove the tif/tiff extension though (including in pencildef.h) or file an issue to try and find a solution. I can confirm that tiff sequence import does not appear to be working.

Issue #942, opened 2018-04-14
[UI] Show a "Remove Color Swatch" Button on the Color Palette UI


Jose-Moreno:
### --Issue Summary--
Currently there's only an "add swatch button", so organizing custom palettes for different projects is really cumbersome.

### --Video or Image Reference--
![image](https://user-images.githubusercontent.com/4912907/38764370-194775ec-3f73-11e8-97bd-791e890a1312.png)

CandyFace:
So i've started working on this: the implementation will be as following:

One can:
* delete a color using the delete button
* Select multiple colors to be deleted, by dragging or holding the equivalent button on the keyboard to select multiple 
![soon](https://user-images.githubusercontent.com/1045397/39959245-323f1700-560f-11e8-864a-58d9f9188842.gif)

* right-click where as a context-menu will appear with three options:
![screen shot 2018-05-12 at 18 05 40](https://user-images.githubusercontent.com/1045397/39959233-1d41f57a-560f-11e8-87d3-e1c576603cb4.png)

This goes into a bigger implementation of making the color palette static, as in not connected to strokes or anything, and adding a new window to keep track of "connected" colors for the vector layer. 
You can read more about my proposal here #962

Kaikogf:
Very must have implementation
:+1:

CandyFace:
Implemented and available in latest nightly build, closing.

Jose-Moreno:
It's working well. Good job! :slightly_smiling_face:

Issue #755, opened 2017-09-12
[ISSUE] Vector Layer will be kept active even when using a Bitmap layer


Kaikogf:
### --Issue Summary--
After using a vector layer and starting drawing/painting on a Bitmap layer, changes applied on the active layer (bitmap) will also affect some colors on the vector layer

### --Actual Results
When changing colours on the bitmap layer, it will affect the last colour applied on the vector layer

### --Expected Results--
Vector and bitmap layers are independent.

### --Steps to reproduce--
. Open Pencil2D
. Select VECTOR layer
. Select Pencil tool and check the "FILL CONTOUR" (for better visualization) and uncheck "show outlines only" feature  on the DISPLAY panel (for visualization of the shapes being filled.
. Draw 3 circles using different colours
. Select now the BITMAP layer and change the colour of the ColorWheel panel: as a result the last filled circle on the VECTOR layer will changed accordingly. It happens when using any other drawing tool

### --System Information--
+ Pencil2D Version:
Pencil2D last build 12 SET2017

+ Operating System:
Win8.1 x 64bits

+ RAM Size:
64GB

+ Graphics Tablet:
HUION GT185 HD

Jose-Moreno:
@Kaikogf @chchwy Hi. I'm sorry to repeat myself, but this is not a bug. If you use Toonboom Harmony 15 or even older versions, this will happen as well. It's intended behaviour when you work with global color palettes for even a small production. Vector color has to be fully editable. bitmap not so much as it's used for sketching or artistic painting.

I understand it's hard to get away from the way image editors like GIMP or Photoshop normally work. But the color palette in Pencil2d or Toonboom is NOT a swatch palette where colors are static. These colors are expected to be editable anytime, they are sort of "live" colors (similar to Illustrator) due to the volatile nature of an animation production where art direction decisions must be made last-minute.

Basically one have to use it like this:
1) For Bitmap, create a single swatch that will serve as your "all-in-one" color switcher.
2) For Vector create ONE color for every part that has a color attached as seen on your color model. 

For example if you have a black t-shirt, black pants and black shoes on your character, you HAVE to create 3 black colors, one corresponding to each part of the character.

The solution for this confusion is for the user to create and name specific color palettes to use for either bitmap and vector layers. 
Pencil2D team could think in the future of creating separate palette windows for both layers, or using visibility tags for color swatches that activate when either layer is active, but this is not a bug. I'm afraid it is actually a standard in animation production.

[...]

PR #961, opened 2018-05-10
Add colorsliders, UI and more...

CandyFace:
changes

+ Added colorslider widget
  + HSV, RGB colorsliders with extendability
+ Color wheel and color sliders are now separated
+ Remember color values from last session
+ Can modify existing swatches on bitmap (forced behaviour for now)
+ Ability to add colors to palette using the color wheel, and via the
dialog through a new button.
+ button styling
+ Added color palette preference icon
+ Improved performance of painting color square


Jose-Moreno:
@CandyFace Man great work! looks really nice I was wondering while you were tinkering with the palette, did you come across the code for this issue? #942 I think scribblemaniac said the code for the remove button was already there but it's just that the button disappeared from the palette, and it's really annoying to have to delete the colors via xml and reimport the palette :stuck_out_tongue_closed_eyes:

CandyFace:
@Jose-Moreno I considered adding the remove button again but it makes the application unstable because it doesn't work properly for vector yet. If you remove a swatch from the palette and switch to vector, then the application will crash. The problem as I see it is that everything is connected to the palette, so before I can bring back that feature, I'll have to fix it so it properly deletes for vector too.

CandyFace:
Having looked through this https://github.com/pencil2d/pencil/issues/755 issue, I'm quite certain it's been fixed in this PR too.

Modifying colors on vector and moving to bitmap will not affect each other.

chchwy:
Awesome @CandyFace! Will review all these PRs this weekend, great job!

chchwy:
Reviewing

app/src/colorpalettewidget.cpp
```diff
@@ -65,6 +66,25 @@ void ColorPaletteWidget::initUI()
     else
         setGridMode();
 
+    QColor savedColor;
+    savedColor.setRgba(settings.value("colorOfSliders").toUInt());
```
chchwy: @CandyFace Could I ask what "colorOfSliders" is for?

chchwy: Yup, I think it will be more consistent, not having 2 different color values in the system.

Also the front color may change when users load a new project, so the color inspector will need to get the new color when `Editor::objectLoaded()` signal is emitted.

CandyFace: Alright I'll change it later today

CandyFace: I've added and tested that it works now.

I'm not sure the implementation is the best way to go though. I've added a new connection that is called for the only purpose of loading our color on init. This avoids setColor changing our value back and fourth between values, as it may at some point get a value which it can't parse correctly.

It's the least messy way I can come up with for now, to avoid adding too much logic to setColor.

chchwy:
Well done @CandyFace

Issue #962, opened 2018-05-12
Connected colors


CandyFace:
Connected colors is proposal to organize vector strokes and to allow more freedom for the color palette.

#### what is a connected color?
A connected color is one that is attached to a stroke

#### Why?
Because currently there's no way of knowing if a color is attached to a stroke. This could possibly be solved differently, like marking the swatches with a symbol or dot or similar, but then there would still be other problems to solve like:
* How do we allow palette presets and show our attached strokes at the same time?
* What if you want your bitmap colors separated from your vector colors?

My solution solves both issues, without UI/UX sacrifice.

#### what this means for the color palette
Having a window for colors attached to vector strokes, means that can we make the color palette completely independent from the canvas, only to be used for easy accessibility of "favorite" colors.
This also means the ability to add palette presets, since vector strokes no longer depend on the colors in the palette

#### How it's going to work

The way you add colors to the list is by drawing on the canvas. Drawing with a different color will produce a new color item to the list, likewise if you remove a stroke from the canvas, the color attached to that stroke will be removed from the list.

![screen shot 2018-05-12 at 18 28 37](https://user-images.githubusercontent.com/1045397/39959466-63729fa6-5612-11e8-95bf-2f09bf5f231a.png)
As my beautiful sketch shows.
The window will have one button, a delete button to remove colors from the list. When a color has been deleted, it will become detached from its stroke. There could be several options here to what should happen to the stroke:
For example:
- Allow the stroke to keep the color but lose the ability to edit unless added again (possibly via a (+) button... an after thought)
- remove the stroke or strokes?
- remove the color (will become invisible line) or make it a default color to signify "detached"

Any thoughts on this?

CandyFace:
Some afterthoughts
![screen shot 2018-05-12 at 20 11 26](https://user-images.githubusercontent.com/1045397/39960304-c8c580e0-5620-11e8-8f49-b645b22cd794.png)

By default the fill color is none, which will be shown as an empty square or rectangle

Kaikogf:
"... As my beautiful sketch shows " :)

When I initially read your first post it came to my mind Krita and GreasePencil (Blender). They both seems to have similar behaviour as the one you are trying to implement withing Pencil2D. Krita for bitmap colours and GreasePencil for vector colours

[this didn't add anything for the discussion .... just wanted to write something ;) ]

CandyFace:
[...]

PR #971, opened 2018-05-19
Fix remove color button and add context menu

CandyFace:
it s back

I've fixed the remove button and added some enhancements.
* Add, replace or remove using right-click through the context-menu shown above
* Drag to select multiple color swatches
* no auto color update on palette anymore, use replace via context menu.

~~When palette is empty, vector layer will use front color instead of palette.~~

Known bugs:
modifying first index in palette, doesn't update vector stroke for some reason..
this is an old bug but I wasn't able to find the cause this time around.
Considering my proposal https://github.com/pencil2d/pencil/issues/962 that changes this, I'm leaving the bug as is for now.

### DIFF
--- a/app/src/colorpalettewidget.cpp
+++ b/app/src/colorpalettewidget.cpp
@@ -17,8 +17,10 @@ GNU General Public License for more details.
 #include "colorpalettewidget.h"
 #include "ui_colorpalette.h"
 
-#include <cmath>
+// Standard libraries
+#include "cmath"
 
+// Qt
 #include <QDebug>
 #include <QListWidget>
 #include <QListWidgetItem>
@@ -26,7 +28,9 @@ GNU General Public License for more details.
 #include <QColorDialog>
 #include <QToolBar>
 #include <QSettings>
+#include <QMenu>
 
+// Project
 #include "colordictionary.h"
 #include "colourref.h"
 #include "object.h"
@@ -50,15 +54,12 @@ ColorPaletteWidget::~ColorPaletteWidget()
 
 void ColorPaletteWidget::initUI()
 {
-    // "Remove color" feature is disabled because
-    // vector strokes that are linked to palette
-    // colors don't handle color removal from palette
-    ui->removeColorButton->hide();
-
     QSettings settings(PENCIL2D, PENCIL2D);
     int colorGridSize = settings.value("PreferredColorGridSize", 34).toInt();
 
     mIconSize = QSize(colorGridSize, colorGridSize);
+
+    ui->colorListWidget->setContextMenuPolicy(Qt::CustomContextMenu);
 
     QString sViewMode = settings.value("ColorPaletteViewMode", "ListMode").toString();
     if (sViewMode == "ListMode")
@@ -71,13 +72,19 @@ void ColorPaletteWidget::initUI()
                              "QPushButton:pressed { border: 1px solid #ADADAD; border-radius: 2px; background-color: #D5D5D5; }"
                              "QPushButton:checked { border: 1px solid #ADADAD; border-radius: 2px; background-color: #D5D5D5; }";
 
+
+    // color number need to be set to add new color to palette
+    // in cases where the last color was saved
+    // otherwise first color will be black.
+    editor()->color()->setColorNumber(0);
+    editor()->color()->setColor(editor()->color()->frontColor());
+
     ui->addColorButton->setStyleSheet(buttonStylesheet);
     ui->removeColorButton->setStyleSheet(buttonStylesheet);
     ui->colorDialogButton->setStyleSheet(buttonStylesheet);
 
     palettePreferences();
 
-    connect(ui->colorListWidget, &QListWidget::currentItemChanged, this, &ColorPaletteWidget::colorListCurrentItemChanged);
     connect(ui->colorListWidget, &QListWidget::itemClicked, this, &ColorPaletteWidget::clickColorListItem);
     connect(ui->colorListWidget, &QListWidget::itemDoubleClicked, this, &ColorPaletteWidget::changeColourName);
     connect(ui->colorListWidget, &QListWidget::currentTextChanged, this, &ColorPaletteWidget::onActiveColorNameChange);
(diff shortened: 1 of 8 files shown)


chchwy:
One more thing needs to find out: What should pencil2d do when users remove a color which is in use?

1. Open Pencil2D, go to a vector frame
2. Pick color `Red` from color palette
3. Use brush tool and draw something with `Red` color
4. Remove `Red` color from color palette

In step 4, the color `Red` which the brush stroke connects to no longer exists, so the brush stroke color should change, but it doesn't.

Go to another frame and then go back to the previous drawn frame, the red brush stroke becomes `Dark Red`, the next color in the palette. 

The behaviours are apparently not wanted.

A simple solution is telling users the color is in use, you can't remove it. This is what pencil2d 0.4.4 did. Or telling user they can bind the strokes to another palette color, not just silently bind to the next color.

CandyFace:
Non of those solutions are very good though, the user should be able to delete all colors in their palette and the vector stroke should not depend on that. To fix this properly though would require us to keep track of strokes and color that does not have a palette index anymore.

Considering the state of vector still, I ignored the problem but it's true that the user should have feedback as to what should happen, since the stroke is bound to a palette color.

CandyFace:
I don't think the strokes should depend on the palette because it conflicts with the standard palette behaviour as seen in any other drawing application. There's no way of telling which strokes are bound to what color, except the color itself. 

If the user decides to change the swatch color on bitmap for example, then the vector color will change too or lose its connection to palette, which is restrictive. The imo. more user friendly solution would be to keep the color palette static and keep track of vector strokes somewhere else.

Alternatively we could have a palette for both bitmap and vector.

The problem that we're currently facing:
The user will expect a similar behaviour no matter if they use vector or bitmap because there's nothing to suggest the contrary. so we could make it explicit that the color palette has different behaviours, depending on the layer you use or we could simply separate the behaviours, this is what I'm suggesting with #962

chchwy:
Mmm...It's a feature in Toom boom, take a look at this video:
https://www.toonboom.com/resources/video-tutorials/video/colour-palette-and-paint

I just read the #962, that's a very interesting idea, like a separate color palette just for vector layers. We can seriously consider implementing it.

CandyFace:
Is there anything you're missing from this PR @chchwy or can we merge?

I will implement the "vector palette" possibly as part of 0.6.3, so this should be an alright solution until then.

app/src/colorpalettewidget.cpp
```diff
@@ -71,20 +72,28 @@ void ColorPaletteWidget::initUI()
...
 
-    connect(ui->colorListWidget, &QListWidget::currentItemChanged, this, &ColorPaletteWidget::colorListCurrentItemChanged);
     connect(ui->colorListWidget, &QListWidget::itemClicked, this, &ColorPaletteWidget::clickColorListItem);
     connect(ui->colorListWidget, &QListWidget::itemDoubleClicked, this, &ColorPaletteWidget::changeColourName);
     connect(ui->colorListWidget, &QListWidget::currentTextChanged, this, &ColorPaletteWidget::onActiveColorNameChange);
 
     connect(ui->addColorButton, &QPushButton::clicked, this, &ColorPaletteWidget::clickAddColorButton);
     connect(ui->colorDialogButton, &QPushButton::clicked, this, &ColorPaletteWidget::clickColorDialogButton);
     connect(ui->removeColorButton, &QPushButton::clicked, this, &ColorPaletteWidget::clickRemoveColorButton);
+    connect(ui->colorListWidget, &QListWidget::itemSelectionChanged, this, &ColorPaletteWidget::selectedItems);
```
chchwy: Is this connect() necessary? in `selectedItems()` it seems doing nothing.

app/src/colorpalettewidget.cpp
```diff
@@ -17,16 +17,20 @@ GNU General Public License for more details.
 #include "colorpalettewidget.h"
 #include "ui_colorpalette.h"
 
-#include <cmath>
+// Standard libraries
+#include "cmath"
```
chchwy: minor thing, but usually system header uses angle brackets

app/src/colorpalettewidget.cpp
```diff
@@ -71,20 +72,28 @@ void ColorPaletteWidget::initUI()
                              "QPushButton:pressed { border: 1px solid #ADADAD; border-radius: 2px; background-color: #D5D5D5; }"
                              "QPushButton:checked { border: 1px solid #ADADAD; border-radius: 2px; background-color: #D5D5D5; }";
 
+
+    // color number need to be set to add new color to palette
+    // in cases where the last color was saved
+    // otherwise first color will be black.
+    editor()->color()->setColorNumber(0);
+    editor()->color()->setColor(editor()->color()->frontColor());
```
chchwy: These two lines should probably not be here. Usually in the initialization state, all the current states like the current frame, color, layer will be initialized in the editor & managers, UIs can only pull the states and updates itself. Not to change the states.

core_lib/src/graphics/vector/colourref.cpp
```diff
@@ -51,3 +53,12 @@ bool ColourRef::operator!=(ColourRef colourRef1)
         return false;
     }
 }
+
+QDebug operator<<(QDebug debug, const ColourRef& colourRef)
```
chchwy: Another minor thing, I guess the return type would be better to be `QDebug&` not `QDebug`

core_lib/src/structure/object.h
```diff
@@ -82,10 +82,14 @@ class Object : public QObject
     QString copyFileToDataFolder( QString strFilePath );
 
     // Color palette
-    ColourRef getColour( int i ) const;
+    ColourRef getColour( int index ) const;
+    void useAsTempPaletteColor(QColor color) { mFrontColor = ColourRef(color, "Front Color"); }
```
chchwy: Could you please explain to me why adding `useAsTempPaletteColor()`?

core_lib/src/structure/object.cpp
```diff
@@ -272,38 +274,90 @@ void Object::deleteLayer(Layer* layer)
...
+
+        msgBox.exec();
+        if (msgBox.clickedButton() == cancelButton)
+        {
+            return false;
+        }
+        else if (msgBox.clickedButton() == removeButton)
+        {
+            return true;
         }
```
chchwy: It's probably not a good idea to pop up a message box in the Object, Object should be pure logic object. It breaks the princile of separating UI & logic, and also makes writing unit test harder as message box will block the function call until someone hits the button.

I would recommend to have a `bool isColourInUse()` and a separate `void removeColor()`. And the moving the message box to ColorPaletteWidget.

chchwy:
Sorry for the late reply. I was sick and had two days in the bed.

app/src/colorpalettewidget.cpp
CandyFace: Ah right, a typo. It should indeed be angle brackets.

app/src/colorpalettewidget.cpp
CandyFace: Ah that's a remnant from earlier testing, will remove

core_lib/src/structure/object.h
CandyFace: Because since vector is dependant on mPalette, if there are no swatches left, the color will become black. This makes sure that it'll use the front color instead if that's the case.
Are you against the method, should I add it to Object::setColour instead?

The other solution is to not allow all swatches to be deleted but I find this to be a workaround rather than a proper solution.

app/src/colorpalettewidget.cpp
CandyFace: Hmm it was necessary in the beginning, because otherwise it wouldn't save the previous color, except when you clicked on one of the swatches in the palette. It doesn't seem to be the case anymore though... so I'm removing it

core_lib/src/structure/object.h
chchwy: Probably the best solution for now (I mean before you implement the connected color), we probably couldn't allow users to delete all palette colors, need to at least keep one color.

CandyFace: Okay so i'll remove the useTemp solution and make it impossible to delete the last remaining swatch.

chchwy: Thanks :)

CandyFace:
Changes implemented, you can now review again and hopefully merge :)

chchwy:
Thanks @CandyFace, well done.

PR #972, opened 2018-05-20
Rearrange default dock widget locations

scribblemaniac:
This is a small change that moves the color inspector to the right (under the color wheel) to match with the location of the sliders before they were made into a new widget. The display widget was moved to the left to balance the sides out. It should look like this now by default:
screen shot 2018-05-20 at 4 20 20 pm

I think this is a much better setup because it has all of the color stuff on the right and all the tools/settings on the left.


chchwy:
Thanks @scribblemaniac

Issue #973, opened 2018-05-21
[BUG] Selection transform proportional re-size not working with modifier [shift] key(s)


Jose-Moreno:
### --Issue Summary--
Hey so, yeah, basically the title. When showing the resizing function I tried to constraint it and it didn't work in the latest Nightly Build. I assume there has to be a regression implemented for it to work since it was working before (0.6.1 at least). Currently using the shift key only affects it's positioning to be constrained to the X or Y axis.

### --Expected Results--
Pressing Shift should allow for proportional resizing as well as constrained / discreet displacement and rotation (the latter has not been implemented, ever but maybe we can use this bug as an excuse :smile: )

### --Steps to reproduce--
1) Draw something
2) Select it with CTRL + A or the selection tool
3) Select the black arrow tool
4) Press the SHIFT key
5) Try to resize from a single corner and move it on different axis to notice that it fails to constraint the proportional edit.

### --System Information--
+ Pencil2D Version:
Version: 0.6.1.1
commit: 81f8da7d4105c579b480c5793abeba953311a3b2
date: 2018-05-18_22:58:16
Development build
Operating System: Windows 7 SP 1 (6.1)
CPU Architecture: x86_64

scribblemaniac:
This bug appears to have been introduced in ed3c4ec/#949.

@CandyFace can you please take a look at this when you get a chance?

Issue #881, opened 2018-02-02
[UX] Increase Selection Tool corner drag threshold


Jose-Moreno:
### --Issue Summary--
This is something really painful I've had to deal with ever since last year nightlies. It seems trying to resize any drawing via selection tool really takes time because the threshold is only taken into account from the inside of the selection, and not the outside, but this is not apparent to the user. Also if you click on top of the corners the selection will disappear, it requires ridiculous precision to resize something if you don't know the trick (clicking inside).

If possible could we increase the threshold or improve it so clicking & dragging the selection corner dots also allow us to resize the selection, and not only clicking the area inside the dotted line?

From what I've experimented the threshold is made of roughly an 8 pixel square area towards the inside of the selection radiating form the corner dot. We should increase this at least to 20px

### --Expected Results--
An Improved selection resize experience.

### --Steps to reproduce--
1) draw something
2) press CTRL + A or create a selection over it
3) Try to click and drag from the selection corners. Evaluate experience.
4) Try to click an drag next to the corner dot but inside the dotted line on it's proximity. Evaluate experience

### --System Information--
+ Pencil2D Version:
Version: 0.6.0 Development build

+ Operating System:
Windows 7

martinvanzijl:
How about this - I've extended the corner margins for clicking and dragging to about 20px, covering inside and outside the rectangle. See below (the green circles are just to show the margins; I'll remove them in the final PR.)

![corner-point-demo](https://user-images.githubusercontent.com/24422213/37419296-0b61d9e2-27ac-11e8-8bac-d272dced0948.gif)


How does that look?

Jose-Moreno:
@martinvanzijl I really like that! and actually I think the circles work better than the little dots as a complement since the user immediately can tell where they are supposed to click to drag the marquee box. Thank you for all of your invaluable submissions!

CandyFace:
Looks good @martinvanzijl 

I also agree with @Jose-Moreno that we could switch to ellipses instead of rectangles.

Taking it a bit further though, I've been meaning to change its style for some time, using the blue outline like we have on some of the icons to make the UI feel lighter.

![screen shot 2018-03-14 at 21 35 33](https://user-images.githubusercontent.com/1045397/37429465-a94b1eca-27cf-11e8-8b49-2f661fa8e61c.png)

The ellipse is filled with white color the outline Rgb(56,161,249)
I've also increased the thickness of the outline and changed the dashed line color.

[...]

PR #976, opened 2018-05-21
Selection move fixes

### LINKED ISSUE #974: Click+drag outside the selected area does not create new selection
### --Actual Results--
When clicking and dragging outside an existing selection with the selection tool, the current selection boundaries are modified.

### --Expected Results--
When clicking and dragging outside an existing selection with the selection tool, the current selection should be deselected and a new selection should be created in the same way as if there were no selection.

### --Steps to reproduce--
1. Choose the Select Tool
2. Create a selection by clicking and dragging
2. Try to create a new selection by clicking and dragging outside of the existing selection

### --System Information--
+ Pencil2D Version: Built from source >= ed3c4ec

+ Operating System:
macOS 10.13

+ RAM Size:
16 GB

---
### --For Code Contributors--
Issue begins with commit ed3c4ec so it is related to #949. Possibly also related to #973.

CandyFace:
This should fix #974, #973 and add the commit that @martinvanzijl made a PR for #881 (but was withdrawn again) some time ago

### DIFF
--- a/core_lib/src/tool/movetool.cpp
+++ b/core_lib/src/tool/movetool.cpp
@@ -167,10 +167,12 @@ void MoveTool::pressOperation(QMouseEvent* event, Layer* layer)
 
         if (mScribbleArea->somethingSelected) // there is an area selection
         {
-            if (event->modifiers() != Qt::ShiftModifier && !selectionRect.contains(getCurrentPoint()))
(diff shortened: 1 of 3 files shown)


chchwy:
Cool, thanks @CandyFace

PR #979, opened 2018-05-24
Optimize bucket fill with cache

scribblemaniac:
This is all from #837. The relevant section from there:

> Inspired by how Krita does flood filling, I have added a hash map to remember if colors are within the threshold for the duration of the operation (threshold and oldColor will remain constant, so only newColor is needed for lookup). Since the vast majority of the pixels that will be encountered are usually from a small subset of colors, this should have significant speed benefits. The Euclidean difference operation is probably a bit slower than the old difference operation, so this is even more important.

Hmm, I wonder if this will work: Closes #837.


chchwy:
Cool. Thanks @scribblemaniac .

PR #983, opened 2018-05-26
Improve error dialog when open file fails

scribblemaniac:
### Example
Old:
screen shot 2018-05-25 at 11 21 15 pm
New:
screen shot 2018-05-26 at 2 30 57 am

### DIFF
--- a/app/src/mainwindow2.cpp
+++ b/app/src/mainwindow2.cpp
@@ -431,9 +431,9 @@ void MainWindow2::tabletEvent(QTabletEvent* event)
     event->ignore();
 }
 
-void MainWindow2::newDocument()
+void MainWindow2::newDocument(bool force)
 {
-    if (maybeSave())
+    if (force || maybeSave())
     {
         Object* object = new Object();
         object->init();
@@ -471,7 +471,7 @@ void MainWindow2::openDocument()
         if (!ok)
         {
             QMessageBox::warning(this, tr("Warning"), tr("Pencil cannot read this file. If you want to import images, use the command import."));
-            newDocument();
+            newDocument(true);
         }
     }
     updateSaveState();
@@ -501,7 +501,7 @@ void MainWindow2::openFile(QString filename)
         if (!ok)
         {
             QMessageBox::warning(this, tr("Warning"), tr("Pencil cannot read this file. If you want to import images, use the command import."));
-            newDocument();
+            newDocument(true);
         }
     }
 }
--- a/app/src/mainwindow2.cpp
+++ b/app/src/mainwindow2.cpp
@@ -467,14 +467,8 @@ void MainWindow2::openDocument()
             return;
         }
 
-        bool ok = openObject(fileName);
-        if (!ok)
-        {
-            QMessageBox::warning(this, tr("Warning"), tr("Pencil cannot read this file. If you want to import images, use the command import."));
-            newDocument(true);
-        }
+        openObject(fileName, false);
     }
-    updateSaveState();
 }
 
 bool MainWindow2::saveAsNewDocument()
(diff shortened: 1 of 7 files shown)


app/src/mainwindow2.cpp
```diff
@@ -535,8 +563,27 @@ bool MainWindow2::openObject(QString strFilePath)
 
     Object* object = fm.load(strFilePath);
 
-    if (object == nullptr || !fm.error().ok())
+    if (!fm.error().ok())
```
CandyFace: Is there a reason you're not checking for object being null anymore?

CandyFace:
Although you should maybe also consider adding this to save document too since some of the corruptions we've seen has happened after the user saved their current project.

If for example the main xml file should somehow have been removed, then it shouldn't cancel saving and throw a similar error message.

app/src/mainwindow2.cpp
scribblemaniac: Stroll down another ten lines or so. I had to put it in a separate if statement because fm.error() won't have a proper title, description and details if fm.error().ok() is true.

CandyFace: Agh... my bad, totally missed that while reviewing 👍

scribblemaniac:
I agree, the save dialog is even more important and should receive the same treatment. I added more error reporting for saving (and opening) in my xmlstreams branch (and I think I have more changes stashed), but it has to be brought up to date with master and I still have to finish a couple sections. Modifying the save functions is obviously not something I want to rush because I might end up accidentally introducing new bugs, but hopefully I'll get around to finishing it before 0.6.2.

CandyFace:
Depending on when you intend or has time to finish it, it could be be a good idea to cherrypick those error reporting changes and include it either in this PR or in a new. I agree that you shouldn't rush to get things done.

Error reporting should be merged ASAP though since they're the only source for us to know if something goes wrong on the users end.

scribblemaniac:
It's not easy to cherrypick since a lot of the error reporting comes from the features of xmlstreams. I'll make it a priority on my Pencil2D todo list though.

chchwy:
Wouldn't it better to give only one single contact way? I reckon keep the forum and remove discord & GitHub. Developers are not always online on discord. Github seems too technical for ordinary users.

Also do you think the error message can be shorter? Users usually don't like a long error message (they probably even don't read them if it couldn't be understood at first glance).

scribblemaniac:
> Wouldn't it better to give only one single contact way? I reckon keep the forum and remove discord & GitHub. Developers are not always online on discord. Github seems too technical for ordinary users.

I included three for convenience. Github is far more convenient for developers and Discord may be more convenient for the user. Even if a developer is not online, usually everything gets responded to eventually. I'm not opposed removing them if you think that one is enough.

> Also do you think the error message can be shorter? Users usually don't like a long error message (they probably even don't read them if it couldn't be understood at first glance).

The image and the short bold heading clearly indicates at a glance that a) something went wrong and b) the file was not opened. If the user clicks away after seeing that, then that's probably all the information they care about. I have no sympathy for people who complain that their file is not opening if they don't find it worth the effort to read a couple of sentences that would probably solve their issue first.

I've written the descriptions with the intent of providing the user with an idea of what is wrong and how to fix it. The text area hides away the technical details that are useful for debugging so the user reads only the things they need to. While I fully welcome any changes that make it easier for the typical user to understand, I think this underlying design is the best way to address the issue, and I would be against removing information that could help users understand and solve the issue themselves.

Note that the dialog pictured in the first post is the longest message that I added because when that error occurs, there is no simple fix. Maybe the file uses a newer Pencil2D format. If not, then they either use a backup, or if the don't have one of those, then having a developer recover the files is really the only course of action. Take out the last option and you have people who can't do anything about their issue. Take out either of the first two options, and you have developers wasting time helping people recover things that don't need recovering.

chchwy:
Fair enough.

One last thing. Can you use variables or #define for the forum/github/discord links? It would be easier to change them in the future.

scribblemaniac:
> One last thing. Can you use variables or #define for the forum/github/discord links? It would be easier to change them in the future.

Good idea, will do.

Issue #950, opened 2018-04-20
[BUG] Pencil2D crashes when attempting to modify a non-existing first (key)frame


ghost:
### --Issue Summary--

Pencil2D crashes when attempting to modify a non-existing first (key)frame.

### --Actual Results--

Pencil2D crashes.

### --Expected Results--

Whatever is action is performed should be ignored.

### --Video or Image Reference--

![image](https://user-images.githubusercontent.com/11491024/39036787-e63d39aa-447e-11e8-8f9b-09136cb069fe.png)
Attempting to use a tool on the first frame like so will cause the program to crash

### --Steps to reproduce--

1. Create a new project. (This happens in existing projects too.)
2. Move the first frame of either the Bitmap layer or Vector layer to another position. (The Camera layer doesn't seem to have this problem).
3. Use a tool on the canvas. (Pencil, selection, etc.)

### --System Information--
+ Pencil2D Version:

Version: 0.6.1.1

+ Operating System:

Operating System: Windows 10 (10.0)
CPU Architecture: x86_64

+ RAM Size:

8GB

+ Graphics Tablet:

WACOM Intuos CTL-490

Jose-Moreno:
@CandyFace I have to add some details. It was reported today that exporting under these conditions also trigger a crash. So If you move the first frame forward and try to export a movie like that, it will trigger the crash as well.

CandyFace:
@Jose-Moreno Yup I can see that, looking into it currently.

CandyFace:
The mentioned crash when exporting has been fixed now too in #984

CandyFace:
Fix will be available in next nightly build, so i'm closing this.

ryyharris:
I found that when I 
1.  Create another keyframe in the vector layer
2. Delete the first keyframe
3. Try and draw when at the first keyframe
The program still crashes. It does not when I do the same steps with the bitmap layer

Jose-Moreno:
@ryyharris I can confirm your steps lead to an immediate crash. Thank you for reporting! I'll reopen this issue in lieu of your findings.

@CandyFace Since you fixed this last time, could you take a look at the same issue on the vector layer when you have a chance? Thanks

The bug happens on vector layers in the latest version.
Version: 0.6.3
Operating System: Windows 7 SP 1 (6.1)
CPU Architecture: x86_64

CandyFace:
Yea I'll look into it.

PR #984, opened 2018-05-26
Look for next frame if previous doesn't exist - #950

CandyFace:
When the mouse is pressed, we expect there to be a frame either on the exact pos or a previous frame, if that's not the case, the key return as nullptr and therefore crash the application.

This PR fixes that problem by looking for the next index if a previous doesn't exist.

### DIFF
--- a/core_lib/src/structure/object.cpp
+++ b/core_lib/src/structure/object.cpp
@@ -451,11 +451,26 @@ void Object::paintImage(QPainter& painter, int frameNumber,
             if (layer->type() == Layer::BITMAP)
             {
                 LayerBitmap* layerBitmap = (LayerBitmap*)layer;
-                layerBitmap->getLastBitmapImageAtFrame(frameNumber, 0)->paintImage(painter);
+
+                // Make sure there's a key before painting anything
+                // otherwise return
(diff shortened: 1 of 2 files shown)


CandyFace:
Don't merge, another bug related to this was found

core_lib/src/structure/object.cpp
```diff
@@ -451,11 +451,26 @@ void Object::paintImage(QPainter& painter, int frameNumber,
...
             {
                 LayerBitmap* layerBitmap = (LayerBitmap*)layer;
-                layerBitmap->getLastBitmapImageAtFrame(frameNumber, 0)->paintImage(painter);
+
+                // Make sure there's a key before painting anything
+                // otherwise return
+                if (!layer->keyExists(frameNumber)) {
+                    return;
+                }
+
```
chchwy: I'm afraid this condition check is not correct. If we are drawing frame 5 and couldn't find a keyframe at position 5, it should find a previous keyframe and draw it (for example, keyframe 3), not just skip it.

I will send a pull request.

CandyFace: Ah that's true 👍

PR #992, opened 2018-06-11
Lock zoom when using certain tools

scribblemaniac:
This prevents various strange errors that may arise from for example zooming out while drawing. Pointed out by checkers on Discord.


scribblemaniac:
> There is a mMouseInUse in ScribbleArea. Couldn't we use it?

Yes, we could. I looked for something like this, but I guess I looked in the wrong places. Updated, and am merging now, thanks!

PR #995, opened 2018-06-16
Gimp GPL palette support

CandyFace:
This PR makes it possible to both import and export gimp palettes.


CandyFace:
Bah... looks like I didn't have a clean history -.-

chchwy:
Good work!

by the way, sometimes `checkchanges.sh` fails on a merge commit like this PR:
https://travis-ci.org/pencil2d/pencil/jobs/393047674
Dunno why.

CandyFace:
Not sure either... anyway you might want to revert the other changes if possible. You should ideally have cherry picked Add support for Gimp Palette (GPL)

I tried fixing it but... gah I hate git when it fucks up.

Issue #951, opened 2018-04-20
[Request] Implement Range Export Feature for Image Sequences


Jose-Moreno:
### --Issue Summary--
Currently only the movie export has the "range" feature that allows you to export from and to a specific frame. It would make sense to have the ability to export a certain range of frames as an image sequence for people that can't seem to export the whole timeline as a video or as individual images.

CandyFace:
Implemented in PR #997

CandyFace:
Will be available in next nightly build.

PR #997, opened 2018-06-17
Add range to export image sequence

CandyFace:
As per request #951 you can now export an image sequence with a given range.


chchwy:
Cool, thanks @CandyFace

Issue #379, opened 2015-12-15
[REQUEST] The SELECTTool follows "rotation" of the canvas


carlosgf:
Tested with Nightly Build (14Dec2015)
Win8,1x64bits

The SELECT Tool is working fine. It just happened that when rotating the canvas, lets say 45º, the 
Select Tool also "rotates" with the canvas, so when you select a shape the select-tool-box will also be at 45º

... so it is a bit weird trying to select a shape which was previously rotated. You may try it by yourself, you will not be ready for that the first time 
:)
The REQUEST is for bringing the SelectTool to stay 0º rotated, independent of the rotation changes of the canvas

Jose-Moreno:
In this case I think we should Implement a Toggle to ignore canvas rotation: That is when utilizing freehand or shape selections by using the toggle you would be able to cut a shape manually without following the canvas rotation. However it should remain possible to toggle between both behaviors, that is following and not following the canvas rotation because sometimes you also select a region before rotating and expect the selection to follow the canvas so you can fine-tune certain details of the drawing (using the selection as a temporary mask)

PR #1005, opened 2018-06-25
Selection/move fixes and new stuff

CandyFace:
This PR takes the selection and move tool a notch closer to something ideal.

I've also refactored and cleaned up things to make it less error prone and verbose... like setSelection(QRectF rect) by removing "somethingSelected", because it really should be implicit  that you've selectedSomething when you set the selection (as the function implies)

New features:
+ UI/UX selection interaction icons
+ Selection ghosting ~~with snap to old position~~
selection 4

+ even more selection threshold
  + Goes both inside and outside
selection 2
+ UI/UX apply transformation (when modified)
  + On layer switch ~~and when leaving tool~~
selection 5

Fixes:
+ Non destructive transformation
+ Aspect ratio scaling only worked with the initial selection
+ Selection visual disappearing when intersecting itself
selection 3


Jose-Moreno:
@CandyFace So the non-destructive transformation would finally "fix" the incremental quality degradation bug that happens when you rotate? I mean Ideally it shouldn't degrade, but if that's impossible at least it should degrade until after the transform is applied, not after every rotation :slightly_smiling_face: 

Either way awesome contribution man! :smile:

CandyFace:
@Jose-Moreno  Indeed it will! and it will also allow moving the entire selection after having been rotated, even if it isn't visually inside the rectangle anymore. 
yes

Jose-Moreno:
@CandyFace Since it's related, what do you think of #379 ? I think there are times when you want the selection NOT to follow the view, and other times where you DO want it to follow the canvas. In Photoshop it works like it does in Pencil2D, that is the selection follows the view rotation hehe.

CandyFace:
@Jose-Moreno Ah yes that issue... Mhh I'm used to the selection tool not being affected by the view rotation so I'd say it's unnatural and a bug but if you say there's a reason why you might want such behaviour, then we should make it an option.

CandyFace:
Alright I'll look into those cases

I agree that maybe snapping during movement was not a desired feature after all, not when you can just undo or cancel the transformation.

About showing a warning when leaving the tool, I think it has a purpose for those who are not sure whether to they want to apply the transformation or not. It's true you could just undo but if you had been asked beforehand, then you wouldn't need to undo.

CandyFace:
Alright I've removed snapping and popup when leaving tool. 
Also fixed a bug when transforming after rotating...

chchwy:
Excellent job @CandyFace!
Unfotunately I'm going into work hell again and have no time to review this PR this week.

@scribblemaniac @Jose-Moreno please merge if you guys have reviewed it and think it's ok.

Kaikogf:
Nice improving here! :+1:

Issue #903, opened 2018-02-24
[ISSUE] Importing GIF Image Sequence will show empty frames


Kaikogf:
I imported 37 PNG image files from a simple animation into Pencil2D but all frames were empty/blank after imported.

The number of frames where imported correctly. Still those frames do not allow to draw on them, all strokes are hiding after releasing  the pen pressure...

Version Pencil2D build 23Feb2018
win 8.1 x 64bits

EDITED: 03April 2018 > PGN files are OK. GIF files shows this issue

Kaikogf:
@chchwy 
Not sure if I made a mistake at the time about this issue with PNG files, but today (build v. 02Apr2018) PNG are OK, only GIF files are the ones to show this issue.
I am going to edit the main title

Jose-Moreno:
@Kaikogf Can you please confirm if this still happens for you on the latest nightly build?

Kaikogf:
@Jose-Moreno 
Tested with version20Jun2018

It brings a new issue. Now pencil2D is unable to load any GIF image into the timeline. The following error appears:

![pencilgif](https://user-images.githubusercontent.com/28032931/42164202-3455331a-7dfd-11e8-9c86-3bc6b022e009.jpg)

It doesn't have to do with the number of GIF images. 
But tested the **_File > Import > Image_**, and it is working fine !
:+1:

Jose-Moreno:
@Kaikogf To bring animated GIF files you always had to use File > Import > Image as a single GIF file is still considered an individual image.

As for the error please be more specific, did it appear when you attempted to import image sequences with the GIF format?

Kaikogf:
@Jose-Moreno 
The error is when trying to import GIF image sequence. Even importing one single image with the import image sequence doesn't work

CandyFace:
I don't think we've ever allowed importing multiple gifs into a project. I found an old build from 2017 and it wasn't possible in that one either.

Not sure what the proper action is here... A gif is already a bunch of images, so the question is whether it makes sense to be able to import multiple.

Jose-Moreno:
@CandyFace I believe the issue is more about having both features working properly while trying to import any GIF file. From a personal perspective I feel there is a design flaw that we should fix while were at it. The still image importer should not be allower to import Animated GIF files (multiple frames) but rather the "image sequence importer" is the one that should be allowed to bring the Animated GIF files, that way we can use the frame spacing property.

[...]

PR #1008, opened 2018-07-03
Improve GIF UI/UX import/export

CandyFace:
Importing/exporting gif has been improved. #903

- Gif has been given its own option under file->Import and file->Export
- Spacing between frames now works for GIF

importexport

### DIFF
--- a/app/src/actioncommands.cpp
+++ b/app/src/actioncommands.cpp
@@ -133,9 +133,20 @@ Status ActionCommands::importSound()
     return st;
 }
 
-Status ActionCommands::exportMovie()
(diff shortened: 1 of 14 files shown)


Jose-Moreno:
@CandyFace wow that was super fast. Awesome! :smile: the only thing I would mention would be to add the (...) next to the "Animated GIF" name to keep the naming style consistent considering the importer / exporter takes the user to an option dialog.

CandyFace:
@Jose-Moreno 
Ah yes looks like I missed that... I've fixed it now 👍

PR #1010, opened 2018-07-05
Export movie/gif fixes

CandyFace:
* tweening didn't work for GIF exports ( hasn't worked since 0.6 apparently)
* fixed a bug that made the "open video" prompt not work when pressing yes. 
This was introduced via #1008 and discovered by @Jose-Moreno 
* Implemented "open file location" dialog for gif exports

### DIFF
--- a/app/src/actioncommands.cpp
+++ b/app/src/actioncommands.cpp
@@ -247,7 +247,7 @@ Status ActionCommands::exportMovie(bool isGif)
                                          tr("Finished. Open movie now?", "When movie export done."));
         if (btn == QMessageBox::Yes)
         {
-            QDesktopServices::openUrl(QUrl::fromLocalFile(strMoviePath).path());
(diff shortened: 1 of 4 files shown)


app/src/actioncommands.cpp
```diff
@@ -239,8 +239,19 @@ Status ActionCommands::exportMovie(bool isGif)
...
-            QMessageBox::information(mParent, "Pencil2D",
-                                             tr("Finished"));
+            auto btn = QMessageBox::question(mParent, "Pencil2D",
+                                             tr("Finished. Open file location?"));
+
+            QString name = dialog->getFileName() + ".gif";
+
+            // Chop name + extension off the string
+            // to get folder path
+            strMoviePath.chop(name.size());
```
chchwy: Using QFileInfo instead?

CandyFace: 👍

Issue #966, opened 2018-05-17
moving more than 35 frames corrupts frames upon save


CandyFace:
### --Issue Summary--
The issue https://discuss.pencil2d.org/t/frame-is-repositioned-randomly/2604/15
suggests that when moving large amount of frames, the frame content seems to corrupt. 

I can confirm that this is the case, although it's quite random what happens, I have experienced that some of the selected frames has been duplicated or removed entirely when saving.

I can't say where the issue stems from but it triggers once you save the project.

I've investigated the save behaviour and there's nothing to suggest in that process that something goes wrong, so my suspicion is that it happens when we offset the frames but somehow first trigger when we save the project (because that's when we save our modified keyframes to disk again)

### --Actual Results--
* Selecting 35 frames or more, offset them 30 to 50 frames
* play and make sure that everything seems to be fine
* save project and play again
As soon as you save and play the project again, you should be able to notice that some frames may randomly have been scrambled or has its content removed.

### --Expected Results--
The expected behaviour should be that one can offset x frames and always expect the same result once offset.

### --System Information--
+ Pencil2D Version: https://github.com/pencil2d/pencil/commit/336a789166d45ec701566a38bd8f4ec61652f4e1

+ Operating System:

CandyFace:
It's a pretty critical bug as it's not possible to undo the process atm, and will corrupt certain parts of your project if experienced, so I say we should prioritize it and maybe add it to 0.6.2 @chchwy what do you think?

chchwy:
Will look into this issue tonight.

Jose-Moreno:
@CandyFace Sorry it took me so long to test this. I downloaded the user test file, and I've attempted to move more than 35 frames, but I'm not managing to make it happen. I saved and tried to see if it scrambled, no luck. I closed and reopened the file but the frames were in the correct order. 

I'm using the latest nightly from June 13th commit: 999c05af9544706a9df72b39766704f30b378954 but maybe i'm doing something wrong.
How many frames do I have to slide into? Does it work if there's an empty space? I was ramming the selected frames into another chunk of frames and verifying if any of the selected frames or the other ones were getting scrambled before and after saving but things were consistent. Do you have any other kind of procedure or hint so I can reproduce this?

CandyFace:
@Jose-Moreno it may take more frames for it to happen. I can still reproduce this in latest commit to master.

Download the file from the forum, select from 241 to 358 and drag so the last frame (from left to right) it at 388

save the animation and press play.

[...]

PR #1012, opened 2018-07-07
Fix saving issue when moving frames

CandyFace:
This implementation contains the following
+ fixes that solves: #966 
+ A way to remove old data folder file references.
  + This happens currently before saving new files and after having saved new files, to avoid cases where a project contains old file references beforehand and when frames has been moved but the reference hasn't been deleted.

### DIFF
--- a/core_lib/src/structure/filemanager.cpp
+++ b/core_lib/src/structure/filemanager.cpp
@@ -292,6 +292,10 @@ Status FileManager::save(Object* object, QString sFileName)
                       tr("Cannot Create Data Directory"),
                       tr("\"%1\" is a file. Please delete the file and try again.").arg(dataInfo.absoluteFilePath()));
     }
+    // There may be old files lying around
+    // from projects before this implementation, cleanup their dirty workfolder
+    // before saving new content
+    removeFilesWithNoXMLReference(sMainXMLFile, sDataFolder);
 
     // save data
     int numLayers = object->getLayerCount();
@@ -355,6 +359,10 @@ Status FileManager::save(Object* object, QString sFileName)
     xmlDoc.save(out, indentSize);
     out.flush();
     file.close();
+
+    // There may be old files may be lying around, cleanup workfolder before
+    // zipping
+    removeFilesWithNoXMLReference(sMainXMLFile, sDataFolder);
 
     dd << "Done writing main xml file at" << sMainXMLFile;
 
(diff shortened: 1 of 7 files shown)


core_lib/src/structure/filemanager.cpp
```diff
@@ -292,6 +292,10 @@ Status FileManager::save(Object* object, QString sFileName)
                       tr("Cannot Create Data Directory"),
                       tr("\"%1\" is a file. Please delete the file and try again.").arg(dataInfo.absoluteFilePath()));
     }
+    // There may be old files lying around
+    // from projects before this implementation, cleanup their dirty workfolder
+    // before saving new content
+    removeFilesWithNoXMLReference(sMainXMLFile, sDataFolder);
```
chchwy: This is probably not necessary. Current file-saving process will only add used files into the zip archive, so it's ok to leave unused files in temporary working folder and ignore them.

If we really want to clean up the temp working folder,  a case which should be considered is:
1. Delete a sound clip
2. Save project
3. Undo <-- will this undo work?

CandyFace:
Don't merge, I found out that saving broke for new projects with no files attached... I'm looking into it atm.

core_lib/src/structure/filemanager.cpp
CandyFace: Hmm if that's the case, then i'll probably remove it again

CandyFace: Indeed it didn't seem to matter, oh well I've removed it again. You can review now :)

core_lib/src/util/pencilerror.h
```diff
@@ -71,6 +71,11 @@ class Status
 
         // Layer
         ERROR_NEED_AT_LEAST_ONE_CAMERA_LAYER,
+
+        // Timeline
+        KEYFRAME_MODIFIED,
+        KEYFRAME_MOVED,
+        KEYFRAME_EXISTS
```
chchwy: Seems these 3 values are not really used? In `saveKeyFrameFile()` it only checks whether it's `Status::OK`.
And Status is designed for error report, so if you like to return some specific values other than a true/false, probably making a plain enum is a better idea.

core_lib/src/structure/layerbitmap.cpp
```diff
@@ -57,13 +57,28 @@ Status LayerBitmap::saveKeyFrameFile(KeyFrame* keyframe, QString path)
...
 
-    BitmapImage* bitmapImage = static_cast<BitmapImage*>(keyframe);
-    Status st = bitmapImage->writeFile(strFilePath);
+    if (bitmapImage->fileName().isEmpty())
+    {
+        bitmapImage->setFileName(strFilePath);
+    }
+    else if(strFilePath != bitmapImage->fileName())
+    {
+        bitmapImage->setFileName(bitmapImage->fileName());
```
chchwy: Can you please explain a bit about this? why does the bitmap image assign the filename to itself?

chchwy: Ah. I think I understand why. You want to keep the previous file path

chchwy:
Thanks for taking the time to find the bug. The situation is more complicated than I thought.

If I got it right, the issue is the user moves a bitmap frame which has been unloaded (or even not loaded yet), so
1. `mImage` is empty and
2. the file path also changes

so Pencil2D completely loses the content of the bitmap frame and no way to get it back, right?

core_lib/src/structure/layerbitmap.cpp
CandyFace: Yes that's correct, because otherwise we'll load the wrong image path and that was essentially the bug.

core_lib/src/util/pencilerror.h
CandyFace: 👍  forgot to remove from an earlier test

CandyFace:
More or less yeah, before when a frame would be moved, in cases where a project would have old files  lying around, it could load from the wrong path or simply not load anything because the path couldn't be found. Now we make sure it uses the old path until we're saved the project, then we rename it.

PR #1028, opened 2018-07-27
Improve color palette selection behavior

### LINKED ISSUE #1024: Selected color in Color Palette
### --Issue Summary--
In Color Palette, selected color patch is overlapped with highlighting color causes not accurate of color shade

### --Actual Results--
p2d1
p2d2

### --Expected Results--
p2d2a

### --System Information--
Pencil2D Version:
Version: 0.6.1.1
commit: 4f73eb168e0539e02789f0fdf009f18588278a02
date: 2018-07-17_10:20:55
Development build

scribblemaniac:
This includes:
- Showing the proper icon color when selected (Fixes #1024)
- Adding a border around the selected color, this is particularly
useful for grid mode which has no other way to display the selection
- Refactored the resizing code so that going to grid mode from list
mode does not result in unexpected icon spacing

### DIFF
--- a/app/src/colorpalettewidget.cpp
+++ b/app/src/colorpalettewidget.cpp
@@ -185,6 +185,13 @@ void ColorPaletteWidget::refreshColorList()
     swatchPainter.drawTiledPixmap(0, 0, mIconSize.width(), mIconSize.height(), QPixmap(":/background/checkerboard.png"));
     swatchPainter.end();
     QPixmap colourSwatch;
+    QPen borderShadow(QColor(0, 0, 0, 200), 1, Qt::DotLine, Qt::FlatCap, Qt::MiterJoin);
+    QVector<qreal> dashPattern;
+    dashPattern << 4 << 4;
+    borderShadow.setDashPattern(dashPattern);
+    QPen borderHighlight(borderShadow);
+    borderHighlight.setColor(QColor(255, 255, 255, 200));
+    borderHighlight.setDashOffset(4);
 
     int colourCount = editor()->object()->getColourCount();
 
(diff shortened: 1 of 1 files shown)


app/src/colorpalettewidget.cpp
```diff
@@ -558,7 +571,25 @@ void ColorPaletteWidget::updateItemColor(int itemIndex, QColor newColor)
...
+
+    QIcon swatchIcon;
+    swatchIcon.addPixmap(colourSwatch, QIcon::Normal);
+
+    // Draw selection border
+    swatchPainter.setPen(borderHighlight);
+    swatchPainter.drawRect(0, 0, mIconSize.width() - 1, mIconSize.height() - 1);
+    swatchPainter.setPen(borderShadow);
+    swatchPainter.drawRect(0, 0, mIconSize.width() - 1, mIconSize.height() - 1);
+    swatchIcon.addPixmap(colourSwatch, QIcon::Selected);
```
CandyFace: Minor thing: selection border should imo. only be shown in grid mode. There's already normal cell selection highlight in list mode.

CandyFace:
Here's what it looks like for the curious:
for the curious

Good job @scribblemaniac 
I had started working on a similar solution but I found the icon selection states odd so I didn't quite get how to remove the highlight. I ended up making a custom item delegate but I see that's not necessary.

scribblemaniac:
> I ended up making a custom item delegate but I see that's not necessary.

I spent some time trying to remove the highlight with stylesheets, and when that didn't work I was just about to try a custom item delegate when I stumbled upon this solution.

app/src/colorpalettewidget.cpp
scribblemaniac: I agree that it isn't necessary in list mode, but I think it's still useful because the selection highlight differs based on the operating system. Here's a screenshot of the QListWidget from the qt docs:



Its style has much less emphasis on a selection than on a mac and a selection border may look better in styles like this.

Also I feel like it would be a bit of a pain to change because you would have to change the icons each time you toggle the mode, and either regenerate or store and retrieve the selection pixmap for each color when switching to grid mode. So part of the reason why I didn't do this is just laziness 😉

CandyFace: The list is updated on every click anyway so the required changes are minimal to hide the border. The style also seems pretty consistent across systems too, so aside from changing the Qt Theme from source code, the style should be somewhat consistent. 

The image you provided is not what it looks like in Pencil2D though, regardless the operation system as far as I know, this is how it looks through wine:
image

The only changes that are required is adding if statements around your selection border painting. ~290 and ~650

scribblemaniac: For some reason I was thinking that refreshColorList wasn't called on mode change, but I can see now that it is, and you're right it's very simple to change. I have pushed the changes needed to show the selection border in grid mode only.

chchwy:
Cool! it's a nice solution. Thank you @scribblemaniac

PR #1038, opened 2018-08-04
Fix #1002 the high memory usage when exporting a long project to a movie

### LINKED ISSUE #1002: [BUG] Excessive RAM memory usage during export and user case crashes
### --Issue Summary--
Various users have been complaining on recent issues with long-er animation projects that are failing to be rendered. 

I investigated and found that Pencil2D eats your RAM memory (at least on windows) when trying to render export video or image sequences for a specific file that had over 800 frames but with a resolution that was close to 4K.

When exporting I noticed that it was using all my RAM (8GB) just to try and export and it crashed. What I had to do was to export as Image Sequence, and even then It required a really long time and it used all my RAM (again) to export the file.

The problem is that I noticed later that the file didn't have a lot of movement, but rather that every drawing was keyframed even if there wasn't change.

Even though rendering heavy animation projects does indeed take time and computer resources, Pencil2D should have a better way to handle the RAM usage. I made a proposal here to consider a few concepts from other programs #965 

### --Expected Results--
Rendering videos or image sequences should not use all the available RAM while rendering, unless specified by the user.

### --Image Reference--
image
_This export test crashed silently so I couldn't get Windows crash report data even from the system log_

### --Steps to reproduce--
([Download the follwing project file to test this issue](https://drive.google.com/open?id=1uc71WKqdk4bYxkPGRal8xAuIENXa1VOL))

Movie Export Test
1) Open the provided file in Pencil2D
2) File > Export > Movie
3) Wait at least for the file to hit 10% of overall export progress and check your memory usage. In my case it was almost half my total RAM capacity.
4) If you want try to let it end and see if it crashes for you.

Image Sequence Export Test
1) Open the provided file in Pencil2D
2) File > Export > Image Sequence
3) Wait at least for the file to hit 10% of overall export progress and check your memory usage. In my case it was almost half my total RAM capacity.
4) If you want try to let it end and see if it crashes for you. I my case it didn't crash, and finished overnight, but it did use all of my computer resources.

### --System Information--
+ Pencil2D Version:
Version: 0.6.1.1
commit: 52db352d2bc25e650448f5b0e9c92aba14e11a64
date: 2018-06-16_11:45:26
Development build
CPU Architecture: x86_64

chchwy:
Fix #1002 

1321 frame project, 1 bitmap layer.

Before
 

After

### DIFF
--- a/core_lib/src/interface/editor.cpp
+++ b/core_lib/src/interface/editor.cpp
@@ -65,15 +65,12 @@ Editor::Editor(QObject* parent) : QObject(parent)
     clipboardBitmapOk = false;
     clipboardVectorOk = false;
     clipboardSoundClipOk = false;
-
-    mActiveFramePool.reset(new ActiveFramePool(200));
 }
 
 Editor::~Editor()
 {
     // a lot more probably needs to be cleaned here...
     clearUndoStack();
-    mActiveFramePool->clear();
 }
 
 bool Editor::init()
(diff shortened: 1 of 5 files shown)


chchwy:
Thanks for reviewing it @scribblemaniac

PR #1039, opened 2018-08-04
Add completion for bash and zsh

J5lx:
A minor thing that’s been on my personal todo list since I last worked on our command line interface.

Right now the bash completion is noticeably more primitive than the zsh completion (mostly because bash’s completion system as a whole is noticeably more primitive than zsh’s) however both should be quite useful.


chchwy:
Thank you @J5lx

PR #1040, opened 2018-08-04
Timeline didn't extend to a correct length when opening a super long project

chchwy:
The issue:
1. Start pencil2d, the timeline length is 240 frames by default.
2. Open a 1000 frames project
3. Timeline length extends to 360 frames, which should be 1000 frames.

In the previous implementation, timeline grows 1.5 times longer based on the current length. But in this case, the timeline length must grow more.

Also, I changed the timeline extending rule. I think the previous 75% rule is a little too much for a long project. I work on a 1000 frames and timeline extends to 1500 frames when I create a keyframe at frame 750. 

The new rule is, Timeline will extend 100 more frames when a user adds a keyframe within 50 frames to the end of Timeline.

### DIFF
--- a/core_lib/src/interface/timeline.cpp
+++ b/core_lib/src/interface/timeline.cpp
@@ -250,6 +250,10 @@ void TimeLine::extendLength(int frame)
     if(frame > (currentLength * 0.75))
     {
         int newLength = std::max(frame, currentLength) * 1.5;
+
+        if (newLength > 9999)
+            newLength = 9999;
+
         mTracks->setFrameLength(newLength);
         updateLength();
     }
--- a/core_lib/src/interface/timeline.cpp
+++ b/core_lib/src/interface/timeline.cpp
@@ -242,17 +242,14 @@ void TimeLine::setLength(int frame)
 
 /** Extends the tineline frame length if necessary
  *
- *  If the new animation length is more than 75% of the timeline
- *  frame length, then double the timeline frame length, otherwise
- *  do nothing.
- *
  *  @param[in] frame The new animation length
  */
 void TimeLine::extendLength(int frame)
 {
-    int frameLength = mTracks->getFrameLength();
-    if(frame > frameLength * 0.75) {
-        mTracks->setFrameLength(frameLength * 1.5);
+    int currentLength = mTracks->getFrameLength();
+    if(frame > currentLength - 50)
+    {
+        mTracks->setFrameLength(frame + 100);
         updateLength();
     }
 }
--- a/core_lib/src/interface/timeline.cpp
+++ b/core_lib/src/interface/timeline.cpp
@@ -240,16 +240,17 @@ void TimeLine::setLength(int frame)
     updateLength();
 }
 
-/** Extends the tineline frame length if necessary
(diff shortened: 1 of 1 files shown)


chchwy:
Ok, I've brought the exponential growing rule back, and using `std::max(currentLength, frame)` to expand to a new length.

CandyFace:
I'd propose that instead of using the length of the frame, we should use the maximum() and value() of qabstractslider to figure out how far away we are from the end of the scroll area.

```c++
    int scrollThreshold = 100;
    if((mHScrollbar->value() > mHScrollbar->maximum() - scrollThreshold)) {
        mTracks->setFrameLength(frameLength * 1.5);
        updateLength();
    }
```

Alternatively we could auto expand when scrolling x frames from the end, without depending on frames.

timeline

There's of course a problem using such approach... the higher you go, the harder is gets to control how far you want to jump. To avoid such problem I've been thinking we could implement drag and extend kind of scroll bar. You can then set min and max and choose exactly how large or thin you want it bar to be.
https://youtu.be/uUWzsbcFPUw?t=61
Audition has a very nice and intuitive scrollbar imo, where you can just adjust the ends, controlling both how far you can scroll and how the size of your timeline content, eg. frames.

Jose-Moreno:
@chchwy @CandyFace @scribblemaniac To be honest guys I think this is getting rather complicated for a functionality not many are even aware of (many people still ask how to extend the timeline). 

Most animation or motion graphics software do NOT auto-extend their timeline.  Normally they rather let you define a range of frames to work with because in animation you as the professional need to define the length of your shot in frames previous to animating. As we're considering about a future timeline re-write, having a definite timeline range (that you can increase or decrease at will) would probably prove more benefical as it would help to lower memory consumption coming from the GUI and that way we can consider implementing the "preview" functionality that has been talked about before, so we would "cache" the timeline range.

All the Adobe programs have this "timeline zoom slider" that lets you dinamically enlarge or shrink the size of the frames to accommodate more, but that's only when the timeline has been assigned that many frames. In After effects you can't have more than the max total amount of defined frames. If you need more you increase the maximum and that's it.

With that said, please let's not pay attention to random users who want to animate 10 thousand frames in a single file. That never happens in real life or in a professional environment pipeline as it's a technical liability because you RISK losing work over using a single file instead of splitting work between multiple workers to make it a more efficient production. The only program you should use to accommodate more than 5 minutes of footage is a video editing application and Pencil2D should not be used like that. I'd rather have the software optimized to define how many frames you want beforehand than trying to cope with the possibility of having more frames than one can manage. Let me know what you think about it :slightly_smiling_face: 

Other than that thank you all for the good work!

chchwy:
Thanks a lot @Jose-Moreno, it's really good to know how other software work.

For the current status of Pencil2D, Pencil2D doesn't have the sense of "project length". Even timeline is only 240 frame long, users can still add keyframes to the positions beyond frame 240. We didn't limit it, even users can't see them on timeline. However, we can discuss it if we want to introduce the "length" of a project to Pencil2D. 

I added a maximum length 9999 to timeline so it won't grow over it.

Any other questions about this PR, guys?

Issue #932, opened 2018-03-30
[ISSUE] When importing an animation (or image) twice it will upset Pencil2D .... :D


Kaikogf:
Pencil2D v. 28Mar2018
Win8.1 Pro 64bits 34GB Huion tablet

Steps:
Use the default user interface
1. Open Pencil2D
2. Go to File > Open and select an animation or single image and load it to Pencil2D.
3. Go again to File > Open and select the same file, and load it to Pencil2D
4. Two different results here:
a- All frames were loaded correctly but they are now empty
b- All painting tools shows a delay and when drawing (rather fast mode) lines will broke as a bezier's lines

5. When not closing Pencil2D and going File > New, the above behavior remains when you repeat the steps

chchwy:
@Kaikogf Can you explain more about the empty frames? The image importing is working on my side.
Also when you say "File > Open", does it mean actually open a `pclx` or File > Import > Image (Sequence) ?

CandyFace:
@chchwy you should be able to reproduce it if you open the same project again.

File->Open or open recent and select the same file.

Kaikogf:
@chchwy 
I just tested it using the following path: File > Open > selecting an animation > OK. Do it twice with the 
same file.
Yes, I mean a pclx file This is for the animations. 

And for the image. I made a mistake here. I thought I was actually opening a pgn/jpg file image but what I was doing was opening another pclx file with an image .... But anyway, importing an image (jpg, pgn) twice does not give any error 
:+1:

Jose-Moreno:
Can confirm this bug with the annotation made by CandyFace.

Jose-Moreno:
After reviewing other issues, I think this is also related to #752 , both must stem form the same problem.

chchwy:
Working on this.

Kaikogf:
Hi @chchwy 
My QT is not able to do teh code for the last build, so here is an error:

E:\PENCIL2D\pencil-master\pencil-master\core_lib\src\managers\colormanager.cpp:20: In file included from ..\..\pencil-master\core_lib\src\managers\colormanager.cpp:20:0: 

Tested an the issue is no longer here
:+1: 

@chchwy 
I have tried the Qt creator for building the code, but it can't. 
Do I have to update some files from Qt?

Kaikogf:
Just to let you know, pencil2D continues to lag a lot when playing an animation with 400mb , more than 350 frames.

Jose-Moreno:
[...]

Issue #752, opened 2017-09-08
Opening the same recent project, causes incomplete sound loading 


CandyFace:
### --Issue Summary--
If you open a project, import a few sounds files, save it and open it from recent again, then the sound files will not be loaded and inserting a new sound file will crash the program with:
```
[Object] couldn't copy sound file to data folder:  "/Users/CandyFace/Documents/Polycode/Asteroid/Sfx/pew.wav"
ASSERT: "!strCopyFile.isEmpty()" in file ../../pencil-mychanges/core_lib/managers/soundmanager.cpp, line 132
```

### --Actual Results
Recent file is loaded again but sound files are not imported correctly. Also, adding new sounds again will crash the program with the error above.

### --Expected Results--
Recent file is loaded again with sound files imported correctly

### --Video or Image Reference--
![recent-bug](https://user-images.githubusercontent.com/1045397/30211195-49f68e04-94a0-11e7-8905-f368dabe0be3.gif)

### --Steps to reproduce--
1. open a new project
2. import a few sound files
3. save it
4. open it again from recent opened files

### --System Information--
+ Pencil2D Version:
commit: b661883877266fa96080ff58359453d74812e464
date: 2017-08-26_16:01:16

+ Operating System:
Mac OS 10.12.6

Jose-Moreno:
[2018] This issue still happens in the latest Nightly Build for:

Version: 0.6.1.1
commit: 999c05af9544706a9df72b39766704f30b378954
date: 2018-06-13_15:28:41
Development build
Operating System: Windows 7 SP 1 (6.1)
CPU Architecture: x86_64

The current work around is to load a new file and the re-load the file with sound again and it will load properly.

chchwy:
Closed as it's fixed in #1043

CandyFace:
I can confirm it has been fixed on Mac os 👍

chchwy:
Thanks @CandyFace

PR #1043, opened 2018-08-07
Extract pclx to a random generated folder name to avoid name conflicts

chchwy:
Related issues #932 #752  

The issue described by @scribblemaniac :
If a user opens the same project twice in a row, Penci2D will:

1. Create data folder for file to load
2. Load everything into the data folder
3. Stop if there is an error
4. Switch to using the new data folder
5. Delete the old data folder

However, if the old data folder and new data folder are the same, then step 5 ends up deleting all the data from our project! 

In order to resolve the name conflicts, the data folder will include an 8 character long random string.

For example:
- YourProject_Y2xD_6xseyd51
- Default_Y2xD_5zy4hsui

### DIFF
--- a/core_lib/src/structure/object.cpp
+++ b/core_lib/src/structure/object.cpp
@@ -45,16 +45,11 @@ Object::~Object()
 {
     mActiveFramePool->clear();
 
-    while (!mLayers.empty())
-    {
-        delete mLayers.takeLast();
-    }
+    for (Layer* layer : mLayers)
+        delete layer;
+    mLayers.clear();
 
-    // Delete the working directory if this is not a "New" project.
-    if (!filePath().isEmpty())
-    {
-        deleteWorkingDir();
-    }
+    deleteWorkingDir();
 }
 
 void Object::init()
--- a/core_lib/src/structure/object.cpp
+++ b/core_lib/src/structure/object.cpp
@@ -164,12 +164,20 @@ void Object::createWorkingDir()
         QFileInfo fileInfo(mFilePath);
         strFolderName = fileInfo.completeBaseName();
     }
-    const QString strWorkingDir = 
-        QString("%1/Pencil2D/%2_%3_%4/").arg(QDir::tempPath()).arg(strFolderName).arg(PFF_TMP_DECOMPRESS_EXT).arg(uniqueString(8));
-
     QDir dir(QDir::tempPath());
-    dir.mkpath(strWorkingDir);
 
+    QString strWorkingDir;
+    do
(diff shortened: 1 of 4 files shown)


core_lib/src/structure/object.cpp
```diff
@@ -164,11 +164,8 @@ void Object::createWorkingDir()
         QFileInfo fileInfo(mFilePath);
         strFolderName = fileInfo.completeBaseName();
     }
-    const QString strWorkingDir = QDir::tempPath()
-        + "/Pencil2D/"
-        + strFolderName
-        + PFF_TMP_DECOMPRESS_EXT
-        + "/";
+    const QString strWorkingDir =
```
scribblemaniac: For my peace of mind, could we change the following few lines to this:
```
QString strWorkingDir;
QDir dir(QDir::tempPath());
do {
    strWorkingDir =
        QString("%1/Pencil2D/%2_%3_%4/").arg(QDir::tempPath()).arg(strFolderName).arg(PFF_TMP_DECOMPRESS_EXT).arg(uniqueString(8));
} while(dir.exists(strWorkingDir));
```
The chances of generating the same string is almost impossible, but I would feel better if we checked for it anyway.

scribblemaniac: The default folders are not getting deleted. This is because the `Object` destructor only calls `deleteWorkingDir` if `filePath()` is empty. Instead I think we should always call `deleteWorkingDir`, and just add a check in that function to make sure that `mWorkingDirPath` is not empty.

On a slightly unrelated note, is there any reason why we don't just call `init` in the `Object` constructor? That way we can be sure that it is called at least once.

J5lx:
Seems good to me except for what scribblemaniac noted. Also, not directly related to the PR, but what does Y2xC and Y2xD stand for? It seems the last letter distinguishes between (de)compression, but what about the rest? Just curious.

core_lib/src/structure/object.cpp
chchwy: The behaviours were introduced in a PR #591, I don't really remember why it's working in this way. So the working directory will be deleted anyway now.

For my understanding, Object::init() is for a new Object. In Pencil2D it's the starting default project and the project when you click from the menu File->New. It's no harm to call init() on an Object from a pclx file but all data will be overwritten immediately.

chchwy:
Y2xC is legacy and not used anymore. I delete it. Preserved Y2xD for the sake of debugging. I can quickly check all pencil2d working directories on my computer by a keyword.

PR #1051, opened 2018-08-26
Reimplement autocropping

scribblemaniac:
This is a second attempt at what I did in #791. Quoting from there:

> Now the BitmapImage will automatically resize itself to be at the minimum size possible without cropping any data. This fixes the second issue of #774 and goes a step further to update the Select All bounding box when the edge is erased with an eraser, deleted with a selection clear, or brush strokes/flood fill with an alpha of 0 are applied. Since it resizes the image when necessary, this will also reduce the file size and speed up playback slightly whenever part of the minimum bounding box of an image decreases (most notably when moving the whole layer).

This has been optimized to prevent this feature from slowing down Pencil2D. The image should not be cropped at all until a brush stroke is finished so there should be little to no extra brush lag. There is also a flag to prevent running the function when the boundaries could not possibly have decreased since last time the function was called.

There are some other partly related changes slipped in here:
- completely cleared bitmap frames are no longer saved as single-pixel images
- an unnecessary image initialization+pasting step was removed in the brush/pencil drawing process, so there may be some speed improvement when using those tools.

### DIFF
--- a/core_lib/src/graphics/bitmap/bitmapimage.cpp
+++ b/core_lib/src/graphics/bitmap/bitmapimage.cpp
@@ -133,14 +133,16 @@ BitmapImage BitmapImage::copy(QRect rectangle)
     return result;
 }
 
-void BitmapImage::paste(BitmapImage* bitmapImage)
-{
-    paste(bitmapImage, QPainter::CompositionMode_SourceOver);
-}
-
 void BitmapImage::paste(BitmapImage* bitmapImage, QPainter::CompositionMode cm)
 {
     QRect newBoundaries;
+
+    bitmapImage->autoCrop();
+    if(bitmapImage->width() <= 0 || bitmapImage->height() <= 0)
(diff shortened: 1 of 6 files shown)


chchwy:
Thanks @scribblemaniac. It's really good to have auto-crop.
The same concerns, the auto-crop seems to happen in the middle of a stroke drawing. Is there any way to avoid it? Maybe we can pick the timing of auto-crop rather than always do it on getting sizes and positions?

scribblemaniac:
@chchwy It doesn't seem to be calling autocrop at all for me until after stroke drawing is complete. Can you provide some way for me to replicate your findings? If it is a problem than we could put it on a timer or maybe only call it when select all or save is called.

chchwy:
Sorry for late reply. I double checked again and you are right, it only does auto-crop after stroke is complete. cool, thanks @scribblemaniac.

PR #1059, opened 2018-09-07
Improve tablet resolution?

scribblemaniac:
From [the Qt docs](https://doc.qt.io/qt-5//qtabletevent.html):
> Qt will first send a tablet event, then if it is not accepted by any widget, it will send a mouse event. This allows users of applications that are not designed for tablets to use a tablet like a mouse. However high-resolution drawing applications should handle the tablet events, because they can occur at a higher frequency, which is a benefit for smooth and accurate drawing. If the tablet events are rejected, the synthetic mouse events may be compressed for efficiency.

So I refactored the code to handle the tablet event instead of ignoring it and Qt possibly compressing it as mouse events.

Please test thoroughly as I don't have a tablet to test this on myself. Particularly on a Wacom as I did remove the "patch".  As far as I can tell for tablet events, `event->posF()` should equal `event->pos() + mTabletPosition - mTabletPosition.toPoint()`.

### DIFF
--- a/core_lib/src/interface/scribblearea.cpp
+++ b/core_lib/src/interface/scribblearea.cpp
@@ -424,7 +424,11 @@ void ScribbleArea::tabletEvent(QTabletEvent *event)
     else {
         editor()->tools()->tabletRestorePrevTool();
     }
-    event->ignore(); // indicates that the tablet event is not accepted yet, so that it is propagated as a mouse event)
+    // Tablet move events should not be ignored as that will propagate it as a mouse event with reduced resolution
+    if(event->type() != QEvent::TabletMove)
+    {
+        event->ignore(); // indicates that the tablet event is not accepted yet, and propagates it accordingly
+    }
 }
 
 bool ScribbleArea::isLayerPaintable() const
--- a/core_lib/src/tool/strokemanager.cpp
+++ b/core_lib/src/tool/strokemanager.cpp
@@ -40,6 +40,22 @@ StrokeManager::StrokeManager()
     connect(&timer, &QTimer::timeout, this, &StrokeManager::interpolatePollAndPaint);
 }
 
+void StrokeManager::genericMoveEvent(QPointF pos)
+{
+    // only applied to drawing tools.
+    if (mStabilizerLevel != -1)
+    {
+        smoothMousePos(pos);
+    }
+    else
+    {
+        // No smoothing
+        mLastPixel = mCurrentPixel;
+        mCurrentPixel = pos;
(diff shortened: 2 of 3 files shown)


chchwy:
Thanks @scribblemaniac will test it as long as I have the tablet with me.

CandyFace:
herp
This is what I'm experiencing now with my tablet..
small strokes are me tapping on the screen, long strokes are continous... something is definitely not right.

scribblemaniac:
> We don't do anything with tablet events... all our tools expect QMouseEvents which are called from scribblearea::mouse<press,release,move>event(). The corresponding tabletmoveevent does not do anything so of course tablet support is broken now that we're trying to use those events.

@CandyFace Thanks for testing. I don't know how I missed that. I guess I saw the stroke manager call in scribblearea and assumed all of them were being done from there, not just the first and last points.

CandyFace:
As discussed in Discord, I have now implemented the tablet overrides in the drawing tools. I also couldn't resist doing a bit of refactoring.

I have tested the implementation using my Artisul D13 tablet.

CandyFace:
Still need to implement for hand, move and select. Will do that tomorrow

CandyFace:
Alright this should be good to go now. I could still refactor a lot to get that dreadful copy paste away though but functionality wise, all tools should work now with proper tablet inputs.

CandyFace:
Agh... will fix when I get home. I haven't been able to test that because I don't have a middle mouse button but I'll figure something out.

scribblemaniac:
> Agh... will fix when I get home. I haven't been able to test that because I don't have a middle mouse button but I'll figure something out.

Okay, I figured it would be easier for me to fix it then. Ended up being a very simple fix. This should be good to go now as far as I can tell. 👍

chchwy:
Looking at this PR again, I have a question.
Is it necessary to introduce these 3 functions?

```
void BaseTool::tabletMoveEvent(QTabletEvent*);
void BaseTool::tabletPressEvent(QTabletEvent*);
void BaseTool::tabletReleaseEvent(QTabletEvent*);
```

Seems they are doing exactly what mouseEvent functions are doing, Is it correct? Or are there any differences for a specific tool?

CandyFace:
@chchwy You are correct, within the tools the functionality is very similar (and identical in some cases) to each other (especially after i've moved all the button related code out) and could be merged.  We could consider getting rid of QTabletEvent and QMouseEvent and make it possible to get them as value from Basetool or... create a new custom struct that holds the values we require from each EventType.

PR #1065, opened 2018-09-28
fix typo, s/offical/official/

mapreri:
This string is also in the translation files, but I assume you have some automated thing to update those.


chchwy:
Thanks, there is no automated thing but anyway :p.

PR #1087, opened 2018-10-03
Small changes that can prevent users from duplicate layer names

davidlamhauge:
It is a hidden bug in Pencil2D, that users can have duplicate layer names.
It is unlikely that anyone would name their layers with the same name, but it should not be possible at all. This PR would mend this bug - as I call it.


chchwy:
Thanks @davidlamhauge 
It's a simple solution to stop users from creating duplicated layer name.

We can implement the layer name increments later on.

Issue #1068, opened 2018-10-01
Add support for tif/tiff import and export


scribblemaniac:
We used to have support for TIFF import and export that implemented in the same way as png, bmp, and jpg, using Qt's built-in functionality. Unfortunately it was not working so it was removed in https://github.com/pencil2d/pencil/pull/948.

Unlike some of the more common formats, tiff is not in the core Qt gui, but is rather a plugin (maintained by Qt). I'm pretty sure the issue was that this plugin wasn't getting loaded properly. Solving this will hopeful be enough to fix the issue and reinclude this format. For more information about Qt image format plugins, see here: https://doc.qt.io/qt-5/qtimageformats-index.html

Jose-Moreno:
@scribblemaniac I have tested this in Version: 0.6.2
commit: ecbe8537aff9e70e99e16797383d21aedf5e39fe
date: 2018-10-05_13:12:56
Development build
Operating System: Windows 7 SP 1 (6.1)
CPU Architecture: x86_64

TIFF Import / Export works as expected. I recommend closing this issue if you deem it proper.
Thanks to @nevdokimof for the implementation and heeding the reviews :slightly_smiling_face: 

**_Note_**: I'll leave this comment as something I'd like to see later on regarding this format, and for which I'll make a separate issue. One of the qualities of TIFF images is that it can store layers. If possible it would be great that the TIFF export from Pencil2D would allow it to store at least bitmap layer information in the tiff files. It would be interesting as well to consider the vector layer information to be included, however these would have to be rasterized on export and converted into bitmap layers for that to work.

PR #1088, opened 2018-10-03
#1068 added support for tiff import/export

nevdokimof:
Had to change "while" to "if" on core_lib/src/interface/editor.cpp:754 because usage of 'while' relies on fact that QImageReader::read(QImage) doesn't reset position inside the image file which apparently ain't working with TIFF files ( QImageReader::read(QImage) always returns true for TIFF which results in infinite loop)

### DIFF
--- a/core_lib/src/interface/editor.cpp
+++ b/core_lib/src/interface/editor.cpp
@@ -751,7 +751,7 @@ bool Editor::importBitmapImage(QString filePath, int space)
         return false;
     }
 
-    while (reader.read(&img))
+    if (reader.read(&img))
     {
         if (!layer->keyExists(currentFrame()))
         {
--- a/core_lib/src/interface/editor.cpp
+++ b/core_lib/src/interface/editor.cpp
@@ -751,7 +751,7 @@ bool Editor::importBitmapImage(QString filePath, int space)
         return false;
     }
 
-    if (reader.read(&img))
+    while (reader.read(&img))
     {
         if (!layer->keyExists(currentFrame()))
         {
@@ -769,6 +769,12 @@ bool Editor::importBitmapImage(QString filePath, int space)
         }
 
         backup(tr("Import Image"));
+
+        // Workaround for tiff import getting stuck in this loop
+        if (!reader.supportsAnimation())
+        {
+            break;
+        }
     }
 
     return true;
--- a/core_lib/src/util/pencildef.h
+++ b/core_lib/src/util/pencildef.h
@@ -26,10 +26,10 @@ GNU General Public License for more details.
     QObject::tr( "AVI (*.avi);;MPEG(*.mpg);;MOV(*.mov);;MP4(*.mp4);;SWF(*.swf);;FLV(*.flv);;WMV(*.wmv)" )
 
 #define PENCIL_IMAGE_FILTER \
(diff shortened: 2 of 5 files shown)


core_lib/src/util/pencildef.h
```diff
@@ -26,10 +26,10 @@ GNU General Public License for more details.
     QObject::tr( "AVI (*.avi);;MPEG(*.mpg);;MOV(*.mov);;MP4(*.mp4);;SWF(*.swf);;FLV(*.flv);;WMV(*.wmv)" )
 
 #define PENCIL_IMAGE_FILTER \
-   QObject::tr( "Images (*.png *.jpg *.jpeg *.bmp);;PNG (*.png);;JPG(*.jpg *.jpeg);;BMP(*.bmp)" )
+   QObject::tr( "Images (*.png *.jpg *.jpeg *.bmp *.tif *.tiff);;PNG (*.png);;JPG(*.jpg *.jpeg);;BMP(*.bmp);; TIFF(*.tif *.tiff);;" )
 
 #define PENCIL_IMAGE_SEQ_FILTER \
-    QObject::tr( "Images (*.png *.jpg *.jpeg *.bmp);;PNG (*.png);;JPG(*.jpg *.jpeg);;BMP(*.bmp);" )
+    QObject::tr( "Images (*.png *.jpg *.jpeg *.bmp *.tif *.tiff);;PNG (*.png);;JPG(*.jpg *.jpeg);;BMP(*.bmp);; TIFF(*.tif *.tiff);;" )
```
scribblemaniac: In this line and the line above, the string should not end with `;;` because that causes there to be an blank entry at the end.

core_lib/src/interface/editor.cpp
```diff
@@ -751,7 +751,7 @@ bool Editor::importBitmapImage(QString filePath, int space)
         return false;
     }
 
-    while (reader.read(&img))
+    if (reader.read(&img))
```
scribblemaniac: I see why you made this change, it freezes for me as well and I don't know why. Unfortunately this is an unacceptable change. From the Qt docs:
> For image formats that support animation, calling read() repeatedly will return the next frame. When all frames have been read, a null image will be returned.

Which means that animated gif export which also uses this function will only import the first frame. My recommended workaround here is to change this back to a while loop and then add this to the end of the loop :
```
// Workaround for tiff import getting stuck in this loop
if (!reader.supportsAnimation())
{
    break;
}
```

scribblemaniac:
Closes #1068

core_lib/src/interface/editor.cpp
chchwy: I have a question. It says it will return a null image at the end of frames, not return a false.

scribblemaniac: Very strange, must be a typo. The return type is definitely bool.

chchwy: Probably it's not a typo, there are 2 two overloaded QImageReader::read()
```
QImage QImageReader::read()
bool QImageReader::read(QImage *image)
```
and they shared the same function descriptions. So I think we could try to check whether the QImage is null, not the bool value.

scribblemaniac: No that description was for `bool QImageReader::read(QImage *image)`, could be a copy and paste issue from the other method though. You could definitely try `while(!(img = reader.read()).isNull())`, but I'm not too optimistic that it will have any different results than the current approach.

chchwy: You're right. The result didn't change.

chchwy:
Thank you @scribblemaniac @nevdokimof

PR #1091, opened 2018-10-05
Filename suggestion (numbers) for new layers in Pencil2D

davidlamhauge:
A function added that suggest new filenames, that differ from existing ones.
If you want a new "Bitmap Layer" and a "Bitmap Layer" exists, it will suggest "Bitmap Layer 2" as filename, and so on.

### DIFF
--- a/app/src/actioncommands.cpp
+++ b/app/src/actioncommands.cpp
@@ -686,8 +686,7 @@ QString ActionCommands::nameSuggest(QString s)
     int j = 2;
     QString tmp = s;
     do {
-        tmp = s + QString::number(j);
-        j++;
+        tmp = s + " " + QString::number(j++);
     } while (sLayers.contains(tmp));
     return tmp;
 }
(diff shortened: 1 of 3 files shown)


app/src/actioncommands.cpp
```diff
@@ -684,25 +664,31 @@ Status ActionCommands::deleteCurrentLayer()
...
-{
-    QMessageBox msgBox;
-    msgBox.setText(tr("Filename must be unique and not empty"));
-    msgBox.exec();
+    // if 's' is not in list, then return 's'
+    if (!sLayers.contains(s))
+    {
+        return s;
+    }
+    int j = 2;
```
CandyFace: Not a strict requirement but I find it to be a bad habit and generally dislike seeing one char variable names. when possible please name variables something that makes sense to the context. j for example could be named increment and string could be named "name".

it's a small thing but it improves code readability a lot.

davidlamhauge: Whether it's a bad habit, I don't know, but I tend to use bool b, String s, and int i, j and k. 
Being aware that I work on a project with many contributors, I will of course try to apply to the coding style.

chchwy:
Thank you @davidlamhauge @CandyFace

Issue #1076, opened 2018-10-01
Implement user customization of Image Pool Cache


Jose-Moreno:
### Issue Summary

Currently there is a hard-coded value for the mMaxSize = 200 in the [ActiveFramePool Class header](https://github.com/pencil2d/pencil/blob/29051b48f109684a53fa4f8137d71b0c76ef9730/core_lib/src/activeframepool.h
) which affects the FRAME_POOL_SIZE when the object is created. This seems to be changed by the [core_lib project file](https://github.com/pencil2d/pencil/blob/3892671dcd7c43f159949e0fb99f8fe078a85feb/core_lib/core_lib.pro) when compiling a project between 32bits (200) and 64bits (400) but it pretty much remains fixed.

The proposal here is to allow for the user to have direct access, through an "advanced" preference setting to allow them to change the image pool size which affects playback speed. This could help lower-end computers to become a bit more performant when working and would help aniamtors with higher-end computers to preview more frames before the pool cache is emptied (thus making the preview slower)

This of course is partly a solution for improving preview as we would still to consider implementing a dedicated Preview functionality for partial frame ranges and the complete timeline.

0x72D0:
seems fun, I will try this one

Jose-Moreno:
@0x72D0 Alright thanks and welcome! If you need anything let us know :slightly_smiling_face:

Jose-Moreno:
**Closing**
----
This has been implemented in Version: 0.6.2
commit: 6471882892c0d34ac20b0a70b9591685c9f68d04
date: 2018-10-12_11:17:33
Development build
Operating System: Windows 7 SP 1 (6.1)
CPU Architecture: x86_64

Thank you Lucas @0x72D0 & @chchwy !

PR #1102, opened 2018-10-06
add a frame pool cache option

0x72D0:
PR for #1076


0x72D0:
yes from what I seen the frame pool cache is set at the beginning of the program and I don't really know how to change it immediately after changing it in the preference... That why I only put a little warning to restart the program each time the value is changed

chchwy:
Okay no worries. I will take care of the rest of the parts.

0x72D0:
ok, thanks!

chchwy:
Merged. Thank you @0x72D0

PR #1103, opened 2018-10-07
issue 1101: Range spinboxes set uneditable at startup

davidlamhauge:
I've added two lines of code that set the spinboxes unabled at startup.
I have tried to open Pencil2D from command line, with a pclx-file as parameter. This file was saved with range set to 1-16, and that range was still active after loading the scene, so it works whether you open Pencil2D with or without a scene.

### DIFF
--- a/core_lib/src/interface/timecontrols.cpp
+++ b/core_lib/src/interface/timecontrols.cpp
@@ -51,6 +51,7 @@ void TimeControls::initUI()
     mLoopStartSpinBox->setFixedHeight(24);
     mLoopStartSpinBox->setValue(settings.value("loopStart").toInt());
     mLoopStartSpinBox->setMinimum(1);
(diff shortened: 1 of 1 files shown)


davidlamhauge:
Hmmm.... I added two lines of code. Nothing dangerous.
@scribblemaniac , @CandyFace or others: Any ideas what could make it fail the checks?

CandyFace:
Weird that travis failed... did you by any chance do a rebase --force or something?

davidlamhauge:
> Weird that travis failed... did you by any chance do a rebase --force or something?

Not a rebase --force, but I did a 'git pull origin master' which is pretty much the same, as far as I can read. 
If that master I am pulling from is not accepted by the quality check, then... - well - i don't understand the mechanisms here.

chchwy:
Sometimes the `checkchanges.sh` fails for no reason. Don't worry about it.

Issue #1094, opened 2018-10-05
*.gpl file palette importer miss-handling whitespace and names


Jose-Moreno:
### Issue Summary
I've been testing the gpl import-export in the case of importing actual gpl's that were exported properly form Pencil2D it will work, however many of the GIMP palettes have no names for colors, so if you try to use them on import Pencil2D it will start thinking the last RGB value of a color is it's name, and it will change the whole palette hue, lightness and saturation. Not fun if we assume many people will attempt to use GIMP and or Krita palettes without names.

Pencil2D should look to identify four value lists as a full color, and that if the fourth value is an empty string it should be given an automatic name like Color1, and subsequently each "unnamed" color will be named like this afterwards.

### Expected Results
I expect color palettes are not changed when being imported into pencil2d. As a user, If I didn't named it, I don't care about the name.

### Video or Image Reference

_The GIMP 2.8 gpl format palette for testing import issue_
[Firecode.gpl.zip](https://github.com/pencil2d/pencil/files/2449273/Firecode.gpl.zip)
This palette is made of warm colors. When you import it you will get green colors and you'll see the color name is made of the last RGB value.

### Steps to reproduce

1. Open Pencil2D
2. File > Import > Palette > Change palette type to *.GPL
3. Import the test palette. Notice the hue, saturation & lightness shift.

### System Information
+ Pencil2D Version:

Version: 0.6.2
commit: ecbe8537aff9e70e99e16797383d21aedf5e39fe
date: 2018-10-05_13:12:56
Development build
Operating System: Windows 7 SP 1 (6.1)
CPU Architecture: x86_64

Jose-Moreno:
I can confirm this works now. Colors without names are imported with an automatically generated name + a suffix number appended. Closing :slightly_smiling_face:

PR #1104, opened 2018-10-07
#1094 fixed .gpl import

nevdokimof:


chchwy:
Thanks @nevdokimof @CandyFace

PR #1108, opened 2018-10-14
Improve .gpl import

scribblemaniac:
Modified the gpl importer to more closely match the behavior of GIMP as it was not working with some valid gpl files. Looked through the GIMP source and tested various edge cases. Also used the automatic naming function for any colors that do not have a name.


chchwy:
Cool @scribblemaniac

PR #1109, opened 2018-10-15
Function to import images by filename

davidlamhauge:
This function will make it possible to import images based on the images filenames. This is mainly a feature that is usefull for traditional animators like myself.
Filenames must follow this format: 'prefix' + 'four digits' + 'suffix'.
If you have a layer with the character 'Joe', you could have the following files: 'Joe0001.png', 'Joe0013.png', 'Joe0015.png', 'Joe0017.png' and 'Joe0019.png'. If you import these drawings with the new function 'importImageSequenceNumbered()', you wil get a new layer named 'Joe' (the prefix), and the drawings will be placed at frames 1, 13, 15, 17 and 19.
I like it. I hope you like it too.

### DIFF
--- a/app/src/actioncommands.cpp
+++ b/app/src/actioncommands.cpp
@@ -22,6 +22,7 @@ GNU General Public License for more details.
 #include <QApplication>
 #include <QDesktopServices>
 #include <QStandardPaths>
+#include <QFileDialog>
 
 #include "pencildef.h"
 #include "editor.h"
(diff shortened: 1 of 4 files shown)


davidlamhauge:
@Jose-Moreno Thanks!
I too see it as a very useful feature, that can be expanded in the way you propose. 
1) You could import all drawings from a spredsheet (csv) into a new scene in Pencil2D, so you had your layers and their drawn/scanned drawings in the right place in the timeline.
2) You could import one or more layers into an existing scene. This kind of reuse is widely used in the business, and with the possibility to copy, move etc. it would be easy to adjust the imported layer,so it matched the rest of the scene.

davidlamhauge:
I split the function in two, so the actual import is separated from the finding filename.
The new function ('addLayerByFilename(strFilePath)') can now be called from other functions, if necessary.

PR #1111, opened 2018-10-19
Fix rename optimization during save

scribblemaniac:
During saving, there was some code to check if a frame changed position but was not modified so that a simple file rename could be used instead of a full image render and write.

However, there were many bugs that prevented this from working properly. I hope I have fixed them all. The most major one was caused by changes to the image loading, which would cause all frames to be loaded into memory during save. This fix will reduce memory consumption and increase the save speed by fixing this.

Thanks @Jose-Moreno for his extensive memory testing which brought this issue to my attention.

I have tested these changes in multiple scenarios and it is behaving as it should, but if I did make a mistake this could reintroduce the file wiping issues in the worst case scenario. So please review and test these changes thoroughly before merging.

### DIFF
--- a/core_lib/src/structure/layerbitmap.cpp
+++ b/core_lib/src/structure/layerbitmap.cpp
@@ -54,8 +54,7 @@ void LayerBitmap::loadImageAtFrame(QString path, QPoint topLeft, int frameNumber
 
 Status LayerBitmap::saveKeyFrameFile(KeyFrame* keyframe, QString path)
 {
-    QString theFileName = fileName(keyframe);
-    QString strFilePath = QDir(path).filePath(theFileName);
+    QString strFilePath = filePath(keyframe, QDir(path));
 
     BitmapImage* bitmapImage = static_cast<BitmapImage*>(keyframe);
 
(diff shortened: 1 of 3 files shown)


Jose-Moreno:
@scribblemaniac Ok I tried to compile it after getting the three files you committed but it seems QT won't let me as I'm getting about 30 errors and each time I try to build it it will halt. So i'd have to test it when there's a nightly i'm afraid.

chchwy:
@Jose-Moreno 
Please download this if you want to test this PR, thanks.
https://www.dropbox.com/s/cubqwtku7q293t3/pencil2d_pr1111.zip?dl=0

core_lib/src/structure/layerbitmap.cpp
```diff
@@ -93,43 +92,56 @@ KeyFrame* LayerBitmap::createKeyFrame(int position, Object*)
...
-        if (b->fileName() != fileName(b))
-        {
-            QString tmpName = QString::asprintf("t_%03d.%03d.png", id(), b->pos());
+        // Move to temporary locations first to avoid overwritting anything we shouldn't be
+        // Ex: Frame A moves from 1 -> 2, Frame B moves from 2 -> 3. Make sure A does not overwrite B
+        QString tmpName = QString::asprintf("t_%03d.%03d.png", id(), b->pos());
+        QDir sA, sB;
+        if ((sA=QFileInfo(b->fileName()).dir()) != (sB=dataFolder)) {
+            // Copy instead of move if the data folder itself has changed
+            QFile::copy(b->fileName(), tmpName);
```
chchwy: Could you tell me in what scenario the data folder will change?

scribblemaniac: When you load a pclx file, and save it as a pcl file, the data folder will change from the temporary directory to the .data directory. @Jose-Moreno first noted problems arising from this scenario which appear to be fixed now.

Issue #1074, opened 2018-10-01
Migrate timeline "frame size" slider widget from preferences to timeline


Jose-Moreno:
### Issue Summary

Changing the timeline frame container width is currently done through the preferences with a slider. While this is interactive, it's a bit buried and the user has no idea it exists. It would also be more efficient if we managed to use the widget a slider to act sort of like a "timeline magnification slider" but directly on the timeline instead of staying in the preferences.

(Pending: Add mockup, immediate reference After Effects magnifying slider)

Along with this I'd like to ask that the maximum width of the frames would also be increased. Wider keyframe drawing containers would help accommodate future thumbnails as seen in the timeline re-write proposal.

Jose-Moreno:
@CandyFace Yes it is! I was just separating a lot of these piled up issues from the feature tracker and to avoid confusion I also labeled it with the "timeline" label to indicate that all of these are related. I forgot to reference the timeline re-write issue though :sweat_smile:
Aside form that maybe I should put them into a "timeline" milestone as well? :thinking: What do you think?

CandyFace:
I think it's nice to have to as a separate issue :) and adding it to a timeline milestone good too

Jose-Moreno:
Fixed by #1121 
We might have to revisit the code once the timeline is rewritten, and it can use some enhancements in the UI/UX front (e.g it's very hard to select super thin frames) but for now it works as intended. Thanks!

PR #1121, opened 2018-11-02
Issue 1074. Slider to zoom framesize at timeline. Option removed from prefs.

davidlamhauge:
A zoom slider to change frame size. 
Values from 4 to 40. (Was 4 to 20).
For some reason it rearranges layout? Besides from that it works fine.

### DIFF
--- a/core_lib/src/interface/timeline.cpp
+++ b/core_lib/src/interface/timeline.cpp
@@ -139,7 +139,7 @@ void TimeLine::initUI()
     zoomSlider->setRange(4, 40);
     zoomSlider->setFixedWidth(74);
     zoomSlider->setValue(mTracks->getFrameSize());
-    zoomSlider->setToolTip(tr("Choose Framesize"));
+    zoomSlider->setToolTip(tr("Adjust frame width"));
     zoomSlider->setOrientation(Qt::Horizontal);
 
     QLabel* onionLabel = new QLabel(tr("Onion skin:"));
--- a/core_lib/src/interface/timeline.cpp
+++ b/core_lib/src/interface/timeline.cpp
@@ -26,6 +26,7 @@ GNU General Public License for more details.
 #include <QMessageBox>
 #include <QLabel>
 #include <QWheelEvent>
+#include <QSlider>
 
 #include "layer.h"
 #include "editor.h"
@@ -131,6 +132,16 @@ void TimeLine::initUI()
     duplicateKeyButton->setToolTip(tr("Duplicate Frame"));
     duplicateKeyButton->setFixedSize(24, 24);
 
+    QLabel* zoomLabel = new QLabel(tr("Zoom:"));
+    zoomLabel->setIndent(5);
+
+    QSlider* zoomSlider = new QSlider(this);
+    zoomSlider->setRange(4, 40);
+    zoomSlider->setFixedWidth(74);
+    zoomSlider->setValue(mTracks->getFrameSize());
+    zoomSlider->setToolTip(tr("Choose Framesize"));
+    zoomSlider->setOrientation(Qt::Horizontal);
+
     QLabel* onionLabel = new QLabel(tr("Onion skin:"));
 
     QToolButton* onionTypeButton = new QToolButton(this);
@@ -142,6 +153,9 @@ void TimeLine::initUI()
     timelineButtons->addWidget(addKeyButton);
     timelineButtons->addWidget(removeKeyButton);
     timelineButtons->addWidget(duplicateKeyButton);
+    timelineButtons->addSeparator();
+    timelineButtons->addWidget(zoomLabel);
+    timelineButtons->addWidget(zoomSlider);
     timelineButtons->addSeparator();
     timelineButtons->addWidget(onionLabel);
     timelineButtons->addWidget(onionTypeButton);
@@ -198,6 +212,7 @@ void TimeLine::initUI()
     connect(addKeyButton, &QToolButton::clicked, this, &TimeLine::addKeyClick);
     connect(removeKeyButton, &QToolButton::clicked, this, &TimeLine::removeKeyClick);
     connect(duplicateKeyButton, &QToolButton::clicked, this, &TimeLine::duplicateKeyClick);
+    connect(zoomSlider, &QSlider::valueChanged, mTracks, &TimeLineCells::setFrameSize);
     connect(onionTypeButton, &QToolButton::clicked, this, &TimeLine::toogleAbsoluteOnionClick);
 
     connect(mTimeControls, &TimeControls::soundClick, this, &TimeLine::soundClick);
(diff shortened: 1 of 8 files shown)


CandyFace:
Aside from the fact that this offset the playback buttons from being centred in the timeline, I'm not sure we should add features to the current timeline, since it's going to be reworked. 

@scribblemaniac started a timeline rewrite quite a while ago, which radially changes the implementation of the timeline, so any new features we add here will have to be migrated. 

it follows the idea of my mockup though so that's nice, but as it stands currently this requires the timeline to have a an even wider width to allow to see all buttons, which I'm not sure we want right now. The proper way to fix this would be to rethink the layout of the timeline and make it fit on less space but I personally think that should be done in the rewrite rather than the current timeline.

it looks good though @davidlamhauge but I wouldn't want to see it merged just to see it possibly go away in the rewrite.

Jose-Moreno:
@CandyFace @scribblemaniac I'm a bit worried about the following though. We spoke about it over discord too, and even tried to give a general consensus . I simply don't think postponing useful features in lieu of the "we're going to rewrite it anyway" excuse we've been giving to ourselves for the past years is the way to go anymore. I'll be frank. I'm not seeing that timeline re-write happening in at least 2 years time from now and that is a generous estimate. 

It's not because there isn't progress already, but because there's simply too much to do before getting it "right". However If we inhibit the program from growing visibly for our users, they simply won't be interested in coming back to it, and we'll lose traction. Yes, it might be a waste of time to improve the current timeline in the long run, but I'd rather have what we can implemented to be used now, so we can get more people using the software, spreading good will, and then once we have all the foundation elements ready target a full re-write of that feature along any other topping that might have been added to it. Let me know what you think.

CandyFace:
@Jose-Moreno I understand what you mean, I'm not sure it's an excuse though since we all want it to happen. The more we built on top of the current timeline, the harder and more time consuming it'll become to make anything that can replace the current one. 

I should probably mention that I don't work on the ethic that the application might lose user traction, so we should just hack away to please users. That's what i'm doing at the job, because I'm paid to do so and is constrained by deadlines. When I work on pencil however I believe that I can take my time to make things properly, even if it means that users will have to wait even longer. In that process I hope to make less buggy and hacky code too that can be easier maintained in the future.

Working on a foundation and later do a re-write might sound ok in theory, but through my experience, in practice that doesn't work in software development because too many things might get coupled together by that time, thus making it even harder to take the stubs apart to rewrite. There are simply too many wires to untangle at that point that you'll be spending more time reading code, trying to understand how to decouple x and y and fix unwanted bugs than writing new code. 

When that's said...
There's been no progress on the re-write in a very long time so maybe it is time to simply say it's been inactive for too long to consider waiting anymore and therefore let people build on top of the old.

davidlamhauge:
I can't decide whether it's going to be implemented or not. I just saw it as an issue (#1074) raised less than 5 weeks ago, so I thought I'd implement it, since no one else had done it through Hacktoberfest.
If you don't want it, then don't merge it. It's 1-2 hour of work. No big deal.

Jose-Moreno:
@davidlamhauge Sorry, didn't mean to spark controversy here. Neither of us certainly want to waste precious contributors work. Thank you for your understanding. Leave the PR up while we deal with this conundrum first and thanks again for your contribution.

Jose-Moreno:
@CandyFace I agree. I do not want to make an unmanageable mess that we have to untangle later. I also agree that we shouldn't exactly work only to "please" users, but at the same time I get inclined to disagree with that statement as much as I agree with it, mainly because we are invariably making a software that other people will use; non-technical users at that. Of course this thought is a balancing act as much as everything in software development; not an easy thing to manage, but we'll get there :yum:

I'm trying to abstract and analyze all of the information we have about this, but however I look at it we need to get to a definitive decision on this topic (the timeline re-write) before moving on, it's simply such an important part of the program that we kind of have to schedule it "soon". Also apologies for my poor choice of words, perhaps "excuse" wasn't what I meant to say but it's more that It feels like this is such an impassable obstacle of a dependency for the growth of the software that it needs to be addressed. 

Of course I don't mean to make this that big of a deal. We simply have to clarify how much of a priority the timeline should be. If possible I'd like us to discuss this with @chchwy and the others that have been notified about it. Of course everyone should work in whatever feature or fix they want, but if those contributions become blocked by this or any other dependency, we need to remove it from the way. Let me know what do you, or anyone else that reads this, think of how we should discuss this issue in the near future to consider getting at least a milestone roadmap for it's future implementation considering we have mockups and a possible (?) prototype code-base. 

As always thank you all for your input :slightly_smiling_face:

core_lib/src/interface/timeline.cpp
```diff
@@ -131,6 +132,16 @@ void TimeLine::initUI()
...
     duplicateKeyButton->setFixedSize(24, 24);
 
+    QLabel* zoomLabel = new QLabel(tr("Zoom:"));
+    zoomLabel->setIndent(5);
+
+    QSlider* zoomSlider = new QSlider(this);
+    zoomSlider->setRange(4, 40);
+    zoomSlider->setFixedWidth(74);
+    zoomSlider->setValue(mTracks->getFrameSize());
+    zoomSlider->setToolTip(tr("Choose Framesize"));
```
CandyFace: A minor detail but...
```suggestion
    zoomSlider->setToolTip(tr("Adjust frame width"));
```

Jose-Moreno:
@davidlamhauge Thanks for the vote of confidence :wink: As for "discussion" it means exactly this. Since each of us has different perspectives on how an issue should be solved, or how a feature should behave, often times we get to different conclusions as to how a functionality should work, therefore when a topic is contested as such we use the "discussion" label to signal that the issue has conceptual or technical inaccuracies to sort, and we do so either on the report or PR's related to it (like we are doing right now) we also try to discuss outside github as you've seen but it's good to leave the discussion history here so future contributors can understand our choices.

core_lib/src/interface/timelinecells.cpp
```diff
@@ -95,6 +96,14 @@ int TimeLineCells::getFrameX(int frameNumber)
     return x;
 }
 
+void TimeLineCells::setFrameSize(int size)
+{
+    mFrameSize = size;
+    QSettings settings (PENCIL2D, PENCIL2D);
+    settings.setValue(SETTING_FRAME_SIZE, mFrameSize);
```
scribblemaniac: Can you use the PreferenceManager instead here? ie:
```
mPrefs->set(SETTING::FRAME_SIZE, mFrameSize);
```

core_lib/src/interface/timelinecells.cpp
```diff
@@ -49,6 +49,7 @@ TimeLineCells::TimeLineCells(TimeLine* parent, Editor* editor, TIMELINE_CELL_TYP
     setAttribute(Qt::WA_OpaquePaintEvent, false);
 
     connect(mPrefs, &PreferenceManager::optionChanged, this, &TimeLineCells::loadSetting);
+
```
scribblemaniac: It doesn't really matter, but if you are making the changes anyway could you delete this trailing line.

Issue #978, opened 2018-05-23
[Request] Implement Flipping / Rolling Feature


Jose-Moreno:
### --Issue Summary--
This is one of the features that are on the request roadmap and it's been requested a few more times. 

In traditional animation Flipping & Rolling are two similar techniques that allow the animator to quickly preview their current animation work by flipping ( jump between the first, last and inbetween drawing of an action) or rolling (preview drawings in ordered sequence flipping with all your available fingers). 

**Explanation of Flipping Techniques**

1) The **"Inbetween" Flipping**: This technique is used mostly for the process known as in-betweening and the minimum amount of drawings you can use is two, but 99% of the time you will use three drawings. These are usually, the first drawing in the action (previous key), the last drawing (next key) and the currently edited inbetween.

REF: https://www.youtube.com/watch?v=0hw5eQF99ng

2) The **rolling** or **"roll flipping"** technique is used to check how a specific part of the action is working out, and the minimum amount of drawings are at most 5, one for each finger to manipulate plus one sitting at the bottom of the stack.

REF: https://www.youtube.com/watch?v=0C8arzQWUoI

3) There's also the **"flip"** or **"stack flipping"** technique which is when you take a large amount of drawings, organize them like a giant flipbook and let it "flip" down to check a large portion of your animation at a relative non-real time speed (it's hard to get constant fps using your fingers lol). This could already be considered as partially covered by the playback behavior though.

REF: https://www.youtube.com/watch?v=6Lmwz7M7SsY

In order to mimic this behavior in a digital environment timeline this functionality should aim to modify the playback behavior as it should be based upon the current drawing / frame container that is being edited, that is, where the playback scrub is located at the moment of activating the function. 

I have to note this functionality is semantically different from the Range function as well how its applied during a job. Whereas the current Range function allows you to preview a portion of the timeline from start to end including every frame in the range (in a way this would fulfill the stack flipping technique), the inbetween flipping and roll flipping functions have to behave differently to allow the animator to see the only the drawing(s) (normally keys or main drawings) behind and /or in front but always coming back to the currently edited drawing.

In the sense of previewing only the keyframe drawings, we have a similar behavior exhibited by the "onion skin toggle match keyframes" command (the little diamond button on the timeline)

For reference you can read about TVPaint's "Flip" function panel: https://www.tvpaint.com/doc/tvp11/index.php?id=lesson-tradigital-animation-advanced-flip-presentation

[...]

PR #1130, opened 2018-11-23
Issue 978: Fliip roll and Flip InBetween feature

davidlamhauge:
This PR addresses issue #978 - the flip-requests.
Flip roll is triggered by Alt + Z, and flip InBetween is triggered by Alt + X. These shortcuts are not yet implemented in the preference manager.
When the flips are triggered, the keyframes are played at an even tempo of 200 msecs/frame. This number is hardcoded, but should be set by the user to a value between 100 and 500 msecs.
The number of drawings in the flip roll is hardcoded to up to ten. Here the user should be able to set a maximum number of frames as well - I think...
By accident, this branch is not checked out from 'master' as I usually do, but from a branch called 'master_resolvewarnings' that I have been working on. For that reason the PR also includes a lot of resolved warnings...

### DIFF
--- a/app/src/actioncommands.cpp
+++ b/app/src/actioncommands.cpp
@@ -309,7 +309,7 @@ Status ActionCommands::exportImageSequence()
     int endFrame  = dialog->getEndFrame();
 
     QString sCameraLayerName = dialog->getCameraLayerName();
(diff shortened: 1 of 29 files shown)


davidlamhauge:
I have made a new PR with almost none warning resolves, so I close this one

PR #1131, opened 2018-11-24
Iss978 flip new

davidlamhauge:
This PR addresses issue #978 - the flip-requests.
Flip roll is triggered by Alt + Z, and flip InBetween is triggered by Alt + X. These shortcuts are not yet implemented in the preference manager. Can also be accessed from menu.

When the flips are triggered, the keyframes are played at an even tempo of 200 msecs/frame. This number is hardcoded, but should be set by the user to a value between 100 and 500 msecs.

The number of drawings in the flip roll is hardcoded to up to ten. Here the user should be able to set a maximum number of frames as well - I think... My question is - where should we put the editables for the flips? In the preference dialog or in the Onion skin dockable, or a third place?

There are some warning resolves from editor.cpp. They are pretty obvious and don't include calculations, so I hope they'll pass.


davidlamhauge:
I have fixed the issues mentioned.
In the preferences, the user can set how many frames there should be included in the Flip Roll, and the user can also set the speed the flips are running at.
There were merge conflicts with the Master branch that are solved, and then I had a shortcut bug that I don't have energy to explain, but it is also solved.
The flip feature works fine, and I look forward to see it merged...

davidlamhauge:
Thanks!

core_lib/src/managers/playbackmanager.h
```diff
@@ -85,13 +88,18 @@ class PlaybackManager : public BaseManager
...
+    int mFlipRollMax;
 
     QTimer* mTimer = nullptr;
+    QTimer* mFlipTimer = nullptr;
     QElapsedTimer* mElapsedTimer = nullptr;
     int mPlayingFrameCounter = 0; // how many frames has passed after pressing play
 
     bool mCheckForSoundsHalfway = false;
     QList<int> mListOfActiveSoundFrames;
+    QStringList mFlipList;
```
chchwy: Is there any reason to use a QStringList rather than a QList<int>?

chchwy:
Merged. Thanks for your contribution @davidlamhauge 
Please check this PR in the 5 Feb 2019 nightly build.

Issue #1133, opened 2018-11-26
Huion Linear Optimizer Issues


scribblemaniac:
Many users have reported issues with Huion tablets. This issue will serve as a tracker for multiple different bugs from these complaints that I believe are all caused by the "Linear Optimizer" option in the HuionTablet app (possibly labeled "Enable Wintab Linear Optimization" on Windows). These issue include #877, #1110, and [Difficulty with Tablet](https://discuss.pencil2d.org/t/difficulty-with-tablet/1369). I am able to reproduce these issues on my Huion H430P only when the Linear Optimizer option is enabled. Unless otherwise noted, these are all applicable to v0.6.2 and the November 9th 2018 nightly build. I have only been able to confirm that these issues occur on mac because the Windows application for my tablet does not have an option to enable the linear optimizer, and linux doesn't offer any options at all to my knowledge.

**If you are experiencing similar problems** to what is described below try disabling the option in the HuionTablet app under the Press Keys section:
![Linear Optimizer option location](https://discuss.pencil2d.org/uploads/default/original/1X/71662f50ab2dda12a0e16ffbaff50a0a48e78056.png)
Not all Huion drivers expose this option unfortunately (including the H420), but hopeful we can find a fix for this.

### Issue 1 – Extreme Lag
Extreme lag occurs when hovering over the tablet and when attempting to draw a stroke. Here is a [video](https://drive.google.com/file/d/12aePv30ygNChFSJKpKHvWabTSGR7GFv6/view?usp=sharing) demonstrating the issue. This issue is not a fixed delay, but rather a delay that increase the more you use the tablet, likely because Pencil2D is not processing events as fast as they are coming in, resulting in a buildup of events.

I built a program to measure the frequency of tablet events (I will open source it soon if you are interested in trying it yourself). It shows that my tablet as a peak rps/pps (reports per second/points per second) of 256. With the linear optimizer enabled, this doubles to 512. Subjective testing on Pencil2D suggests that this is an underestimate, possibly because of tracking events, which I have not yet been able to get working on my application.

Only counting unique tablet events (defined as events with a different event->posF() than the previous event) shows the same 256 rps with and without the linear optimizer. Unfortunately ignoring these events in scribblearea does not solve the problem.

Possible solutions for this include making the tablet move event processing faster and/or skipping events when the application is receiving more than it can process.

### Issue 2 – Stroke Continues After Release
[...]

PR #1134, opened 2018-12-04
Remove Show Tab Bar from the View menu on mac

scribblemaniac:
Newer versions of macOS add a "Show Tab Bar" option to the "View" menu automatically. Pressing it screws with the layout a little bit like this:
double_titlebar
There is no way to remove the menu item with Qt, see [this bug report](https://bugreports.qt.io/browse/QTBUG-61707). To fix this I integrated a small amount of native objective-c code for mac builds.

This also enables us to easily add support for more native mac features in the future. I implemented mouse coalescing in this same way, but is currently not being used. It seems like what we are currently doing for setting mouse coalescing is not working, so when we actually disable it the brush stroke lags significantly similar to #1133. According to my event counter the peak number of events doubles from 128 to 256. This shouldn't cause a noticeable slowdown because we can process that many tablet events just fine. I've spent a few hours trying to debug this but am unsure exactly what the underlying problem is. I think it goes deeper than were just processing the events too slowly though. For now, we can just leave mouse coalescing enabled.

### DIFF
--- a/app/app.pro
+++ b/app/app.pro
@@ -30,7 +30,8 @@ INCLUDEPATH += \
     ../core_lib/src/tool \
     ../core_lib/src/util \
     ../core_lib/ui \
-    ../core_lib/src/managers
+    ../core_lib/src/managers \
+    ../core_lib/src/external
 
 HEADERS += \
     src/mainwindow2.h \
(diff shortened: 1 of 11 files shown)


chchwy:
Thanks, @scribblemaniac

PR #1135, opened 2018-12-04
Error counting digits fixed

davidlamhauge:
Too bad I first discovered this bug now, but better now than after the release.


chchwy:
Fix of #1109

PR #1136, opened 2018-12-05
Export only keyframes possible in exportImageSequence

davidlamhauge:
The possibility to import by filename has been merged.
The feature here will make it possible to only export the keyframes from a layer. Thus you can easily reuse animation from one scene to another.
There are also a few warning resolves. Only 'NULL' -> 'nullptr'


davidlamhauge:
Merge conflict resolved, and this time the Travis went well too...

chchwy:
Thanks @davidlamhauge

PR #1137, opened 2018-12-11
Improve darkmode QDockWidget titlebar

CandyFace:
This improves the look of QDockWidget titlebar on mac OS with Mojave Dark mode, making it look much closer to native...
Before:
image2

After:
image

Having looked at Qt bug report, Darkmode appearance still seems to be WIP, so there's bound to be bugs, especially since it's still a very recent addition...

I've already reported one: https://bugreports.qt.io/browse/QTBUG-72486
which explains why the qlabels in the timeline are dark when they should be light.


core_lib/src/interface/basedockwidget.cpp
```diff
@@ -17,10 +17,22 @@ GNU General Public License for more details.
...
+#include "platformhandler.h"
 
 BaseDockWidget::BaseDockWidget(QWidget* pParent) 
 : QDockWidget(pParent, Qt::Tool)
 {
+
+#ifdef __APPLE__
+    if (PlatformHandler::isDarkMode())
+    {
+        setStyleSheet("QDockWidget::title { background-image: url(:icons/darkmode-gradient.png); "
```
scribblemaniac: Could we use QLinearGradient instead here? Ex: `background: QLinearGradient( x1: 0, y1: 0, x2: 0, y2: 1, stop: 0 #4e4f4d, stop: 1 #424241); }`

core_lib/src/external/platformhandler.h
```diff
@@ -5,6 +5,7 @@ namespace PlatformHandler
 {
 
 void configurePlatformSpecificSettings();
+bool isDarkMode();
```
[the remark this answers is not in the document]
CandyFace: Both Appveyor and Travis successfully compiled on Window and Linux, so I don't think so?

core_lib/src/interface/basedockwidget.cpp
CandyFace: done.

chchwy:
Thanks @CandyFace

PR #1141, opened 2018-12-14
Resolved warnings for NULL and old-casts

davidlamhauge:
This PR resolved many of the warnings in the Interface folder, in core_lib.
Some warnings remain. In scribblearea.cpp alone there are 59 warnings concerning 'implicit conversions'. Someone should look at this one day. I'm not sure it's within my reach to decide whether Penci2D should use floats, doubles, qreals or what...


chchwy:
Thanks @davidlamhauge 
In most cases, float is the go-to type unless higher precession is required.

Issue #981, opened 2018-05-24
Implement canvas properties "preset" manager


Jose-Moreno:
### --Issue Summary--
A few people, and another one today (on discord) have asked for the possibility to change the default camera resolution that comes "hard coded" with every new file in Pencil2D. 

I'm not talking about making a "new file" dialog, but rather to expose the camera layer default size and any other feasible constants, as a preference option and take it from there.

Normally this implementation would ideally come with a way to have "presets" for TV, Web and other platforms which would become an ordered list of different configurations that can modify and replace the "default" editor settings for Pencil2D on the current and each new file. 

For reference in other animation software when you create a new file you can:

1. Setup the measurement Units (P2D works with pixels but in the future we might want to support other measurements due to the relationship between field size / inches for animation paper and camera work in traditional animation so at least the imported paper can have their "real" size recognized by the software.
2. Setup the canvas size (Width x Height)
3. Setup the duration of the shot (timeline frame length)
4. Setup the FPS
5. Setup the Bakground Solid Color / Texture / Transparency Channel
6. Setup pixel ratio / aspect ratio (in different media and video specifications the pixel changes it's ratio, i.e 1.0 are called Square Pixels, because they have a 1:1 ratio, same with the screen, 1.33:1 is the pixel size for the famous 4:3 aspect ratio, 1.375:1 is the aspect ratio for film from the MPAA, and so on. This is also a key aspect you need to be compliant of when submitting video to film festivals or government institutions, they ask for this stuff literally)

In Pencil2D most of these elements are already present in the preferences option dialog except No.1 (uninplemented), No. 2, No. 4, and No.6 (unimplemented)

Most of these properties also already modify the software globally and are remembered when opening the software or creating a new file (timeline length, background color / texture) Ideally all of these options should be modifiable.

I know we don't want to over complicate things, but still a preset manager for projects would be nice. 

For the preset idea we could use a dropdown box that is populate with the list of presets along an <add> and <remove> button to add and remove custom presets. Also when creating a new preset we would need to enter the required values into a form or options dialog similar to the one for the video exporter.

When selected these presets would immediately change the current file values, and would maintain the same values for any new files that are created after that choice.

This is a quality of life (QoL) request, since I acknowledge that changing these parameters doesn't take long, but if you're working with many files (shots / cuts) with the same parameters that differ form P2D's defaults, it becomes a chore. Let me know what you think about this :slightly_smiling_face:

J5lx:
[...]

PR #1143, opened 2018-12-15
Fps and field size in settings

davidlamhauge:
This is a feature I've missed since day one with Pencil2D.
Three crucial values have been hardcoded into Pencil2D, and the is FPS which is 12, and field size which are 800x600 pixels.
For me who always work with 24 fps and 1280x720, it's been annoying to set these values every time I open a scene in P2D, unless its a saved file.
12 fps and 800x600 are still default values, but now every new file will open with the last values you used.
I hope this PR will be accepted.


core_lib/src/managers/playbackmanager.h
```diff
@@ -84,7 +84,7 @@ class PlaybackManager : public BaseManager
     int mMarkOutFrame = 10;
     int mActiveSoundFrame = 0;
 
-    int mFps = 12;
+    int mFps;
```
CandyFace: Why remove the default value? this could lead to undefined behaviour in worst case scenario.

core_lib/src/structure/objectdata.h
```diff
@@ -65,7 +67,7 @@ class ObjectData
     QTransform mCurrentView;
     
     // playback manager
-    int  mFps    = 12;
+    int  mFps;
```
CandyFace: same comment, this should have a value to avoid undefined behaviours in case the fps won't be set.

core_lib/src/structure/layercamera.cpp
```diff
@@ -77,7 +78,15 @@ void CameraPropertiesDialog::setHeight(int height)
 LayerCamera::LayerCamera( Object* object ) : Layer( object, Layer::CAMERA )
 {
     setName(tr("Camera Layer"));
-    viewRect = QRect(QPoint(-400, -300), QSize(800, 600));
+    QSettings settings (PENCIL2D, PENCIL2D);
+    int fieldW = settings.value("FieldW").toInt();
+    int fieldH = settings.value("FieldH").toInt();
+    if (fieldW < 2)
```
CandyFace: You should check against both values.
eg...
```suggestion
    if (fieldW < 2 || fieldH < 2)
```

chchwy:
Thank you @davidlamhauge I think generally this PR good.
It could be even better to have these settings in the preference window, but it's already a very useful feature for now.

Jose-Moreno:
@davidlamhauge There was a discussion about the latter on this issue here #981 I personally think it would be great to have a preset manager for such things and now that you mention it would work wonders to have specific layer types and their specified quantity loaded as well, nice thought! :smile:

PR #1144, opened 2018-12-15
Renable dark mode on macOS

scribblemaniac:
Compiles and runs just fine on macOS 10.13, still needs to be tested on macOS 10.14 with and without dark mode. Based on code from Chromium.

### DIFF
--- a/app/data/Info.plist
+++ b/app/data/Info.plist
@@ -45,8 +45,6 @@
 			</array>
 		</dict>
 	</array>
-        <key>NSRequiresAquaSystemAppearance</key>
-        <string>True</string>
 	<key>CFBundleExecutable</key>
 	<string>${EXECUTABLE_NAME}</string>
 	<key>CFBundleIconFile</key>
--- a/core_lib/src/external/macosx/macosxnative.mm
+++ b/core_lib/src/external/macosx/macosxnative.mm
@@ -6,6 +6,10 @@
 
 namespace MacOSXNative
 {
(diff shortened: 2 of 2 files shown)


CandyFace:
it works on Mac OS 10.14 but what about the build server which runs on  10.12?
```
../../core_lib/src/external/macosx/macosxnative.mm:38:46: warning: instance method '-bestMatchFromAppearancesWithNames:' not found (return type defaults to 'id') [-Wobjc-method-access]

                [[NSApp effectiveAppearance] bestMatchFromAppearancesWithNames:@[

```

although that's just a warning, I think we should have it verified on a machine that actually runs 10.12 to make sure it works.

chchwy:
Can't we just wrap all the dark mode code with the 10.14 #define? The dark mode is only available for Mojave. All other mac os versions before Mojave don't even need it.

CandyFace:
The define is not enough by itself, since it already exists in 10.13 SDK. It would have to be something like the define this commit adds.
I'm not sure it works as intended though, the MAC_OS_X_VERSION_MIN_REQUIRED flag is ignored unless a target or min mac os version is set.

chchwy:
Ok, I see. We need a runtime mac os version check, not compile time check.

scribblemaniac:
Do we have any way to actually test this on 10.12? Because that is still something we should do.

chchwy:
I can test it. I have a macbook which runs 10.12

PR #1145, opened 2018-12-21
Zoom shortcuts for 25% up to 400%

davidlamhauge:
This PR introduces zoom shortcuts for 25%, 33%, 50%, 200%, 300% and 400%. The shortcuts are set to Shift+4, Shift+3, Shift+2, 2, 3 and 4. In the menu it is below 'Zoom in' and 'Zoom out' in a submenu, so the menu doesn't look cluttered.
It can be argued that the mouse wheel can zoom fast, but not all tablet pens have wheels, and sometimes you want to go from 300% to 50% fast.
I think it will be a popular addition.


davidlamhauge:
Added '1' as a shortcut for 100%. I thought about it yesterday, but today I decided it was needed. 
The Ctrl+H shortcut resets to 100%, but it resets the rotation too, and if you draw at 300% it can be helpful to look at it at 100% and quickly go back to 300%, without resetting the rotation.

Issue #1154, opened 2019-01-06
[ISSUE] The Drawing tools left round marks on most strokes


Kaikogf:
### Issue Summary

While laying strokes using different drawing tools, most of the strokes have a dot at the end of the stroke

 
#

+ Pencil2D Version:
20/12/2018

 

+ Operating System:
Win 8.1 x 64 bits

 

+ RAM Size:
32GB

 

+ Graphics Tablet:
HUION HD185

![defeito pencil2d](https://user-images.githubusercontent.com/28032931/50739451-72e92180-11d8-11e9-812d-5e93d9ec44ec.gif)

CandyFace:
Related to https://github.com/pencil2d/pencil/issues/677

It's not a consistent bug though, I have tried to replicate using 3 different tablet brands without experiencing this problem on mac.

scribblemaniac:
@Kaikogf What is your operating system?

davidlamhauge:
XP-PEN 22E pro on MacBook and HUION 580 on Ubuntu 18.04 tested: No problems... Can't replicate the error.

chchwy:
I can replicate this issue. Win 10 + Wacom Bamboo CTH670

Jose-Moreno:
@davidlamhauge @chchwy @Kaikogf We were on discord with Candyface and Scribblemaniac a few hours ago testing 667 which is related to this one. Scribblemaniac managed to implement a few fixes while I kept testing compiled versions with 5.12. It seems 5.12 didn't fix out issue, but the changes made to scribblearea.cpp did you can read the history in the #general-development channel hopefully a PR will be made soon so KaikoGF can test as well with a nightly build to close both issues soon enough.

Kaikogf:
> @Kaikogf What is your operating system?

Win 8.1 x 64 bits

Kaikogf:
I managed to build the last code [07/01/2019] in order to test this issue, and it is still present

Jose-Moreno:
@Kaikogf The fix we were testing before is still not merged, so wait a bit until it is reviewed :slightly_smiling_face:

Jose-Moreno:
@Kaikogf Can you please test nightly build for JAN 30th 2019 and confirm if this has been fixed for you? Here's a direct link to the build.
**Win 64**: https://drive.google.com/open?id=12ekVl_44GTCGJLh_3zcyjLhlwa6jAoXV

Kaikogf:
@Jose-Moreno 
[...]

PR #1155, opened 2019-01-06
Fix pressure artifacts at end of brush stroke

### LINKED ISSUE #677: [BUG] stabilization function create artifacts on drawing strokes

scribblemaniac:
Both tablet and mouse events were getting called during a brush stroke. Don't ask me why, because I don't know. Qt 5.12 did not fix this issue. Setting `Qt::AA_SynthesizeMouseForUnhandledTabletEvents` to false also had not effect. Since the tablet press occurs consistently before the mouse press, we can work around this issue by disabling part of the mouse events when the tablet is in the middle of a brush stroke.

Fixes #677. Possibly relates to #1154.

Note: If this code is used for ScribbleArea::mouseMoveEvent, there is a significant degradation of stroke quality on computers affected by this bug:
```c++
void ScribbleArea::mouseMoveEvent(QMouseEvent* event)
{
    updateCanvasCursor();

    if (!areLayersSane())
    {
        return;
    }

    Q_EMIT refreshPreview();

    // Workaround for tablet issue (#677 part 2)
    if(!mMouseInUse) return;

    mStrokeManager->mouseMoveEvent(event);
    mCurrentPixel = currentTool()->getCurrentPixel();
    mCurrentPoint = currentTool()->getCurrentPoint();

    // the user is also pressing the mouse (= dragging)
    if (event->buttons() & Qt::LeftButton || event->buttons() & Qt::RightButton)
    {
        mOffset = getCurrentOffset();
        // --- use SHIFT + drag to resize WIDTH / use CTRL + drag to resize FEATHER ---
        if (currentTool()->isAdjusting)
        {
            currentTool()->adjustCursor(mOffset.x(), event->modifiers()); //updates cursors given org width or feather and x
            return;
        }
    }

    if (event->buttons() == Qt::RightButton)
    {
        getTool(HAND)->mouseMoveEvent(event);
        return;
    }

    if (currentTool()->isDrawingTool())
    {
        if (event->buttons() & Qt::LeftButton)
        {
            currentTool()->mouseMoveEvent(event);
        }
    }
    else
    {
        // MOVE, SELECT OR HAND TOOL
        currentTool()->mouseMoveEvent(event);
    }

#ifdef DEBUG_FPS
    if (mMouseInUse)
    {
        clock_t curTime = clock();
        mDebugTimeQue.push_back(curTime);

        while (mDebugTimeQue.size() > 30)
        {
            mDebugTimeQue.pop_front();
        }

        if (mDebugTimeQue.size() > 30)
        {
            clock_t interval = mDebugTimeQue.back() - mDebugTimeQue.front();
            double fps = mDebugTimeQue.size() / ((double)interval) * CLOCKS_PER_SEC;
            qDebug() << fps;
        }
    }
#endif
}
```
This makes me think that the mouse move events being received when the tablet is in use are actually unique events and not duplicate/synthesized events. Note that with the the code in this PR, the mouse events are getting added to the stroke manager, but not to the tool. Not sure if this is the best way to do it. I think blocking mouseReleaseEvent is the critical part of this fix.


chchwy:
@scribblemaniac Could you please check one thing?

I remember that Qt will resend a mouse event if no one accepts the tablet event (calling event->accept()). 
I just did a quick glance at the tablet event handling and didn't see any tablet event accepted.

scribblemaniac:
> I remember that Qt will resend a mouse event if no one accepts the tablet event (calling event->accept()).

This was one of my first thoughts, however when I looked into it, `event->accept()` is getting called at the bottom of `ScribbleArea::tabletEvent`, and I see absolutely no way that that function is returning before accepting the event short of an exception. Not to mention that we should be able to replicate this on more systems if that was the issue.

chchwy:
I didn't see the `event->accept()`, can you tell me where it is?

chchwy:
Oh sorry never mind. My bad.

CandyFace:
before patch
image

after patch
image

There's virtually no difference between the strokes imho. Through my experimentations, it all comes down to the size of the drawings, ie. a bottleneck of the canvas not updating fast enough, thus halting the tablet events.

scribblemaniac:
@CandyFace Thanks for testing, I'm glad to hear everything is working.

I don't know if you stayed to the end of @Jose-Moreno 's livestream for this issue but we made some discoveries then. There seemed to be a noticeable degradation in quality when compared to the master branch (even when he compiled that himself rather than using the nightly build). However this issue was fixed after I changed ScribbleArea::mouseMoveEvent from the code above to the code in this PR. I can only think that extra events for mStrokeManager helped somehow. I personally don't experience any difference with my tablet for either version of the code, so it's probably specific to users who were experiencing the original issue.

CandyFace:
Hmm yes it definitely looks like it is affected in mousemove event... that's so strange. I did a quick measurement of the events before and after the change in mousemove but I see no difference in amount of samples between tablet and mouse.

I guess we'll have to dig deeper to find the reason why this happens.

chchwy:
Tested.
Found another issue, bucket tool doesn't work with tablet as it works with mouse.
Well done @scribblemaniac

Issue #1165, opened 2019-01-23
Specific tools working with mouse but not with tablet


Jose-Moreno:
### Issue Summary

Alright as discussed i'm raising a single issue since all of these seem to be related to #1164 and the fix should be related as well.

So after testing the latest source code (JAN 22nd 2019) and compiled with Qt 5.12 (the version has no effect in this as previously thought because I compiled it with 5.11.3) I've gathered a list of the tools that are working with a mouse input but not with tablet input:
- [x] Eyedropper
- [x] Polyline
- [x] Smudge / blur tool (test it by pressing ALT)
- [x] Selection tool**

** This one is the most critical in terms of destroying the artwork. In the case of the selection tool the clicks are acknowledged and it creates a selection box, but it also creates a second duplicate selection, so even though you can still create the selection, it is behaving strangely and the results are quite worrying as it completely destroys the drawing, however it works as expected when using the mouse.

### Expected Results

Each tool should work in the same way when using both a stylus and a mouse.

### Video or Image Reference + Steps to reproduce
The video below will provide a clear understanding of the current issues:

https://www.youtube.com/watch?v=pME-hLPT1l8

### System Information

Version: 0.6.2
Development build
Operating System: Windows 7 SP 1 (6.1)
CPU Architecture: x86_64

+ Graphics Tablet:
Genius MousePen 8x6

chchwy:
Please test with 30 Jan nightly build

Jose-Moreno:
@chchwy Okay tested. Everything is working as expected except for the following:

**1. Eyedropper**, even though it's working now with both mouse and tablet input, the preview functionality has changed. In the previous nightlies if you use the eyedropper tool or invoke the tool via pressing the ALT modifier, you automatically get the preview square without clicking. Now you only get it when you click and drag over the paint. This is just a minor difference, and while I think it's more logical to invoke the "preview picker" only when you are actively searching for the color (e.g click & drag) we have to document this behavior change in the release notes and future documentation.

**2. Selection Tool** The selection is now mostly working with both mouse and tablet input. However when moving the selected drawing with my tablet I am having a very low reproduction behavior (2 out of 50, and counting, attempts) that triggers a behavior which seems to be related to **Issue C** from #1053 which I added recently, Since I don't have conclusive evidence that it's related to this issue or tablet events I'll mark it as solved. Thank you everyone! :smile: :tada:

Issue #1162, opened 2019-01-22
Polyline tool graphical and behavior glitches


Jose-Moreno:
### Issue Summary

Hi. Basically the polyline tool is have a set of unwanted behaviors that were not found in the Nightly build from october 12th and 14th from 2018 but are found otherwise in all the builds starting from **pencil2d-win64-2018-11-01** up to the most recent one for jan 18th 2019.

The issues are explained in the video attached below

### Expected Results

Polyline should work in modal mode as shown in october 2018 nightly builds, where the line is always shown clearly after placing a new point and clicking and dragging should not update the previously placed point.

### Video or Image Reference + Steps to Reproduce

https://youtu.be/f7swdYSmlgw

### System Information
**Working correctly**

Version: 0.6.2
commit: 6471882892c0d34ac20b0a70b9591685c9f68d04
date: 2018-10-12_11:17:33

**Exhibiting unwanted behavior**

**From**
Version: 0.6.2
commit: b12c7109eedf4519af6bc036ee8fb2f45a3e1a67
date: 2018-11-02_08:24:27

**To**
Version: 0.6.2
commit: 90d1ce51d59b109af0d956d09e0c400f388df312
date: 2019-01-18_10:46:16

Development build
Operating System: Windows 7 SP 1 (6.1)
CPU Architecture: x86_64

Jose-Moreno:
Closing this issue. It has been fixed in:
Version: 0.6.2
commit: 832d3db6d5fdfb117dc9a7b4945592387c489059
date: 2019-01-31_10:24:15
Development build
Operating System: Windows 7 SP 1 (6.1)
CPU Architecture: x86_64

PR #1168, opened 2019-01-30
Fix tablet events and cleanup events handling

CandyFace:
We now have a specific PointerEvent class that should take care of all the mouse and tablet events and get rid of a so much duplicate code.

The PR however also fixes:
#1165
#1162

Additional fixes:
* Eyedropper icon resetting when the release event is done
* Select tool leaves a small selection piece when trying to tab to deselect.
 - This is WIP solution...

I've tested it with my mouse and tablet and it works here but I've come to learn that, that measurement is not good enough so please test it with a mouse and tablet yourself to make sure it doesn't break anything.

### DIFF
--- a/core_lib/src/util/pointerevent.cpp
+++ b/core_lib/src/util/pointerevent.cpp
@@ -1,0 +1,247 @@
+#include "pointerevent.h"
+
+PointerEvent::PointerEvent(QMouseEvent* event)
+{
+    this->mouseEvent = event;
+}
+
+PointerEvent::PointerEvent(QTabletEvent* event)
+{
+    this->tabletEvent = event;
+}
+
+PointerEvent::~PointerEvent()
+{
+}
+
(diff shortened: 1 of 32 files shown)


core_lib/src/util/pointerevent.cpp
```diff
@@ -0,0 +1,244 @@
...
+    }
+    else if (this->tabletEvent)
+    {
+        return this->tabletEvent->pos();
+    }
+    else
+    {
+        // if we land here... the incoming input was
+        // neither tablet nor mouse
+        return mWidgetPos.toPoint();
```
chchwy: Any possibilities the event is not mouse nor tablet?

CandyFace: Hmm no I guess not, well not until we implement another event like touch event.

PR #1178, opened 2019-02-14
Bitmap Coloring layer version 2

davidlamhauge:
I have made some changes based on the response I've got so far, so here is version 2 of the bitmap coloring feature.
A new video can be seen [here](https://youtu.be/KXmzGP-T8Nw).
I look forward to hear comments and suggestions for improvements.


CandyFace:
It indeed looks more simplified and improved David, I however still don't think we should merge this, as I find it too pipeline specific.

A few comments before I venture into why I think this PR should be reworked
This is just information and provides no real value to the user, it's ambiguous unless the user knows the exact method.
image

This whole stack (except the replace line) could be simplified to a spinbox and two radio buttons.
+ Fill area
+ Thin lines
+ Apply to all frames.
image

So this basically reverts the changes?
image

### What should be done instead
What you're doing is what could (and should imo.) be done automatically as part of the import process.
image

A more streamlined version could look something like this (ignore the resample part):
https://youtu.be/ktzGfaYsyb0?t=90

it narrows the whole process down to one window and turns the entire thing into a pre-process step which I imo. find much faster and easier to use.

The color layer becomes needless, there's no need for an extra widget that is only useful when importing scans and the whole process is much simpler.

We could basically snatch that "Black and white" import option and add similar parameters, like the threshold, fill area and thin line features you show in your video.

A preview widget with before and after would also be nice, since it means we would be able to get around any implementation of backup handling but also to provide a nicer user experience as nothing is destroyed until the import button is pressed.

With the current implementation there's simply too much fiddling with parameters and stuff you have to do to get from a to b.

I still think like the idea that you can replace lines and leave the color information intact but I think we could use it differently, for the bucket tool maybe.

davidlamhauge:
I haven't much time now, but I feel a need to clarify something.
As I understand it @CandyFace , then you propose to implement a feature that vectorizes the bitmap drawing, and handles all the process on that layer. 
That would certainly be a welcomed feature, that Pencil2D needs, and must have at some point, but it is not he purpose of my PR.
My PR is towards a bitmap coloring process where you use, keep and preserve the animators artwork in the coloring process. Whether it is animated in Pencil2D or animated on paper and scanned, you maintain the animators work. This means that you must have a separate layer for the coloring. 
The process can be steamlined further, and it could be integrated in the menu in stead of a dock, and the texts could be more clear. It would also be fine with me if the color layer and bitmap layer could be merged in the end, but in the process, I can't see how it could be done without the color layer.

CandyFace:
@davidlamhauge That is not what I mean no, I am exclusively talking about bitmap handling as that is what your PR entails.

Bitmap to vector is out of scope, my proposal is primarily about reworking the execution of your PR design/UI and UX wise.

If you want to color a layer, then create a new bitmap layer, and put it above and use that instead. 
This could be recreated via the ability to duplicate a layer and adding a feature to fill the lines of a drawing.

davidlamhauge:
Thanks for your comments and suggestions @CandyFace, @scribblemaniac and @Jose-Moreno . When I read your comments I see what every software struggles with, namely the two opponents Simplicity and Flexibility. If you want simplicity you loose flexibility, and if you go for flexibility you loose simplicity. We have to find something in the middle that has maximum simplicity without loosing too much of the flexibility.
My first version was very flexible, but I'll admit that it probably was too flexible. So here is what I think should be done:
1) 
If you animate in Pencil2d, then there are no worries, but if you animate on paper and scan the drawings, you must have a way to load these onto a layer. I like Candyface and Joses ideas, to do it in the import dialog. You could have 3 choices, so you could (1) 'Import as is...', (2) 'Import with Threshold...', and (3) 'Import with Threshold and Select...'. Often the animator writes numbers, timing charts and other messages on the drawing, so if we just import the whole image and give it a threshold, we need to spend much time afterwards, cleaning up the imported drawings. That's why it for some scenes will be necessary to select the wanted area from each drawing.
Right now, the threshold is a value you choose between 150 and 250. Default is 200. I could probably make an algorithm that calculates a threshold for the first drawing, and suggests that to be used.
2)
Now let's assume we have a layer with our animation, and the layers name is 'Bob'. Next step is to make a copy of this animation onto a new Bitmap Layer that I call 'Bob_C', thus saving and using the animators original artwork. On this copy of the animation layer, each keyframe is processed so pixels that were 100% transparent are still transparent, and all other pixels are black - or almost black. I use QRgb(1,0,0,255) in the hope that it will be a unique color that noone else uses, but if there is a way to secure that the color is unique, please tell me how. On the color layer (Bob_C) we will set a 'mIsColorLayer' flag, and on the animation layer (Bob) we will set a 'mHasColorLayer' flag.
This black line should be processed to a 1 pixel wide black line that has closed, fillable areas, and is aligned with the animators original drawing. This gives some practical issues:
A) There could be small areas with transparent pixels within the animators line. These should be eliminated, so the 'Line Thinning' algorithm can be more efficient. This is what I use the 'Fill Area' parameter for. To fill areas up to a centain size. Values from 1 to 20 can be chosen. (Right now there is a bug, so 6 is a fixed value, but this will of course be corrected).
B) There could be gaps in the drawing, that needs to be closed. I have no algorithm for that, so the animator/user will have to do it. It would be nice if we could define a Pen for the color layer, that has no antialiasing, is 1-2 pixels wide, and has the uneditable color QRgb(1,0,0,255). Then we would be certain that all editing on the color layer is in the unique color.
These steps A and B has to be taken _after_ the color layer is created. If we did it before, we had to edit the original animators artwork, and that is not right. When the user has closed the open gaps in the color layer drawings, I think it would be fine if we filled the white areas AND thinned the line, in one operation.
3)
The coloring process...
4)
The coloring is finished, and we have to replace the black line with the average of the neighbouring pixels. This should be done in one go, and I think one should have the option to merge the two layers, if they want to. Then we would end up with one nice colored layer, called 'Bob'.

As I see it, no matter how much we automate the process, there will be certain steps that the user has to follow. Step (1) could be in the 'Import' menu, but I would like to have step (2) and (4) in the dock. It could be elsewhere, but no matter where we put them, I think they should be grouped together somehow.

Let's say that I have animated a scene, and colored it. When I'm finished I see that I have used the wrong skin-color on one drawing. Or maybe I've accidently deleted a keyframe, and finds out to late, so I can't undo it. In this case I should have the possibility to go over the whole process again, but just for this one drawing. So that should be included in the dialogs; 'All drawings' or 'Active keyframe'.
You could call the process tedious, but as a person who has 20+ years in the animation business behind me, I can tell that animation IS tedious. It is 2% inspiration and 98% perspiration. You have to be concentrated and focused, and even then you make mistakes from time to time.
I haven't addressed some of the ideas you had Jose. It's mainly because I don't know how they work. The blue and red lines in the scan obviously has a purpose, but it's nothing I've worked with before.
I like the simplicity of Pencil2D. I've said it before and I mean it! It's a part of our common vision. On the other hand I hope there will also be room for more advanced features like this coloring method, a more sophisticated way to define camera movements, and what other ideas we get from time to time. We should retain "...maximum simplicity without loosing too much of the flexibility."

davidlamhauge:
I have made a thorough rewrite, following the mockups made by @CandyFace , and after much discussion with team members, and talks with CandyFace and @Jose-Moreno .
There are still some issues I think, but we're getting closer. It feels good.

davidlamhauge:
coloringfeature

davidlamhauge:
I have two issues:
1) When you have scanned drawings, you check the Threshold checkbox. The button seems in the way. Is it needed?
2) I have two buttons 'Select - Next' and 'Cancel' that are enabled when you press 'Apply' and have chosen to select from scanned drawings. Could we find a more elegang solution? It works fine, though!
3) There is no progress indicator for the tracing, thinning and finishing. Must be done.

davidlamhauge:
coloringfeature
Updates UI - Ubuntu this time...

davidlamhauge:
coloring_ver2_2
Updated to a new and even more simplified look.

davidlamhauge:
After half a year, I revisited this feature and I think it still holds. Pencil2D needs this feature.
I made two small changes:
1) In the combobox at the top, it now starts with 'Choose method...'. Previously it started with the method where you colorize on the same layer. You now have to actively choose method.
2) I've changed the groupbox 'Scanned drawings?' to 'Prepare drawings?'. In the box there was a select tool button to the right. I've changed this button to be a reminder that you're only preparing the selected area. The button/reminder could be left out...

There is also a small optimization in bitmapimage.cpp.

davidlamhauge:
I have updated this feature so it is more logical and user friendly.
1) I have removed the process of making scanned drawings transparent, from the coloring window. The 'Pegbar alignment' entry in 'Edit' is now called 'Prepare scanned drawings', with two menu entries: 'Pegbar alignment' and 'Add transparency to paper'.
In the 'Add transparency to paper', you get a preview window, so you can see how the drawing will look when transparency is added.
2) The coloring process is more informative, and should hopefully give the user a good experience.

I have removed the WIP. This is ready for review.

davidlamhauge:
Her is a 10 minute video, where I show how three scanned drawings end up like three beautiful colored animation frames. New changes are shown.
https://youtu.be/6C4AUjyueYE

Jose-Moreno:
> Her is a 10 minute video, where I show how three scanned drawings end up like three beautiful colored animation frames. New changes are shown.
> https://youtu.be/6C4AUjyueYE

Nice job! 🎉 This is certainly a great step forwards as many exciting opportunities will open just by adding the current feature. 

Once this is reviewed and merged I'm hoping other contributors will be interested in bringing back and expanding the _color layer "bleed" feature_ for the fill tool. I'm sure such tool will synergize well with this coloring feature, allowing the user to color selected target layers automatically by using a a specific layer as an alpha mask reference, such as a lineart layer would be very interesting 👍

Similarly if and when we manage to also get back the _fill expand feature_ the old Pencil used to have, people will be able to color depending on their preferred workflow: either with separate layers like the current implementation or on the same layer without halo artifacts as seen in opentoonz or retas paintman.

Another path of future improvement that will hopefully come once this PR is merged is the automatization of the blending feature, particularly when using the fill tool on color traced drawings to fill highlight and shadow areas directly (i.e color traces).

Either way thank you for opening so many potential doors for enhancement! Once again great job 🙌 😄

davidlamhauge:
Thanks @Jose-Moreno. Yes, the feature has been improved, and is maybe more logical now, where you prepare your drawings first, and then use the coloring feature.

Your old wish for the color bleed and expand feature is not forgotten. The last week I've had many debug sessions with CandyFace, and one thing we talked about (while waiting for compilation) was this feature. If you look at the drawings used in my video, these drawings could easily have been colored with the bleed/expand coloring. My coloring is mainly necessary if you have shadows, highligths and the like.

PR #1181, opened 2019-02-23
Fix canvas bugs

CandyFace:
Fixed not being able to undo brush stroke
Fixed Canvas cursor not updating unless button was pressed.


chchwy:
Thanks @CandyFace

PR #1188, opened 2019-03-02
Fix Jose's Issues

### LINKED ISSUE #677: [BUG] stabilization function create artifacts on drawing strokes

scribblemaniac:
Fixes full pressure stroke/reed issues again. Closes #677. Also fixes duplicate selection box under the same environment.

### DIFF
--- a/core_lib/src/interface/scribblearea.cpp
+++ b/core_lib/src/interface/scribblearea.cpp
@@ -613,6 +613,7 @@ bool ScribbleArea::allowSmudging()
 void ScribbleArea::mousePressEvent(QMouseEvent* e)
 {
     if (mStrokeManager->isTabletInUse()) { e->ignore(); return; }
+
     PointerEvent event(e);
     mMouseInUse = true;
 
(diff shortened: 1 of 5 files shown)


Jose-Moreno:
Well more than just "my" issues, particularly the "reed" pattern since it was found by Kaikogf as well. these were, as far as testing allowed to conclude, windows specific issues found in both version 7 and 10 of that OS after being tested with different tablet manufacturers (genius / wacom). Thanks for fixing these :+1:

scribblemaniac:
@Jose-Moreno Well there "your" issues in the sense that they are seemingly only reproducible in your environment, but I suppose Kaikogf did report something similar once. Since we really don't know for sure why your tablet is behaving the way it is (many other people use Wacom and Windows), there's a lack of a better term to describe this group of issues.

chchwy:
Thanks, everyone @scribblemaniac @Jose-Moreno @CandyFace

PR #1191, opened 2019-03-18
Issue flip

davidlamhauge:
Fixes a bug that made Pencil2D to crash, if you pressed the timeline digits while a flipping was performed.
Changed the timing in flip inbetween slightly, while I was at it. It's better now.


chchwy:
Thanks @davidlamhauge

PR #1199, opened 2019-03-18
Fixes #940 - force dot separator on numbers

### LINKED ISSUE #940: camera layer problem
Ubuntu 16.04
Любая ночная сборка.
При работе со слоем камеры. Если использовать одновременно сдвиг и масштаб то сохранения работают некорректно.
Any night build.
When working with the camera layer. If you use both the shift and the scale, then the save does not work correctly.

CandyFace:
This has been tested on Ubuntu together with @davidlamhauge

### DIFF
--- a/app/src/main.cpp
+++ b/app/src/main.cpp
@@ -22,6 +22,7 @@ GNU General Public License for more details.
 #include <QSettings>
 #include <QFileInfo>
 #include <QDebug>
+#include <clocale>
 
 #include "editor.h"
 #include "mainwindow2.h"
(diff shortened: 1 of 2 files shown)


scribblemaniac:
Will this effect the language that Pencil2D is displayed in, or how numbers are displayed to the user?

CandyFace:
The change does not affect the overall language, only how numbers are written. It does not affect the visual representation in Qt widgets either.

chchwy:
Thanks @CandyFace

PR #1205, opened 2019-03-30
Pegbar alignment

davidlamhauge:
A feature relevant for animators that animate on paper and import their scanned drawings into Pencil2D. If they use the scanners document feeder, they can expect som inaccuracies, so their artwork is not aligned.
This feature alignes the center pegbar of the drawings.
Usage:
1) select the keyframe that is to be the point of reference. This is often a field guide.
2) With the select tool, select an area around the center peg hole, that encircles all center peg holes. 
3) Select peg bar alignment in the menu.
The artwork is now horizontally and vertically aligned.


core_lib/src/graphics/bitmap/bitmapimage.cpp
```diff
@@ -650,6 +650,44 @@ void BitmapImage::drawPath(QPainterPath path, QPen pen, QBrush brush,
...
+    {
+        for (int y = top; y <= bottom; y++)
+        {
+            if (qGray(constScanLine(x,y)) < grayValue)
+            {
+                return x;
+            }
+        }
+    }
+    return 10000;
```
CandyFace: Should return -1 here instead of 10000.

core_lib/src/graphics/bitmap/bitmapimage.cpp
```diff
@@ -650,6 +650,44 @@ void BitmapImage::drawPath(QPainterPath path, QPen pen, QBrush brush,
...
+    {
+        for (int x = left; x <= right; x++)
+        {
+            if (qGray(constScanLine(x,y)) < grayValue)
+            {
+                return y;
+            }
+        }
+    }
+    return 10000;
```
CandyFace: Should return -1 here instead of 10000.

app/src/mainwindow2.cpp
```diff
@@ -44,7 +44,11 @@ GNU General Public License for more details.
 #include "soundmanager.h"
 #include "viewmanager.h"
 
+#include "layer.h"
```
CandyFace: Core files should not exist in app. Only widgets and UI related should exist here.

app/src/mainwindow2.cpp
```diff
@@ -44,7 +44,11 @@ GNU General Public License for more details.
...
 #include "layercamera.h"
+#include "layerbitmap.h"
```
CandyFace: Core files should not exist in app. Only widgets and UI related should exist here.

app/src/mainwindow2.cpp
```diff
@@ -44,7 +44,11 @@ GNU General Public License for more details.
...
+#include "bitmapimage.h"
```
CandyFace: Core files should not exist in app. Only widgets and UI related should exist here.

core_lib/src/graphics/bitmap/bitmapimage.cpp
davidlamhauge: The function 'findLeft' searches for, and returns, the left-most x-value of the peg bar you have selected. This value could in some cases be -1, which is pretty close to the center of the canvas. That's why I chose 10000 as a return value.

core_lib/src/graphics/bitmap/bitmapimage.cpp
davidlamhauge: As above...

core_lib/src/graphics/bitmap/bitmapimage.cpp
CandyFace: Hmm... in that case I'd prefer if we could maybe settle on another return type. Given that the coordinate can be anything, even 10000, returning a custom value that can hold both the intended value but also a potential Errorcode would be better and easier to understand for people trying to read the code too.

So it could be something like... 
```
... will be defined in pencilerror.h
struct StatusInt {
        int value = 0;
        ErrorCode errorcode;
    };

and then return that value in your findLeft and findTop top function.

Now you can simply return a StatusInt and catch the appropriate errorcode instead.
This could come in handy in many other places, so I think that would be a nice addition to the project.

Comment (davidlamhauge) [reply core_lib/src/graphics/bitmap/bitmapimage.cpp]: I have never made a struct before, but I can see that it often is defined as a class.
However, I can't get it to work unless I put it inside the Status class, after the ErrorCode Enum, like this:
```
        ERROR_NEED_AT_LEAST_ONE_CAMERA_LAYER
    };

    struct StatusInt {
        int value = 0;
        ErrorCode errorcode;
    };

```
Is that the way to do it? It must somehow know what ErrorCode is?

Comment (CandyFace) [reply core_lib/src/graphics/bitmap/bitmapimage.cpp]: Yeah that's correct, the struct needs to the origin of the values inside it. Putting the struct above Errorcode isn't allowed because ErrorCode doesn't exist yet.

**Comment by davidlamhauge** [comment]
I've separated the functions, so the core functionality is in the core-lib. I've tested it as good as I can, and it works fine.
One day I would like to know how you "...write tests for it that doesn't rely on UI." @CandyFace

File: app/src/mainwindow2.cpp (hunk 5 above)
Comment (CandyFace) [reply app/src/mainwindow2.cpp]: Remember to remove the header files again that you don't use.

File: app/src/mainwindow2.cpp [hunk 6]
```diff
@@ -416,6 +421,68 @@ void MainWindow2::clearRecentFilesList()
...
+    {
+        QMessageBox::information(this, nullptr,
+                                 tr("No reference Key selected!"),
+                                 QMessageBox::Ok);
+        return;
+    }
+
+    PegBarRegistration* pegreg = new PegBarRegistration(this);
+    QStringList bitmaplayers;
+    for (int i = 0; i < mEditor->layers()->count(); i++)
```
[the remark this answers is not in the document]
davidlamhauge: I had it inside the class, but I use LayerManager to sort out which layers are Bitmap layers. Since core files are not allowed in app, I had to move it to MainWindow2.

davidlamhauge:
I have removed unused headers, and renamed the class to the convention.
It compiles and runs fine, but somehow the checks failed?

CandyFace:
You forgot to rename the _ui file.
```
../../app/src/pegbaralignmentdialog.cpp:2:35: fatal error: ui_pegbarregistration.h: No such file or directory
 #include "ui_pegbarregistration.h"
```

davidlamhauge:
I finally got it renamed - I hope...

CandyFace:
I'll be merging this tomorrow if there are no further comments.

Jose-Moreno:
@CandyFace I can't build David's branch to test right now, it seems to give me a few errors for missing methods on ScribbleArea (i.e deselectAll() and getSelection() ) 

I'll wait until it's merged to test it as there were some issues I presented to David before and I think they are specific to windows since last time he told me he wasn't getting them at all with the same test resources (e.g drawings were not being properly aligned on first pass, zooming affected alignment, operating on already aligned would mess up the alignment instead of preserving it)

scribblemaniac:
Sounds like there might be some merging issues from what @Jose-Moreno said. Additionally, I would like to review this before it gets merged. It shouldn't take too long so please hold out on merging this right away.

davidlamhauge:
I have made the changes so it can compile now.
Tested it on Ubuntu, and it works as expected.

davidlamhauge:
On discord we've (@Jose-Moreno , @CandyFace and @scribblemaniac ) had a discussion whether it would be an enhancement if the dialog was modal. With the last merge the dialog is modal. So far just checked on Ubuntu.
As @scribblemaniac suggested it was a challenge programming wise, but with some tweaking I got it to work :-)

CandyFace:
Two quick comments from your observations in discord:
1) You are able to open several alignment-dialogs. This should be prevented.
This can be prevented by adding a reference to the header and seeing if it exists already, quick fix.
2. Add ```setWindowFlag(Qt::WindowStaysOnTopHint);``` to your dialog window, this will make it stay on top if it's possible. Works on Mac OS.

app/src/mainwindow2.cpp
```diff
@@ -417,6 +416,102 @@ void MainWindow2::clearRecentFilesList()
...
+    disconnect(mEditor->layers(), &LayerManager::currentLayerChanged, this, &MainWindow2::updatePegReg);
+    disconnect(mEditor->layers(), &LayerManager::layerCountChanged, this, &MainWindow2::updatePegRegLayers);
+    disconnect(pegreg, &PegBarAlignmentDialog::cancelPressed, this, &MainWindow2::closePegReg);
+    disconnect(pegreg, &PegBarAlignmentDialog::alignPressed, this, &MainWindow2::alignPegs);
+    disconnect(pegreg, &PegBarAlignmentDialog::layerListClicked, this, &MainWindow2::updatePegReg);
+    pegreg->close();
+    pegreg = nullptr;
+}
+
+void MainWindow2::updatePegRegLayers()
```
CandyFace: This function should be moved into the pegbar class. If you need to access Editor and related core functionality, add a ```void setCore(Editor*)``` in your header and keep a reference of Editor named "mEditor". Alternatively you can add it as parameter in the constructor.

app/src/mainwindow2.cpp
```diff
@@ -417,6 +416,102 @@ void MainWindow2::clearRecentFilesList()
...
+    {
+        if (mEditor->layers()->getLayer(i)->type() == Layer::BITMAP)
+        {
+            bitmaplayers.append(mEditor->layers()->getLayer(i)->name());
+        }
+    }
+    pegreg->setLayerList(bitmaplayers);
+}
+
+void MainWindow2::updatePegReg()
```
CandyFace: This too is only related to pegbar so it should moved to that class.

app/src/mainwindow2.cpp
```diff
@@ -1397,3 +1492,27 @@ void MainWindow2::changePlayState(bool isPlaying)
     }
     update();
 }
+
+void MainWindow2::alignPegs()
```
CandyFace: Same as above, move to pegbar class.

app/src/mainwindow2.h
```diff
@@ -65,6 +66,10 @@ class MainWindow2 : public QMainWindow
     void undoActSetEnabled();
     void updateSaveState();
     void clearRecentFilesList();
+    void pegBarReg();
```
CandyFace: The intent of the function could be clearer, we're launching a dialog, so the method could be called openPegBarRegDialog or pegBarDialogPressed or similar.

core_lib/src/interface/editor.cpp
```diff
@@ -1078,6 +1079,58 @@ void Editor::swapLayers(int i, int j)
...
+        layerbitmap = static_cast<LayerBitmap*>(mLayerManager->findLayerByName(layers.at(i)));
+        for (int k = layerbitmap->firstKeyFramePosition(); k <= layerbitmap->getMaxKeyFramePosition(); k++)
+        {
+            if (layerbitmap->keyExists(k))
+            {
+                img = layerbitmap->getBitmapImageAtFrame(k);
+                retLeft = img->findLeft(rect, 121);
+                if (retLeft.errorcode == Status::FAIL)
+                {
+                    QMessageBox::information(nullptr, nullptr,
```
CandyFace: I understand what you want to do here but Editor and Core in general should not contain GUI feedback because in case a test is made, the program will get stuck if something like a qmessagebox is triggered. Although we don't have many tests for now, it should always be possible to run a method when implementing in core, without triggering GUI feedback.

I think it would be a good idea to have a general error or warning popup that can be emitted from Core and accepted in Mainwindow2. I'll make a PR on your branch.

davidlamhauge:
I have adressed all your requests @CandyFace. 
I forgot to erase some marked out code in MainWindow, but I'll wait to commit this small change, until I know if there will be more requests. Thanks for helping out!

davidlamhauge:
Thanks for the comments @CandyFace .
1) The only way I can eliminate the close-button bug, is to disable the close button in the upper corner. If you know another way, please let me know.
5) The feature looks for peg holes that has a gray value of 121 or less, so peg holes could be red, blue or other colors, but they should have a tone that is medium gray or darker. We could add it as a fourth prerequisite and/or the level could be slightly higher (150-200), but we should not look for white pegholes. Nobody else does, and I've never heard of it before.

PR #1211, opened 2019-04-16
#950 - always make sure there's a keyframe when drawing

### LINKED ISSUE #950: [BUG] Pencil2D crashes when attempting to modify a non-existing first (key)frame

CandyFace:
Fixes #950

### DIFF
--- a/core_lib/src/interface/scribblearea.cpp
+++ b/core_lib/src/interface/scribblearea.cpp
@@ -464,15 +464,6 @@ void ScribbleArea::tabletEvent(QTabletEvent *e)
 
 void ScribbleArea::pointerPressEvent(PointerEvent* event)
 {
-    mIgnoreInput = false;
-    if (currentTool()->isDrawingTool()) {
-        if (!isKeyframeSane())
-        {
-            mIgnoreInput = true;
-            return;
-        }
-    }
-
     if (event->button() & Qt::LeftButton || event->button() & Qt::RightButton)
     {
         mOffset = getCurrentOffset();
@@ -520,8 +511,6 @@ void ScribbleArea::pointerMoveEvent(PointerEvent* event)
 {
     updateCanvasCursor();
 
-    if (mIgnoreInput) { return; }
-
     if (event->buttons() & (Qt::LeftButton | Qt::RightButton))
     {
         mOffset = getCurrentOffset();
@@ -546,8 +535,6 @@ void ScribbleArea::pointerMoveEvent(PointerEvent* event)
 
 void ScribbleArea::pointerReleaseEvent(PointerEvent* event)
 {
-    if (mIgnoreInput) { return; }
-
     if (currentTool()->isAdjusting())
     {
         currentTool()->stopAdjusting();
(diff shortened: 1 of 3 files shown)


core_lib/src/interface/scribblearea.cpp
```diff
@@ -573,28 +586,35 @@ void ScribbleArea::handleDoubleClick()
...
-        VectorImage* vectorImage = currentVectorImage(layer);
-        if (vectorImage == nullptr) { return false; }
+
+    int action = mPrefs->getInt(SETTING::DRAW_ON_EMPTY_FRAME_ACTION);
+    KeyFrame* lastKeyframe = layer->getLastKeyFrameAtPosition(mEditor->currentFrame());
+    KeyFrame* currentKeyframe = layer->getKeyFrameAt(mEditor->currentFrame());
+
+    // the first frame has to exist to use this action
+    if (action == DUPLICATE_PREVIOUS_KEY) {
+        if (lastKeyframe == nullptr) { return false; }
```
scribblemaniac: This can be combined with `&&` to eliminate the unnecessary nesting.

core_lib/src/interface/scribblearea.cpp
```diff
@@ -573,28 +586,35 @@ void ScribbleArea::handleDoubleClick()
...
-    {
-        BitmapImage* bitmapImage = currentBitmapImage(layer);
-        if (bitmapImage == nullptr) { return false; }
+    if (currentKeyframe == nullptr) {
+        // We can assume keyframe is sane here because a
+        // keyframe will be created before drawing on the canvas.
+        if (action != (CREATE_NEW_KEY | DUPLICATE_PREVIOUS_KEY)) {
+            return false;
+        }
     }
```
scribblemaniac: I'm confused about what you are trying to do here. A keyframe will be created if `action == CREATE_NEW_KEY`. I think it's also plausible that `DUPLICATE_PREVIOUS_KEY` should act like create new key if there is no previous key (duplicating the currently shown empty image) but you already return false for that above. When `action == KEEP_DRAWING_ON_PREVIOUS_KEY`, then there's a problem and it should return false. If I was doing this I would just use the last condition I just mentioned. Anyway, whatever you want to do, it doesn't make sense to use `|` because `CREATE_NEW_KEY` and `DUPLICATE_PREVIOUS_KEY` are not set to be bit flags. CREATE_NEW_KEY is 0 and DUPLICATE_PREVIOUS_KEY is 1, so this currently just evaluates to `action != DUPLICATE_PREVIOUS_KEY`. This causes all input to be ignored when using the "Create new key" behavior.

CandyFace: Yeah that's completely bogus... I guess that's what I get from trying to program, late at night...

The idea was basically that if we're on the last keyframe, then we check for if it was DUPLICATE_PREVIOUS otherwise we look at the current frame and check if it's CREATE_NEW_KEY or DUPLICATE_PREVIOUS_KEY. The bitwise condition is bogus, not sure what I thought there.

I'll rework it and yeah it makes much more sense to look at the last action instead.

CandyFace:
I've reworked the PR... basically based on what you said @scribblemaniac , I agree that it's better if we always allow the user to draw on the canvas instead of ignoring their input.

The new logic makes sure that if you draw on a canvas while there is no previous keyframe, then one will be created for you. This is true for both "drawing on previous keyframe" and "duplicate previous keyframe"

core_lib/src/interface/scribblearea.cpp
```diff
@@ -923,47 +902,30 @@ void ScribbleArea::handleDrawingOnEmptyFrame()
...
             break;
+        }
         case DUPLICATE_PREVIOUS_KEY:
         {
             if (previousKeyFrame)
             {
                 KeyFrame* dupKey = previousKeyFrame->clone();
                 layer->addKeyFrame(frameNumber, dupKey);
-                mEditor->scrubTo(frameNumber);  // Refresh timeline.
+                mEditor->scrubTo(frameNumber);
```
scribblemaniac: I know this was here before but could we remove this line? Layer::addKeyFrame calls scrubTo already.

CandyFace: I think you made the same mistake I did. Layer::addkeyframe does not call scrubto, the method Editor::addkeyframe does though. I removed it initially too but then noticed that timeline didn't update as I thought it would, so I added it again :P

scribblemaniac: Yup that's totally what I did. Nevermind 😋

scribblemaniac:
I am merging this now with only one reviewer due to our current limited resources and imminent v0.6.4 release.

PR #1212, opened 2019-04-17
Warnings resolved

davidlamhauge:
I have the vision, that someday we can build Pencil2D without warnings of any kind.
This PR resolves about 60 warnings. It's mainly 0 or NULL that's changed to nullptr, and old C casts changed to C++ casts.


core_lib/src/managers/viewmanager.cpp
```diff
@@ -181,7 +181,7 @@ void ViewManager::rotate(float degree)
 {
     if (mCurrentCamera)
     {
-        mCurrentCamera->rotate(degree);
+        mCurrentCamera->rotate(static_cast<qreal>(degree));
```
scribblemaniac: Given that the camera uses qreal for rotation, we should probably use that as the parameter for this function. Likewise for `ViewManager::rotation`, scaling, etc.

PR #1213, opened 2019-04-19
Bugfixes 2

scribblemaniac:
Here's my second set of bugfixes for v0.6.4, same deal as last time. c0af82a was the last big issue I wanted to fix. Lots of issues remain, mostly vector or selection tool related, but they can wait.

~~e1db3a7 will probably be partially undone by #1211 and will require some extra attention after merging both of these PRs.~~ Edit: #1211 No longer conflicts with my changes here.

### DIFF
--- a/app/src/actioncommands.cpp
+++ b/app/src/actioncommands.cpp
@@ -49,6 +49,7 @@ GNU General Public License for more details.
 #include "aboutdialog.h"
 #include "doubleprogressdialog.h"
 #include "checkupdatesdialog.h"
+#include "scribblearea.h"
 
 
 ActionCommands::ActionCommands(QWidget* parent) : QObject(parent)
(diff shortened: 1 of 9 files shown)


app/src/actioncommands.cpp
```diff
@@ -542,6 +543,11 @@ void ActionCommands::duplicateKey()
 {
     Layer* layer = mEditor->layers()->currentLayer();
     if (layer == nullptr) return;
+    if (!layer->visible())
+    {
+        mEditor->getScribbleArea()->showLayerNotVisibleWarning();
```
CandyFace: I'd like if we could avoid having actioncommands know scibblearea, instead refer to showLayerNotVisibleWarning through Editor via a proxy method that calls scribblearea.

scribblemaniac: Agreed. I have added a proxy method, please check to see if you think it is an acceptable solution.

scribblemaniac:
I am merging this now with only one reviewer due to our current limited resources and imminent v0.6.4 release.

PR #1226, opened 2019-06-18
Constrained rotation

### LINKED ISSUE #1075: Implement transform tool "discrete" rotation with shortcut modifier
### Issue Summary

Ideally when rotating sometimes we require discrete rotation, that is, rotation performed in customizable but exact increments. e.g always rotating 15°,30°, etc for precise work. This implementation would at least require the following characteristics:

- [ ] Move Tool: Use a key short cut modifier (CTRL+SHIFT) to activate "discreet mode" (similar to proportional 1:1 ratio scaling.
- [ ] Move Tool - Options Panel: Have an input box to customize the rotation increment value in degrees.

scribblemaniac:
This PR enables constrained rotation for rotating selections. So when rotating selections, if you hold the shift key, the rotation will snap to the closest angle that's a multiple of n. n is 15 by default, but can be modified in the preferences to any number that divides 360 cleanly. This is to avoid some issues related to the math used.

Rotation Constraint Demo

Completely Fixes #1075


CandyFace:
Except for some refactoring (I've made a pull request) I think it looks good, I have comment though, the first is about the fixed number for hand tool. Either we should use the same preference value as movetool or it should be mentioned that only selected tools uses the preference.

My other comment is about the constant for hand tool, it doesn't respond as well as movetool, when you hold down alt and drag around, you kind of have to wriggle the mouse cursor to rotate but it's still inconsistent and in chunks.

scribblemaniac:
The constrained rotation was not supposed to be added to the hand tool, seems like I forgot about that and did it anyway during the second commit. Unfortunately the hand tool's rotation works fundamentally differently than the move tool right now, and it makes it impossible to properly implement the rotation constraints. Basically, the move tool recalculates the full rotation based on how much the mouse has moved since pressing down, and the hand tool only calculates the partial rotation based on how much the mouse has moved since the last mouse move event. The rotation constraint must round the full rotation, not the delta, which is not currently available to the hand tool.

I will either remove the constraint support for the hand tool all together for now, or I will change the hand tool rotation if it's not too much work. It will use the preference if I get it working.

CandyFace:
I think removing it for now seems like the best choice, then we can merge this and properly fix the problem for hand tool.

scribblemaniac:
> Woo documentation, well done.

A rare sight indeed 😆 . I will merge this tomorrow if no objections come up.

scribblemaniac:
Thanks @CandyFace for the review and improvements.

PR #1233, opened 2019-07-03
Fail to open a pclx with Unicode filenames on Windows

chchwy:
Just got a bug report from the user forum
https://discuss.pencil2d.org/t/miniz-error-doesnt-let-me-record/3755/11

I can confirm the current v0.6.4 will fail to open any .pclx if the filename contains Unicode characters.

The issue is caused by the miniz update. We used to use a modified version of miniz to handle Unicode on Windows (Opening a file by `_wfopen_s()` rather than `fopen_s()`). But the change was overwritten when updating to miniz 2.1.0.

We might need to prepare a hot-fix release as soon as possible for that.

If anyone needs an example, please download this pclx:
https://www.dropbox.com/s/zgzp1vbscykspsb/%E8%A8%B1%E5%8A%9F%E8%93%8B.pclx?dl=0

### DIFF
--- a/core_lib/src/miniz.cpp
+++ b/core_lib/src/miniz.cpp
@@ -2976,7 +2976,7 @@ void tinfl_decompressor_free(tinfl_decompressor *pDecomp)
 #ifndef MINIZ_NO_ARCHIVE_APIS
 
 #ifdef __cplusplus
-extern "C" {
+//extern "C" {
 #endif
 
 /* ------------------- .ZIP archive reading */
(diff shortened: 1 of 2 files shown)


CandyFace:
it might be a good idea to report this bug to miniz, so we can fix it properly. I don't like us applying fixes on a modified version.

chchwy:
@CandyFace This issue has been reported since 2017 and they don't have any good solution.
https://github.com/richgel999/miniz/issues/75

There are 2 reasons I don't send this fix to the upstream:

1. This fix needs `std::codecvt` in C++ standard library to convert a UTF-8 string to std::wstring, but miniz cannot use C++ standard library because it's a C library. However, Pencil2D is written C++, so we are fine.
2. This fix assumes that all the filenames passed into miniz APIs are encoded in UTF-8, which is not true for most of the use cases on Windows. But we are using Qt, QString can be easily converted to a UTF-8 string, so we are good.

That's why I modified the source code. It's a proper fix for the unicode filename issue and it actually gets the job done. I don't understand why you don't like it.

chchwy:
By the way, it's a fix only for Windows platform. No need to do anything for Mac & Linux.

CandyFace:
> richgel999/miniz#75

Ah I didn't mean to sound discouraging, I haven't said I don't like I just don't like fixing dependencies ourselves, since we then rely on updating with the our fixes each time we update the library. We've already had a fix like and that was forgotten until now, so there's no saying when it'll happen again. If we are going to fix dependencies ourselves, then we should fork the project and use that instead.

chchwy:
Yup, you're right. I should have forked miniz when I modified it. I will do it

PR #1240, opened 2019-08-06
Fix translate bug. Fix a warning

chrisju:
fix 2 chinese translation errors

### DIFF
--- a/core_lib/src/interface/scribblearea.cpp
+++ b/core_lib/src/interface/scribblearea.cpp
@@ -902,8 +902,8 @@ void ScribbleArea::handleDrawingOnEmptyFrame()
                 KeyFrame* dupKey = previousKeyFrame->clone();
                 layer->addKeyFrame(frameNumber, dupKey);
                 mEditor->scrubTo(frameNumber);
-                break;
             }
+            break;
             // if the previous keyframe doesn't exist,
             // fallthrough and create empty keyframe
         }
--- a/translations/pencil_zh_CN.ts
+++ b/translations/pencil_zh_CN.ts
@@ -401,7 +401,7 @@
     <message>
         <location filename="../app/ui/colorpalette.ui" line="307"/>
         <source>Large Swatch</source>
-        <translation>打色块</translation>
+        <translation>大色块</translation>
     </message>
     <message>
         <location filename="../app/ui/colorpalette.ui" line="310"/>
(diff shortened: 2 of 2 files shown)


chchwy:
Thanks @chrisju

core_lib/src/interface/scribblearea.cpp
```diff
@@ -902,8 +902,8 @@ void ScribbleArea::handleDrawingOnEmptyFrame()
                 KeyFrame* dupKey = previousKeyFrame->clone();
                 layer->addKeyFrame(frameNumber, dupKey);
                 mEditor->scrubTo(frameNumber);
-                break;
             }
+            break;
```
scribblemaniac: No no no, this was right before. We intentionally want to create a new key in the duplicate mode if no previous key exists so it's as if it is duplicating the empty frames. Hence why there there are comments directly below this explaining that we want to fall through if the previous keyframe does not exist. @chchwy Please revert this.

chchwy: Sorry, my bad :(

translations/pencil_zh_CN.ts
```diff
@@ -1827,7 +1827,7 @@
     <message>
         <location filename="../app/ui/mainwindow2.ui" line="886"/>
         <source>Move Frame Backward</source>
-        <translation>先后移动帧</translation>
+        <translation>向后移动帧</translation>
```
scribblemaniac: Does changing the ts files work or does this need to be changed on Transifex? I was under the impression that it needed be changed on Transifex or it will just be overwritten next time the translations are pulled, but I am not fully familiar with that process.

chchwy: I need to upload this CN.ts file to transifex otherwise it will be overwritten next time as you said.

I will do that.

core_lib/src/interface/scribblearea.cpp
chchwy: fixed.

scribblemaniac: If we move the comments to right below the close bracket on line 909, it apparently will remove the warning on GCC. See https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wimplicit-fallthrough. Wouldn't be a bad idea to do this.

scribblemaniac:
@chrisju Thanks for your contributions. While our code does produce a warning where you change the code, the placement of the break was intentional to have a certain behaviour. We will look into other options to remove the warning.

In case you are not aware, we use the Transifex translation platform to make it easier for people to translate and proofread our application. You can find our project here: https://www.transifex.com/pencil2d/pencil2d/ and apply for the Chinese team if you are interested in contributing further to the translations. We are grateful for any help you can provide.

core_lib/src/interface/scribblearea.cpp
chrisju: Sorry, my reckless

PR #1241, opened 2019-08-06
make spinbox bigger

chrisju:
The spinbox in color inspector is too small.  
I set the mini width of the spinboxs to 55 instead of 0 can fix it.

old:  

new:

### DIFF
--- a/app/ui/colorinspector.ui
+++ b/app/ui/colorinspector.ui
@@ -131,6 +131,12 @@
      </property>
      <item row="2" column="2">
       <widget class="QSpinBox" name="BluespinBox">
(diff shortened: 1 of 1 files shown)


chrisju:
@CandyFace   
I found similar case in QColorDialog in my computer. 
Is it my computer's problem?

chchwy:
@chrisju What OS are you using?

chrisju:
@chchwy Linux Mint 19.1 Tessa,  Xfce desktop

CandyFace:
@chrisju if it affects standard ui components, such as your color picker dialog there, then it's likely caused by the theme you're using.

PR #1242, opened 2019-08-08
Implement last color and  used color history

chrisju:
I implent it at left bottom of Color Inspector. Have 2 buttons. Left one is for switching to last color(has shortcurt Alt+C). Right one is for showing a list of used colors to choose.

origin:  

last color exists:

show used color list:


chrisju:
@chchwy  it is a downloading failed， what can i do？

==> Installing postgis dependency: protobuf
==> Downloading https://homebrew.bintray.com/bottles/protobuf-3.7.1.high_sierra.
curl: (6) Could not resolve host: homebrew.bintray.com
Error: Failed to download resource "protobuf"
Download failed: https://homebrew.bintray.com/bottles/protobuf-3.7.1.high_sierra.bottle.1.tar.gz

CandyFace:
@chrisju don't worry about the CI build failing on Mac... it stalled on some dependency before it could compile. I've compiled it locally and it works fine.

davidlamhauge:
I've tried to compile and run it, and it works fine I think.
There is one thing I would like to see altered. The popup box with recent colors, show the colors in hex values. It should be the names of the colors, i.e. 'Bob eyes', 'Bob shirt' etc. That's more informative and useful.

CandyFace:
@davidlamhauge the names of the colors might not be known at the time, since they're maybe not in the list of colors but simply picked from the color picker, but if that's the case I agree, otherwise it should be generated names.

@chrisju 
I have some feedback regarding the UI:

* The new color widget doesn't have a border like the current color widget, it would be preferable if they look alike.
* The arrow is ambiguous in the current context

Proposal:
I think it would be ideal show a few of the most recent colors, visible on the inspector widget, then the arrow could be next to the last color swatch which would then open a dropdown that would show the remaining swatches with their generated name or hex value.

Clicking on the visible color swatches should change the main color.

A quick and dirty sketch.
image

chrisju:
@CandyFace 
Thanks for your feedback and proposal.
* The current color widget is just a label, but the new color widget is clickable. When a widget has border and is clickable, it seems strange.
* The arrow is ambiguous. Can this resolved by set a tooltip later? And the arrow is ugly but i don't know which image can make it contextual and pretty.

I think your sketch is a little complex.

CandyFace:
> @CandyFace
> Thanks for your feedback and proposal.
> 
> * The current color widget is just a label, but the new color widget is clickable. When a widget has border and is clickable, it seems strange.
> * The arrow is ambiguous. Can this resolved by set a tooltip later? And the arrow is ugly but i don't know which image can make it contextual and pretty.
> 
> I think your sketch is a little complex.

Alright that's fair enough, I've implemented it myself instead since I agree that might have been a bit too much for a first contribution 👍 

For the curious:
stuff

Jose-Moreno:
@chrisju @CandyFace If i may, I would suggest Leaving the color preview as a single element to maintain simplicity, i feel the color preview should always display the current color otherwise it becomes confusing.

Perhaps the overall height of the color history could fluidly become taller when resizing the panel as well in case you need the swatches to be seen bigger.

For both ZZ and your solution i think the drop down menu should go.  Instead of the menu just having the swatches horizontally as you've laid out would look and work better to traverse the color history, and having this carousel styled widget that changes the indexed colors in the predefined array would benefit with selector arrows at either side to control the limited view of the swatches (see image edit below)

If you have more than 2 colors, cycling between colors would be best when using the selector arrows  and these can help preserve space. It also feels more intuitive to move the colors, and since only a set amount of colors would be required to be shown at all times, as in your proposal, this would help  the user always be aware of a portion of the overall history.

Note that a color history is meant to be short listed, If people want the first color the ever used in a session they should rather save it on the swatches for future use. Color History is meant to be used for blending or to fix a mistake in the context of painting / coloring respectively.

Lastly, I also think letting the **current color used** to be shown as selected similarly to the color palette thumbnail selector, that way the swatch on either side of the selection would be the **last color used**

image

Thank you both for your hard work :bowing_man:

scribblemaniac:
I was thinking the same thing as @Jose-Moreno: it is confusing having the two colors. People could mix up the last color and current color, or mistake it for background/foreground colors as many image editing programs have. 

As for resizing, I don't really care, but it could add another row to the recent colors if there is enough vertical space. We don't need a giant box to display just one color.

One other idea that I initially had and I'll just throw out there is that we could have a :arrow_down_small: to the right of the current color that would show the dropdown list to mimic the design of combo boxes like what @chrisju did except to the right, possibly even in the same widget by messing with the stylesheet of a combo box.

chrisju:
I agree with @Jose-Moreno

CandyFace:
satisfaction

What's new:
+ dropdown menu has been removed
+ Added arrow buttons to traverse list
+ Current used color shown as selected
+ Shortcuts for left and right button

@chrisju I've updated the PR on your repo.

chrisju:
@CandyFace I think arrow is not a good design. It is a little complex, and if right arrow is clicked, the last color may be not show in the history, this may make user confusing.

CandyFace:
@chrisju 
What exactly do you find complex? The arrows are merely there to traverse the list of colors. Pressing the right arrow will get you the last color as long as you stand on the leftmost row, but maybe that's the problem? What if the last color is always added next to the selected color, would that help? Pressing right would then always give you the last colors?

Also, if you expand the view then you'll eventually be able to see all colors, it's not fixed anymore.

chrisju:
@CandyFace 
Two complex points than no arrows:  
1. the operation of clicking arrow.
2. Judge a color's history index.

If remove arrows, it looks more simply.

CandyFace:
@pencil2d/developers any comments? should we remove the arrows and just keep it to a list of 5 colors? I can do that I just need to know what you guys think.

The only reason I kept the dropdown menu was because I thought there might have been a reason to do so... Imo. there's no need for more than a few colors.

The first color will still be shown as selected but there will be no color rotation and no shortcuts, although this doesn't provide the ability to switch between last and current color, which was the initial proposal... 
colors

scribblemaniac:
I'm fine with keeping the arrows, but I think they should behave slightly differently. When you click on a non-centered color, I think the colors should shift over so that the selected color is the middle one unless that color is one of the first or last two colors in which case it just shifts as much as possible without creating any empty space. That way if the leftmost color is selected, you know that you're at the last color. Additionally, the arrows should be disabled when you can't move further in that direction.

What is the current behavior when a new color is created while a color other than the last is selected?

CandyFace:
The currently behaviour is that a new swatch is always created from index 0, that is the left most row. Pressing the arrows will simply offset the colors in the list, so there's no end.

CandyFace:
Closing this since the PR doesn't contain the newest changes. We should start a new discussion, revisiting the ideas here and what the goal is though. Will create a new PR when appropriate.

scribblemaniac:
Thank you for your contribution @chrisju! I hope we will be able to get your feature included in a Pencil2D version in some form in the future.

PR #1243, opened 2019-08-12
ignore a specific -Wimplicit-fallthrough warning

chrisju:
ignore a specific -Wimplicit-fallthrough warning


CandyFace:
I'd rather want to ignore that specific warning for now, since there's no harm in having it instead of placing ugly as sin pragma statements into the switch. Alternatively we could increment to c++17 standard, then we could use [[fallthrough]]; but otherwise I would rather ignore it.

chrisju:
@CandyFace That's ok. I will close this request.

PR #1244, opened 2019-08-17
Add function to fill non alpha pixels

davidlamhauge:
This PR is rather simple. It adds a function to BitmapImage, that replaces all pixels inside mBounds with a chosen color.
I think it is a good general-purpose function, and I know I can use it in one of the features I'm working on, + the bitmap-coloring feature + two more features I plan to work on.
Could it be merged right away, it would be easier for me, since I don't have to wait for a more complex PR to be accepted and merged, before the function is in the code base.

### DIFF
--- a/core_lib/src/graphics/bitmap/bitmapimage.cpp
+++ b/core_lib/src/graphics/bitmap/bitmapimage.cpp
@@ -541,7 +541,7 @@ void BitmapImage::setPixel(QPoint p, QRgb colour)
     modification();
 }
 
-void BitmapImage::setAllAlteredPixels(QRgb color)
+void BitmapImage::fillNonAlpha(QRgb color)
 {
     if (mBounds.isEmpty()) { return; }
 
--- a/core_lib/src/graphics/bitmap/bitmapimage.cpp
+++ b/core_lib/src/graphics/bitmap/bitmapimage.cpp
@@ -541,7 +541,7 @@ void BitmapImage::setPixel(QPoint p, QRgb colour)
     modification();
 }
 
-void BitmapImage::fillNonAlpha(const QRgb &color)
+void BitmapImage::fillNonAlphaPixels(const QRgb color)
 {
     if (mBounds.isEmpty()) { return; }
 
--- a/core_lib/src/graphics/bitmap/bitmapimage.cpp
+++ b/core_lib/src/graphics/bitmap/bitmapimage.cpp
@@ -541,14 +541,12 @@ void BitmapImage::setPixel(QPoint p, QRgb colour)
     modification();
 }
 
-void BitmapImage::fillNonAlpha(QRgb color)
+void BitmapImage::fillNonAlpha(const QRgb &color)
 {
     if (mBounds.isEmpty()) { return; }
 
     BitmapImage fill(bounds(), color);
     paste(&fill, QPainter::CompositionMode_SourceIn);
-
-    modification();
 }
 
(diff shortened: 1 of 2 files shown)


CandyFace:
So it's basically a fill operation that preserves alpha, I'd suggest that we call it something like that.
My suggestion would be `void BitmapImage::fillNonAlpha(const QRgb& color)`Aside from that I see no reason not to merge it right away.

davidlamhauge:
I discussed it with @scribblemaniac earlier today, and we struggled to find the right name. I wanted something that cooresponded to 'setPixel', si I called it 'setAllAlteredPixels', but I can easily live with fillNonAlpha.

CandyFace:
Could you also please change "QRgb color" to "const QRgb& color" to the parameter. There's no need to make a copy of the QRgb, we can simply use the reference. 

The difference is:
QRgb color will create a new object and use that, while const QRgb& will use the reference from the object argument and therefore does not create a copy. We make it "const" because it's good practice to make parameters immutable whenever possible.

scribblemaniac:
> Could you also please change "QRgb color" to "const QRgb& color" in the declaration. There's no need to make a copy of the QRgb object, we can simply use the reference.

A QRgb is just a typedef for an unsigned int, so it is probably better to copy it than reference it.

CandyFace:
Ah yes that's true, I was thinking of QColor... well the const argument still stands.

core_lib/src/graphics/bitmap/bitmapimage.cpp
```diff
@@ -541,6 +541,15 @@ void BitmapImage::setPixel(QPoint p, QRgb colour)
...
 }
 
+void BitmapImage::fillNonAlpha(QRgb color)
+{
+    if (mBounds.isEmpty()) { return; }
+
+    BitmapImage fill(bounds(), color);
+    paste(&fill, QPainter::CompositionMode_SourceIn);
+
+    modification();
```
scribblemaniac: Paste calls `modification()`, so we don't need to call it again here.

davidlamhauge:
I am confused - at a higher level.
Should I change it? Or is it, as I understand it, if I use QColor, that I should use const?

CandyFace:
> I am confused - at a higher level.
> Should I change it? Or is it, as I understand it, if I use QColor, that I should use const?

Disregard the QColor bit. Using QRgb is fine but it should still be const.

Jose-Moreno:
Cool. For the name if there's still doubt, how about: `setOpaquePixels()` ? the transparency range would be implied as opaque pixels can only be non-transparent

davidlamhauge:
It may seem like a small issue, but the name is important. Some new coder, five years from now, should find the code base tidy, organised and understandable.
Right now the name is 'fillNonAlpha', which explains what it does.
I would prefer something that corresponds to 'setPixel', that sets the QRgb og one pixel. How about calling the function 'setNonAlphaPixels'?

CandyFace:
In that case call it fillNonAlphaPixels, I'm not fan of using "set" here since that prefix is often used to indicate that we set a member variable with the input parameter, which does not happen here.

Jose-Moreno:
`Non-alpha == Opaque`  If `set` has a different connotation that's fine, then `fillOpaquePixels` would be coherent, or is it the case that `nonAlpha` as an identifier category is being used as convention in other methods? in such case then i guess for consistency it should be `fillNonAlphaPixels` :man_shrugging:

CandyFace:
I'm merging. If we find a bug later, we can fix it then...

PR #1246, opened 2019-08-19
Feature that can Change Line Color on Bitmap layer

davidlamhauge:
If you animate with a black or a dark pencil, it can be hard to make clean ups in black.
With this feature you can change the line color on one or all drawings on the active layer. You can do it as many times as you want, to any color you want.

There are some unresolved issues, that I want your opinion on:
1) There is no warning, when you press 'Change Line Color', Should there be a messagebox confirmation?
2) If you choose to change only one drawing, you must be on that keyframe, to make the change. Should it be enough to be on the keyframe that covers the frame you are on?

Besides that, I haven't implemented any undo-redo yet. I guess I must wait for that to be implemented.


davidlamhauge:
David_Pencil2d_20190819_20_33_33
A short gif to show how it works

app/src/colorpalettewidget.cpp
```diff
@@ -143,6 +151,45 @@ void ColorPaletteWidget::removeItem()
...
+    {
+        if (layerbitmap->keyExists(editor()->currentFrame()))
+        {
+            layerbitmap->getBitmapImageAtFrame(editor()->currentFrame())->fillNonAlphaPixels(newlinecolor);
+            editor()->updateFrame(editor()->currentFrame());
+        }
+    }
+    else
+    {
+        int org = editor()->currentFrame();
```
CandyFace: The convention here is usually to call it currentFrame

app/src/colorpalettewidget.cpp
```diff
@@ -143,6 +151,45 @@ void ColorPaletteWidget::removeItem()
     clickRemoveColorButton();
 }
 
+void ColorPaletteWidget::replaceLineColorSingle()
```
CandyFace: replaceLineColor is singular already, there's no need to add "Single" to it, it would make more sense to me if the method names were switched. This should be named replaceLineColor and the method below should be called replaceLineColorForAllKeyframes.

app/src/colorpalettewidget.cpp
davidlamhauge: I'll change org to currentFrame.

app/src/colorpalettewidget.cpp
davidlamhauge: Sounds reasonable.
I made 'replaceLineColor' first. 
When I later came to think of the possible wish to change line colors for only one drawing, I just added 'Single' to the function name. Your suggestions are clearer.

davidlamhauge:
Should Add, Replace and Remove also be disabled on vector layers?
I could probably have been more precise in my naming. My sole intention is that you can change line color for the current keyframe, or all keyframes on the layer you are working on. I don't think there will be a demand for changig line color on all bitmap layers at the same time.
To be sure that you don't accidentally choose the right option, I've made the context menu like a submenu. I think it works fine.

CandyFace:
I'm looking at the code again and I still come to the same conclusion as previously, I don't think the actions should reside in a context menu in the palette widget. It's not an action based on the palette but rather a modification you can apply to a layer/keyframe, the palette color is just a part of the process.

I'd rather have the actions available in the Layer tab, under its own section
like:
====================
Delete Current Layer 
===================
Change linecolor -> Current Keyframe
Change linecolor -> Layer

davidlamhauge:
My last commit should solve the remaining issues.

PR #1249, opened 2019-08-20
Autosave & Grid preference settings adjustment

Jose-Moreno:
- Autosave max. value set to 8192 in filespage.ui
- Autosave initial value set to 256 in preferencemanager.cpp (this will provide an average of about 5~7 minutes of uninterrupted drawing before the autosave alert kicks in)
- Grid W & Grid H minimum values set to 1px in generalpage.ui

If there's a mistake or a missing alteration that needs to be made please point it out descriptively as usual. Thanks :relaxed:

### DIFF
--- a/app/ui/filespage.ui
+++ b/app/ui/filespage.ui
@@ -46,10 +46,13 @@
          </size>
         </property>
         <property name="minimum">
-         <number>5</number>
+         <number>4</number>
         </property>
         <property name="maximum">
-         <number>200</number>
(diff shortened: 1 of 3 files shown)


CandyFace:
I assume this is a quick fix to never be bothered by the autosave popup ;) Seems fine to be though we really should reconsider the behaviour of that one.

Jose-Moreno:
Thanks guys :smile:

Issue #1125, opened 2018-11-05
Group Onion skin functionality in a panel


Jose-Moreno:
### Issue Summary

This is a request to group the onion skin functionality, that is scattered through Pencil2D, into a single place to avoid segmentation of the user experience and improve the efficiency in workflow as well as to avoid misconceptions (user does not know certain features exist) and confusion (people instinctively click the onion skin button on the timeline > doesn't work > label it as nonfunctional)

I'd like to ask for a new onion skin panel to be designed and implemented as well.

**Discussion**: The problem with having the display panel hold the onion skin at the moment vs having it grouped with the other parameters, is that if you want to turn just the onion skin options off or the "canvas visibility options" only you have to also hide the other one as well. 
I also think it would be better to leave the display panel as a general purpose panel that handles anything related to how the user views the canvas, and create an onion skin panel to groups said functionality together as other panels do in general (e.g tools, options, palette, etc).

**Note**: This migration would not include any future layer-based onion skin switches or other future widgets that are dependent upon onion skin operations like the proposed "shift & trace" functionality. Just the elements on the preferences and the button on the timeline.

**Migrating functionality**

- [x] Move the Preferences > Tools > Onion skin section outside the preferences and onto the panel. These options are often needed for quick access.
- [x] Move the Onion Skin: Toggle Match Keyframes to the panel
- [x] Move the Onion Skin display during playback from Preferences to the Panel

### Video or Image Reference

**Note**: Only consider the components, not the graphic design itself or the space used by the components.

![image](https://user-images.githubusercontent.com/4912907/48025458-75b09300-e111-11e8-8559-2f002d72d6d0.png)

_Is it possible to have a double slider as in the mockup with QT? or would it have to be programmed from scratch?_

scribblemaniac:
> Is it possible to have a double slider as in the mockup with QT? or would it have to be programmed from scratch?

There is currently no qt widget for a double slider like that. We may be able to use [this code](https://stackoverflow.com/questions/17361885/range-slider-in-qt-two-handles-in-a-qslider?noredirect=1&lq=1) as a starting point though.

[...]

PR #1252, opened 2019-08-22
Move onion skin mode selector to the preferences window

scribblemaniac:
We all should be well aware by now of how much confusion that "onion skin" button in the timeline is. Well now it's in the preferences with the other onion skin options and has been better described so there should be no more confusion.

Eventually this will be superceded by #1125, but I think this is a good short- to mid-term solution until an onion skin panel is implemented.


scribblemaniac:
> A general comment though, shouldn't we move the onion skin settings into the timeline preferences? it feels misplaced to have it under tools.

Yes I had the same thought when working on this, but I did not want to complicate this PR. I agree that the timeline page would be a more appropriate place for it.

chchwy:
Looks all good to me.

PR #1255, opened 2019-08-25
Split tool logic for bitmap and vector

CandyFace:
There's a lot of code but it's mostly just copy/paste and removing the obsolete code per layer, so it shouldn't be that bad. I've tried to avoid changing logic anywhere, so nothing should be broken.

I've cleaned up a lot of boilerplate code and removed all the unnecessary layer checks... This should make it much easier to maintain the tools in the future.

Additionally I've tweaked the tool options UI a bit, though I can revert that..
tooloptions

All tools has been tested and seems to work as intended...


CandyFace:
Closing this for now, there are a lot of other pull requests that should be looked at before this one.

PR #1256, opened 2019-08-26
Add preset support

### LINKED ISSUE #981: Implement canvas properties "preset" manager

scribblemaniac:
This PR introduces "presets" to Pencil2D which allow users to almost completely customize what they get upon startup and easily switch between different startup options. In the preferences, there is a new section which looks like this:
Preset preferences
Pressing "+" will add a new preset that appears exactly the same as the current project. Pressing "-" will remove the currently selected preset (except for the "Default" preset which is always there). Pressing "Make Default" will make the selected preset that is first shown on the list.
If "Ask on startup" is unchecked, then the default preset will automatically be loaded. If it is checked, then this dialog will pop up whenever you create a new project:
Preset dialog
If you cancel it will open the regular Pencil2D default (not the user default). If you accept, it will load that preset from the dropdown box. If you accept and the checkbox is enabled, it will set the preset from the dropdown box to be the default and it will set ask on startup to false, causing that preset to always be used without the dialog.

Behind the scenes, a preset is little more than a pclx file saved to a hidden folder that is opened on startup. The main difference is that saving will not overwrite the preset file, but will act as if it is an entirely new file. This means that you can save many things in the preset including but not limited to the camera size, the number names and configuration of layers, color palettes, and even actual drawings and animations. The only things that aren't saved are window configurations and preferences, which remain set globally.

By default there are no presets other than the default.

Closes #981


davidlamhauge:
I can't wait for this to be reviewed and merged into the master.
I also wish I could review it, but I'll leave that to those with better skills.
When I get home, I'll downloaad and test the feature.
Until then - Good work @scribblemaniac !

davidlamhauge:
I've tested this PR and it works as expected. 
The only thing I couldn't understand, was that there was a default preset called "TVanimatic 16:9", when I started the banch up first time. I thought that there should be a "default" and nothing else, from the start.
Otherwise I think that it's easy to use, and that you have full control to create the presets that  you want. Can't wait to see it reviewed and merged.

scribblemaniac:
@davidlamhauge Are you sure you haven't tried this branch before? I mean you can see the code as well as I can, there's nothing in there about "TVanimatic 16:9". I can't imagine how else it got in there.

davidlamhauge:
It was on a computer at work. I had a pause, so I tested it.
I made a grep on the source, and couldn't find "animatic", so I was really puzzled. The explanation must be, that I have tried the branch on it once before, deleted the branch, and now, when I try it again, it is saved there, somewhere.

scribblemaniac:
That makes sense. The presets are saved to `QStandardPaths::AppDataLocation/presets`, which varies by operating system, but the paths can be found here: https://doc.qt.io/qt-5/qstandardpaths.html.

CandyFace:
1. I find it difficult to understand what the changes of a preset entails, especially if it doesn't concern the UI much. I think there's a need to explain what's included in a preset and what's not because currently it's not easy to figure out unless you know what's included in a pencil project file.

2. The welcome/preset window appears before the main window, is there a reason for that? 

> The main difference is that saving will not overwrite the preset file, but will act as if it is an entirely new file
3. So changes are saved all the time or how do I make sure my changes are saved?

davidlamhauge:
As @CandyFace mentions, there could be more explanation on how it works, but when you know how it works, it's so easy to use that you never will need to see that information again. Therefore, I'll suggest that you can open a help-window from the dialog, in stead of writing all the explanations on the dialog every time.
Besides that I have two comments:
1) The preset window should be hidden by default, so Pencil2D starts up as it does now, unless you actively check the "Ask on startup" box.
2) "+" and "-" on the dialog could be replaced by "Add preset file" and "Remove preset file". I write 'could' in stead of 'should', since it's not that important to me.

Whether or not the preset window appears before the main window, is not that important to me either.

CandyFace:
Alternatively instead of adding an explanation, we could have a details window that lists all the settings listed in the project file. This would make it easy to understand what the preset contains and what it alters... 
Ideally this could also configure the project file directly but that could be a quality of life thing...

chchwy:
I am reviewing this PR.

I think we need a clear description somewhere for users about what the plus button actually does. Take me as an example, I have no clue what it's using as a Preset. An informative sentence like "Save the current project as a Preset" is much better than a "+".

chchwy:
Also found a crash when creating a new project. I am working on it.

core_lib/src/structure/object.cpp
```diff
@@ -61,6 +61,8 @@ void Object::init()
 
     // default palette
     loadDefaultPalette();
+
+    createDefaultLayers();
```
chchwy: `createDefaultLayers()` This method is intentionally not placed inside the `Object::init()`. Doing so breaks many unit tests.

app/src/presetdialog.cpp
```diff
@@ -0,0 +1,100 @@
...
+    QDialog(parent),
+    ui(new Ui::PresetDialog),
+    mPrefs(preferences)
+{
+    ui->setupUi(this);
+
+    initPresets();
+
+    connect(ui->buttonBox, &QDialogButtonBox::accepted, this, &QDialog::accept);
+    connect(ui->buttonBox, &QDialogButtonBox::rejected, this, &QDialog::reject);
```
chchwy: The two buttons have been connected in the `presetdialog.ui`. No need to connect in .cpp again, I will be removing them.

Issue #1105, opened 2018-10-07
Implement scrub playback tracking in timeline view


Jose-Moreno:
### Issue Summary

After using the timeline for a bit, I noticed that during playback and when manipulating the playback scrub header it's difficult to track it after it "leaves" the current view.

I feel the scrub should always be visible on the timeline view either during playback, when we use the start / end buttons on the playback controls or when we use the frame / keyframe jumping shortcuts.

This would mean the timeline view should "keep track" of the scrub until it reaches the end of the animation (last keyframe). 

A possible behaviour would be to start following the scrub only when the scrub reaches between 50% ~ 70% of the timeline (or similar to the frame length algorithm) moving the timeline horizontal scroll to match the scrub.
That way there wouldn't be an issue when viewing the first keyframe and tracking up to the last keyframe would be possible as well. 

As a separate case if the user get's to the last maximum frame possible (9999) the timeline should stop tracking when you reach the last portion of the timeline (about ~1% of the current frame length)

davidlamhauge:
Shouldn't this be closed?

Jose-Moreno:
@davidlamhauge I didn't know the PR was merged. I don't have time to test this right now but i'll be free next week, so once I test it I'll close it, apologies for the delay.

Jose-Moreno:
@davidlamhauge Hi David, I tested the latest master, and well so far the red time indicator seems to be getting "followed" only when it gets to the last part of the visible panel and that's fine.

One thing though is that when the playback occurs, if you drag the horizontal scroll bar during playback the timeline does not update and the panel will incorrectly display the frame numbers, and it will updated only after it reaches the final part of the panel.

It was my understanding that this issue had also been fixed on your PR, so I'm not sure why I can't see the effects of that. I'm using Windows 7 Ultimate for my tests by the way. Thanks for your work :slightly_smiling_face:

Jose-Moreno:
Ok I finally tested David's #1257 on the latest dev build. think this is good enough.

Right now the only issue I see is a rare corner case where dragging the horizontal scrollbar for a long animation, and doing so during playback will not properly update the frame dial numbers (unlike when the playback is paused) . The timeline visible region will only update until the playback indicator reaches near the end of the visible portion of the timeline.

That means, even if we move the timeline region during playback, only the current frame that is being read by the playback indicator gets updated, but the regular dial numbers display at the top will not be updated until the playback conditional case is triggered and the region is updated as a whole.

[...]

PR #1257, opened 2019-08-26
#1105 Added checks on timelinecells, so scrub is tracked

davidlamhauge:
This PR see to it, that the scrubber always is visible during playback.
It is a rather simple solution, since we shouldn't do too complicated things on the timeline, while we wait for the Timeline rewrite.
This one has bugged me (and @Jose-Moreno) and possible others for some time, so here is a solution.
1105_scrubPlaybackTracking

### DIFF
--- a/core_lib/src/interface/timelinecells.cpp
+++ b/core_lib/src/interface/timelinecells.cpp
@@ -781,7 +781,10 @@ void TimeLineCells::trackScrubber()
     else if (width() < (mEditor->currentFrame() - mFrameOffset + 1) * mFrameSize)
     {
         // Move timeline forward if the scrubber is offscreen to the right
-        mFrameOffset = mFrameOffset + ((mEditor->currentFrame() - mFrameOffset) / 2);
+        if (mEditor->playback()->isPlaying())
+            mFrameOffset = mFrameOffset + ((mEditor->currentFrame() - mFrameOffset) / 2);
+        else
+            mFrameOffset = mEditor->currentFrame() - width() / mFrameSize;
         emit offsetChanged(mFrameOffset);
         mTimeLine->updateContent();
     }
--- a/core_lib/src/interface/timelinecells.cpp
+++ b/core_lib/src/interface/timelinecells.cpp
@@ -425,7 +425,7 @@ void TimeLineCells::paintEvent(QPaintEvent*)
             paintOnionSkin(painter);
         }
 
-        if (mPrevFrame != mEditor->currentFrame())
+        if (mPrevFrame != mEditor->currentFrame()  || mEditor->playback()->isPlaying())
         {
             mPrevFrame = mEditor->currentFrame();
             trackScrubber();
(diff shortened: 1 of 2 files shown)


core_lib/src/interface/timelinecells.cpp
```diff
@@ -425,8 +425,20 @@ void TimeLineCells::paintEvent(QPaintEvent*)
         }
 
         // --- draw the position of the current frame
```
CandyFace: Please update or move the comment so it matches the behaviour context.

core_lib/src/interface/timelinecells.cpp
```diff
@@ -425,8 +425,20 @@ void TimeLineCells::paintEvent(QPaintEvent*)
...
-        if (mEditor->currentFrame() > mFrameOffset)
+        if (mEditor->playback()->isPlaying() && mEditor->currentFrame() < mFrameOffset)
         {
+            mFrameOffset = mEditor->currentFrame() - 1;
+            emit offsetChanged(mFrameOffset);
+            mTimeLine->updateContent();
+        }
+        else
+        {
+            if (mEditor->playback()->isPlaying() && width() < (mEditor->currentFrame() - mFrameOffset + 1) * mFrameSize)
```
CandyFace: The code here could be tied up by putting it into a separate function, calling it trackScrubber() or something similar.

davidlamhauge: Moved tracking to separate function. It seem to work like before.

CandyFace:
Since the number of changes here are small, we can merge this right away, however to give the chance of someone else to review it, I'll wait till tomorrow before merging.

davidlamhauge:
It is indeed small changes, but let it rest a day if someone else wants to review it. Thanks!

CandyFace:
I agree that would be useful.

davidlamhauge:
Two small errors fixed.
When moving left, the scrubber went out of sight.
If you used the horizontal scrollbar to move the scrubber out of sight, and then pressed Play, the scrubber would remain out of sight, until it reached the place you scrolled to.
Now everything should work as expected, whether you Play, Scrub the timeline, or move to previous/next frame or keyframe.
Feel free to test it!

davidlamhauge:
I have made so it behave likewise forwards and backwards.
I don't understand the last paragraph about pausing the scrubber, so I haven't addressed that.
I meant this PR to be a quick-fix for an old problem, while we're waiting for the new timeline, but I'll of course be ready to make more changes, if they are wanted.

davidlamhauge:
It is less timeline updates, that's the reason for jumping halfway back, when it goes of screen. I thought that it would be too much if it went all the way over to the left.
But - I am ok with letting it jump to the far left, if we agree on that.

scribblemaniac:
@CandyFace Do you have any comments on the discussion about how far back the scrubber should go during playback?

CandyFace:
> @CandyFace Do you have any comments on the discussion about how far back the scrubber should go during playback?

Not really, I think it's fine to scrub to the middle of the view.

CandyFace:
Two developers has approved the changes, so I'm merging. If there's a need to further discuss the functionality, a new issue can be created.

PR #1258, opened 2019-08-31
Movie import

### LINKED ISSUE #1070: Implement video import
### Issue Summary

This is a request to implement direct video import into Pencil2D with support for formats like: AVI, MP4, MKV, OGV

Video import is a quality of life feature that can be used for reference (e.g self-shot video for acting analysis purposes) or rotoscopying actions (for further visual reference on this technique look for modern movies like "Waking Life" or "A Scanner Darkly").

**Considerations**
This implementation could be done by converting the relevant video to an image sequence using FFMPEG and then importing it into a dedicated bitmap layer.

We could also consider creating a new type of layer (video layer) that would reference the video file externally instead of copying it inside the pencil2d compressed project folder, as the possibility of having bigger (and uncompressed) video files could lead to longer saving files and possible corruption of the PCLX data. This would also allow for a quicker cache / playback of the video inside Pencil2D, pretty much like your own video player inside but with the purpose describe above.

scribblemaniac:
This PR adds the ability to import movies. From the usual import menu, there are now two options, "Movie Video..." and "Movie Audio..." which are meant to be used to extract the frames to a bitmap layer and the audio to a sound layer respectively.

Honestly I was feeling a bit lazy while doing this so I know there is room for improvement. My priority was just getting a working version ready for the next release as this is a very helpful feature for those users who aren't already experts with ffmpeg :wink: There is still the opportunity to add an import dialog with additional features to this such as: only importing from a specific start and end time (easy for ffmpeg), audio/video channel selection, or even some kind of preview. There could be some progress bar for the audio import, but you'd have to parse the duration out from the ffmpeg output which is not as easy as the frame number for the video import. The error messages are very unhelpful. There are a few things in the import code I came across that probably need refactoring, particularly moving some things out of mainwindow. Just ideas for the future, I personally think this should be good enough for use as-is.

Closes #1070


davidlamhauge:
It is a very basic implementation, but it looks very promising.
When testing it, I accidentally started importing a 10 minute movie, and I had no chance to abort it. This should be fixed.
Otherwise I haven't found problems, but there is room for improvement.
I've not done a thorough test, but aside from the ability to abort the import, there are nothing wrong with merging this, and see what happens when the users use it.

Jose-Moreno:
Excellent work! This has been a most requested over the years, so it is a welcome addition!

I think however that this one could wait at least for a progress bar with a cancel button before merging. Indeed having an import dialog with preview and trimming options (start / end time) would be great for future enhancements, but for the bare minimum of visual feedback, a progress bar, and the ability to cancel the operation are required to communicate to the user what is happening and that there is are no (obvious) errors, or if there's a user mistake to roll back the operation before waiting too long (as David commented). 

I honestly can't tell you how many times I've thought a program has locked up because I didn't see visual feedback for long import operations and closed the program thinking it crashed :joy: 

With the new undo / redo scheme this should also have an accompanying action that rolls back all the imported frames, or reapplies the imported frames for a movie clip in case there was an unattended user error and the amount of frames is too long to delete manually frame-by-frame (and because consistency).

Since I know your time is limited, maybe if someone else want to contribute to your branch and consequently to this PR to add these qualifying specs, it should be possible. Either way thanks again for tackling this, hopefully the others can chime in with their comments :pray:

scribblemaniac:
All the operations are practically cancellable now.

scribblemaniac:
@CandyFace There's plenty of room for improvement in that regard for whomever is adventurous enough. The ffmpeg/ffprobe binaries could be decreased significantly by compiling a more minimal version. The Qt components could probably be reduced with [Qt Lite](https://blog.qt.io/blog/2017/05/31/qt-lite-qt-5-9-lts/). You could remove the ffprobe dependency entirely if you parsed the duration from the ffmpeg output, but that would be a pain and much more prone to breaking in the future.

Of the things we can optimize with the application (speed, memory, and size), I think size is the least important of them. Of course we shouldn't be reckless with it, but space is a fairly abundant resource for most modern systems, and app sizes are naturally larger than they used to be.

scribblemaniac:
I've sweetened the deal :grin: Now this also adds support for many more audio file formats (*.wma *.ogg *.flac *.opus *.aiff *.aac *.caf) and "fixes" the mp3 playback issue on windows by converting it to a wav on import.

CandyFace:
> I've sweetened the deal 😁 Now this also adds support for many more audio file formats (*.wma *.ogg *.flac *.opus *.aiff *.aac *.caf) and "fixes" the mp3 playback issue on windows by converting it to a wav on import.

Awesome, I think it was always a good idea to convert the audio files to wav internally when importing, to make sure codecs wouldn't be a problem. The user doesn't notice anyway so it just makes it easier for us.

This doesn't make use of FFprobe though does it?

Jose-Moreno:
@scribblemaniac Awesome! A quick question though, will these formats be only for videos with audio containers with those codecs or will this improvement expand to the sound import dialog as well? (I hope it does :pray: )

scribblemaniac:
> This doesn't make use of FFprobe though does it?

No it does not, but that's mostly because there is no progress bar for audio import.

> A quick question though, will these formats be only for videos with audio containers with those codecs or will this improvement expand to the sound import dialog as well?

They are are part of the Import > Sound action, but they use the same function as the movie audio import behind the scenes. It should work with virtually any audio codec in videos already.

core_lib/src/interface/editor.cpp
```diff
@@ -863,6 +864,208 @@ bool Editor::importImage(QString filePath)
...
+    }
+
+    QString ffmpegPath = ffmpegLocation();
+    if (!QFile::exists(ffmpegPath))
+    {
+        qDebug() << "Please place ffmpeg.exe in plugins directory";
+        return false;
+    }
+
+    // Get frame estimate
```
davidlamhauge: You make a frame estimate here. I think it should be used to limit the amount of frames to be imported.
I imported a mp4, that lasted 13 seconds. It took a minute.
After that I tried to import a mp4 that lasted 3:56 minutes. After 4-5 minutes it started importing pictures. After importing about 350 frames, the software crashed.
What is the biggest video you've tried to import?
Shouldn't we set a maximum of 300-500 frames?

scribblemaniac: There is no technical limit. If you are on Windows, it probably crashed because you ran out of memory, but that's entirely dependent on the system and is difficult for us to set a good limit on. I can't test this right now, but I will later.

One solution is to output the frame from ffmpeg directly to the temporary folder (or moving files into the temporary folder), which prevents having to have all the frames loaded into memory and would cut the import time in half, however this would not be good for pcl files where writing to the temporary folder is practically the equivalent of saving.

scribblemaniac:
FFprobe is optional now :expressionless:

core_lib/src/interface/editor.cpp
CandyFace: We could load in chunks at a time, since we can't rely on people understanding that their video might generates thousands of images, this will still take a hit in loading time but imo. it's better than crashing the program and won't be as bad as moving frames to the temporary folder for each frame.

CandyFace:
Can be merged as soon as the conflicts has been resolved.

scribblemaniac:
I will resolve the conflicts soon, but I'm actually going to move this to [WIP] for now. I have an idea to reduce the memory burden of this feature.

scribblemaniac:
I've just implemented the idea mentioned earlier. Frames are saved to the disk right away (but are not part of the project file until it is actually saved). This means lower memory usage, less crashing, and faster saving. Unfortunately this only works for frames being added to a space without an existing keyframe and there's no way around that since the import frame has to be merged with the existing frame.

I've also resolved the merge conflicts. I believe this feature is ready to be added. There are a few quality of life improvements I would still like to see added including import settings similar to the image sequence import (spacing, alignment, ...), better temporary folder cleanup, better error handling, and warnings for large imports, however I feel that this is good enough now that it should be added to v0.6.5. The new audio import is particularly important with many of our users still having issues with mp3 files. Given that @davidlamhauge and @CandyFace have already reviewed most of this, I am hopeful that it will not take too long to finish reviewing.

CandyFace:
Awesome, I will review it today.

app/src/actioncommands.cpp
```diff
@@ -62,6 +64,118 @@ ActionCommands::ActionCommands(QWidget* parent) : QObject(parent)
...
+                             st.title(),
+                             st.description(),
+                             QMessageBox::Ok,
+                             QMessageBox::Ok);
+    }
+    else if (!st.ok())
+    {
+        ErrorDialog errorDialog(st.title(), st.description(), st.details().html(), mParent);
+        errorDialog.exec();
+    }
```
CandyFace, line 169: This is not correct, now the user will see an empty dialog every time they cancel the operation. There's no explanation, just a window with a exclamation mark. The reason reason it was !st.ok() && st != Status::CANCEL is because it's only meant to trigger a popup if the operation fails.

Right now you'll see this if you cancel 
image

app/src/actioncommands.cpp
```diff
@@ -62,6 +64,118 @@ ActionCommands::ActionCommands(QWidget* parent) : QObject(parent)
...
```
CandyFace, line 117: Same as below..

app/src/actioncommands.cpp
```diff
@@ -131,6 +254,44 @@ Status ActionCommands::importSound()
...
+    return st;
```
CandyFace, line 292: Same as above

core_lib/src/interface/editor.cpp
```diff
@@ -25,6 +25,7 @@ GNU General Public License for more details.
 #include <QImageReader>
 #include <QDragEnterEvent>
 #include <QDropEvent>
+#include <QProgressDialog>
```
CandyFace, line 28: We should remove this again

core_lib/src/interface/editor.h
```diff
@@ -27,6 +27,8 @@ GNU General Public License for more details.
 
 class QDragEnterEvent;
 class QDropEvent;
+class QProgressDialog;
```
CandyFace, line 30: This too should be removed

app/src/actioncommands.cpp
scribblemaniac: Yeah, I was trying to get a message to display when it was cancelled because it exceeds the frame limit so I thought I'd change the user cancelled events to SAFE, but I did not realize how many points there were where it could be cancelled by the user. And I also didn't want to use ErrorDialog for it because we don't need the extra debug information for importing something that is too long. I have updated it with a solution that should work better. Let me know what you think.

PR #1261, opened 2019-09-09
Layer import, from *.pclx to active project

davidlamhauge:
The ability to reuse animation is essential today. Not just characters, but also rain, snow and other effects can be reused.
This PR makes it possible to import layers from whatever *.pclx file you want.
I don't think the import should be in the undo/redo, since it doesn't affect the other layers, and it's easy to delete layers if you want.
It's not a huge PR, but I think it's impact will be...

### DIFF
--- a/app/ui/mainwindow2.ui
+++ b/app/ui/mainwindow2.ui
@@ -984,7 +984,7 @@
   </action>
   <action name="actionImportLayers_from_pclx">
    <property name="text">
-    <string>Layers from *.pclx</string>
+    <string>Layers from PCLX...</string>
    </property>
   </action>
  </widget>
--- a/app/ui/mainwindow2.ui
+++ b/app/ui/mainwindow2.ui
@@ -49,7 +49,7 @@
      <x>0</x>
      <y>0</y>
      <width>800</width>
-     <height>22</height>
+     <height>26</height>
     </rect>
    </property>
    <widget class="QMenu" name="menuFile">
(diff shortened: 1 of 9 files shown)


CandyFace:
This looks nice @davidlamhauge and code wise I have no problems with it. I discovered a bug though that crashes the application.

1. Open the "Import Layers from other *.pclx files"
2. Click on Select file
3. press cancel
The result is that the program crashes, presumably because you don't handle the cancel state, since it shows the progress bar briefly.

Feature wise I think it would be nice if it showed a few images of the content or something from a layer but aside from that I think it looks good. We can always improve it.

davidlamhauge:
David_Pencil2d_20190910_22_19_38
Here's a gif that shows the feature. You can import Bitmap, Vector, Sound and Camera layers.
@CandyFace  I've fixed the bug, so I now check that a filename is returned from the open file dialog. I'll commit the changes within minutes.

app/ui/mainwindow2.ui
```diff
@@ -981,6 +982,11 @@
     <string>Peg bar Alignment</string>
    </property>
   </action>
+  <action name="actionImportLayers_from_pclx">
+   <property name="text">
+    <string>Layers from *.pclx</string>
```
CandyFace: The import name doesn't have to be this explicit, I think we could call it "PCLX" without anything else and that would be fine.
A project file consists of layers and frames, so there's no need to mention that explicitly.

davidlamhauge: Just to be sure: you want me to change 'actionImportLayers_from_pclx' to 'actionImport_PCLX'?

CandyFace: Naah the action itself is fine I was talking about the UI string that is shown 
eg. "Layers from *.pclx" -> "pclx"

davidlamhauge: I could do that but I think it's too uninformative. Wouldn't it give the impression that you import the whole pclx file, without being able to pick from it?

CandyFace: Well you can technically import the whole project file so I don't think that's wrong but in that case it could be written as "Layers from pclx" The extension is imo. unnecessary.

PR #1262, opened 2019-09-11
#1174 More consistent palette import behavior

### LINKED ISSUE #1174: Inconsistent palette import behavior
### Issue Summary

Depending on which palette file format is used, the new palette may append or overwrite the existing palette.

### Actual Results

When importing a Pencil2D palette (.xml), the existing palette will be replaced by the new palette. When importing a GIMP palette (.gpl), the existing palette will remain and the new palette will be added to the bottom.

### Expected Results

I would expect the behavior to be consistent regardless of the file format being imported.

### Steps to reproduce

Here are two palettes to make reproducing the bug easier: [testpalettes.zip](https://github.com/pencil2d/pencil/files/2854567/testpalettes.zip)

1. Create a new project
2. Import the testpalette.gpl file
3. Observe how the Vivid Purple color has been added to the end of the default palette.
4. Create a new project
5. Import the testpalette.xml file
6. Observe how the default palette colors have been removed and replaced with the Vivid Purple color.

### System Information
+ Pencil2D Version: January 6th Nightly Build (0dde212a5c55658fa65b4e2f9788632b3ba02235)

+ Operating System: macOS 10.13.6

+ RAM Size: 16 GB

+ Graphics Tablet: None

<hr>

This issue shouldn't be too hard to fix, but we need to discuss what behavior is desired. Personally I think the current gpl behavior is best because then you can easily import multiple palettes for the same project. Having said that, maybe it would be good to filter out duplicates that have the same name-color pair.

davidlamhauge:
David_Pencil2d_20190911_20_20_22
This PR addresses the inconsistent behavior on opening, importing, saving and exporting palettes.
As we agreed, this is how it goes:
1) Opening a palette, replaces the existing palette with a new one. If at least one color is in use on a vector layer, a warning is given. Only one warning.
2) Importing a palette, appends the new palette to the old one.
3) Saving a palette can be done to the Pencil2D xml format.
4) Exporting a palette can be done to GPL, XML, and other formats as soon as the formats are implemented.
I hope I've covered all (or most) issues we've talked about.
This closes #1174


app/src/mainwindow2.cpp
```diff
@@ -1295,6 +1297,54 @@ void MainWindow2::importPalette()
...
+    {
+        mEditor->object()->exportPalette(filePath);
+    }
+}
+
+void MainWindow2::openPalette()
+{
+    int count = 0;
+    int maxNumber = mEditor->object()->getColourCount();
+    bool openPalet = true;
```
CandyFace: typo, should be openPalette. It's a minor thing but makes it much easier to search.

app/ui/mainwindow2.ui
```diff
@@ -81,6 +82,7 @@
      <addaction name="actionExport_Animated_GIF"/>
      <addaction name="separator"/>
      <addaction name="actionExport_Palette"/>
+     <addaction name="actionExport_Other_Palette_formats"/>
```
CandyFace: I don't see a need to have multiple export actions, just provide one "Export Palette" then let the dialog. Remove the Export palette and keep the dialog where you can select extension.

core_lib/src/structure/object.cpp
```diff
@@ -516,7 +516,7 @@ void Object::importPalettePencil(QFile& file)
     QDomDocument doc;
     doc.setContent(&file);
 
-    mPalette.clear();
+//    mPalette.clear();
```
CandyFace: This should be removed

davidlamhauge:
I have merged the newest Master branch into this branch, and fixed the typs.
I also removed the option to choose whether to export as Pencil2D or GIMP format. It works, but on my Ubuntu you must add suffix yourself.
I'm too tired to fix it today... :sleepy:

davidlamhauge:
Now, there is always an extension.

davidlamhauge:
It again saves without file extension in Linux (Ubuntu 18.04).
Otherwise it works fine @CandyFace

CandyFace:
The save without extension on linux should be fixed via https://github.com/pencil2d/pencil/pull/1299

CandyFace:
I'd say this is ready to be merged whenever the remaining conflicts has been fixed.

davidlamhauge:
Conflicts has been fixed.

core_lib/src/util/fileformat.h
```diff
@@ -63,6 +63,8 @@ GNU General Public License for more details.
 #define PFF_DEFAULT_PALETTE_EXT \
     QString(".xml")
 
+#define PFF_PALETTE_EXT_FILTER \
```
CandyFace: This define is overriding the one from master and should be removed. It's redundant since the macro is defined above.

davidlamhauge: Removed

Issue #732, opened 2017-08-23
[ISSUE] The SWATCH colors on the COLOR PALETTE keeps changing when selecting another color on the COLOR WHELL


Kaikogf:
### --Issue Summary--
When selecting a default colour from the COLOR PALETTE list, it will be replaced with a diferent color selected from the COLOR WHEEL
[I think I already reported this on another comment but I lost its track]

### --Actual Results
A colour selected on the COLOR PALETTE is changed by the one picket on the COLOR WHEEL
[it makes no sense to have them named as a specific color if they keep changing everytime]

### --Expected Results--
The default list of swatch colours on the COLOR PALETTE should be kept fixed, and any change on the COLOR WHEEL should not affect them

### --Steps to reproduce--
1. Open Pencil2D
2. Notice that the default swatch color already selected is BLACK
3. Go and select another color on the COLOR WHEEL, keeping an eye on the black swatch
4. RESULT: The default swatch's color was changed to the new selected color on the COLOR WHEEL
5. Same happens to any swatch color on the list

### --System Information--
+ Pencil2D Version:
Last build from 23 August 2017

+ Operating System:
Windows 8.1 x 64bits

+ RAM Size:
32GB

+ Graphics Tablet:
HUION GT185 HD

---

Jose-Moreno:
@Kaikogf Hey, glad to see you around. Actually, this is not a bug. This is intended. Just so you know this is how it works on Industry standard software like Toonboom Harmony. On TB however you can "double click" on the palette to bring the color wheel, so it works as a safety feature. But the only way to color on that program is by using the palette first, picking up the colors you will use, and then applying them on the canvas. Sort of like a real painter if you will.

Maybe we can ask for the developers to add a "safety lock" so you can choose not to change the selected palette color. However this is NOT a bug, it is intended that you USE the color palette as the main mean of coloring because it is geared mostly towards an animation production environment where there's a color model sheet you can import for a character or an environment, and you, as the animator are not supposed to change the colours. Normally a color stylist does that for you. You just animate.

On smaller teams you might have to color as the animator, but again this should be dictated by the character designer / color stylist.

For independent animators, I understand that one can have a rather artistic approach, but even though Pencil2D is not a painting app the principle still applies.

Jose-Moreno:
Just to get the point across so others who join the discussion can understand the "why" of this behavior here  I attach a screenshot of toonboom harmony "colour palette" view

[...]

PR #1265, opened 2019-09-16
Update swatches after Replace command

### LINKED ISSUE #1055: Color swatches do not update after using replace command.
### --Issue Summary--
Ok so when using vector layers when you change a color it's value updates but the thumbnail doesn't. se Using the "replace" command updates the swatch color value and thumbnail but not it's name.

However when a bitmap layer is active the palette behaves a bit differently. 

You can change the color locally, and while the value changesthe thumbnail doesn't update, but once you select a different swatch the value of the swatch itself returns to a previous value. Using the replace command does not update the swatch color value, its name, but it does update its thumbnail.

Related to #943 using the replace command on the first swatch of the palette modifies the color but not the thumbnail in vector, but doesn't change anything for bitmap layers.

Summary:
Vector: 
- Replace command updates:
    - [x] Color
    - [x] Thumbnail (Not the first swatch)
    - [ ] Generated Name
Bitmap: 
- Replace command updates:
    - [ ] Color
    - [ ] Thumbnail
    - [ ] Generated Name

Discussion:
It seems not updating the swatch thumbnail when modifying a color with the wheel/ inspector was implemented recently and is intended to avoid unwanted changes to the color, however I've noticed that the vector layer does not respect the "replace" command philosophy as you can permanently change the color without using such command. Ideally this functionality should be consistent on both layer types.

### --Expected Results--
The replace command should properly update all the noted properties and values (the generated name is debatable but still..) and its use should be consistent on either layer type.

Will record a video soon.

### --Steps to reproduce--
Will update later.

### --System Information--
+ Pencil2D Version:
Version: 0.6.1.1
commit: 0a2c87b523fc81c655daf391a26af967134ba733
date: 2018-08-20_11:39:21
Development build
CPU Architecture: x86_64

davidlamhauge:
I am a little confused over this PR.
it closes #1055 , but it only contains one new line of code. I feel I've overlooked something. 
I've tested it several times though, and it seems to work.


davidlamhauge:
I've moved the function so the palette and object are updated almost simultaneously.
I've also added the real-time updating of the swatch, but I too think it should be an act of will, to change the palette. Real-time updating opens up for many accidental changes to the palette, but now we can try it, and discuss it further.

Jose-Moreno:
@scribblemaniac Hmm I think partly due to the discussions that took place in #732 and #755 the result was commit https://github.com/pencil2d/pencil/commit/da990cc5126488ac29d3f971bd05b9164b79fa2e made by chchwy, but that behavior was not ideal, so later on @CandyFace implemented #961 which allowed for both bitmap and vector layers to alter colors in the palette separately.

So after that the real time update of swatches for vector layers was removed because of that, but I agree that it wasn't 100% clear if it was a by-product or an intentional removal :thinking: 

I personally would like that updating the swatch updates it in real time for vector only, however I'm also thinking that considering the replace command was implemented to deal with the color swatch state update, it would not be coherent if we revert back to the real-time update despite being a more intuitive way of previewing the changes when using a vector workflow.

If I were to be totally honest, for me the way it _should_ work is where the Bitmap palette behavior is always **static** and vector is always **dynamic**, so there's contrast between the two workflow "modes"

**How swatches behave In relation to applied colors on canvas**
- **Bitmap** Swatches are **Static** (Changing a color swatch does not update the canvas)
- **Vector** Swatches are **Dynamic** (Changing a color swatch does update the canvas)

**How swatches behave In relation to the Color Inspector / wheel:**
- **Bitmap** Swatches are **Static**
- **Vector** Swatches are **Dynamic** (Read proposal below)*

**Proposal** *: For vector swatches, they could change real-time through the color inspector but **only** when specifically modifying the selected active swatch through the **select color** dialog.

While it would be similar to the previous behavior, in order to give coherence to the replace command and avoiding changing the swatch inadvertently, -_which was the main problem for users before it was excised mainly because people are accustomed to select the color first and then change it but didn't pay attention to the layer they were on_- we could make this real-time update to happen only in this sort of "edit mode" for swatches that is the **Select Color dialog**.

In Harmony for example you can only change the color when you double click the swatch to "edit" and it will bring a panel similar to the color selector we already have in Pencil2D. 

So in Pencil2D If you select the color using the Inspector / Wheel first, you could right click, press replace swatch and all is good. 

But if you select the color swatch and then change the color in the inspector, then double click the swatch (or use a new command called Edit), the current color RGB values would be returned to the selector dialog to avoid retyping / re-selecting the color (this would also work when you select the swatch once, then double click so the starter color would be that of the swatch)

Then using the Select Color dialog slider, the rgb input boxes or if you picked a pre-defined color this would update the color in real time on canvas.

Lastly if you press cancel it would default back to the previous color the swatch had recorded (not the one on the inspector necessarily).

davidlamhauge:
I think I understand what you're writing @Jose-Moreno, but it's not easy for me. As said before, I've never worked with ToonBoom or other software, and I try to make sense in behaviour that I don't understand.
After reading your comment, I still think, that the behaviour that most users would find logical or easy to understand, is that the Bitmap layer is static as it is, but the Vector layer is dynamic in the sense that it updates the scribblearea in real-time, when you pick a color in the Color Wheel or Inspector. When you've found your preferred color, you right-click on the color and press 'Replace'.
It's not 100% real-time, but I think it would be easier to understand for the users.
That said - I'll implement whatever you guys decide. I have no plans on using the Vector layers myself.

davidlamhauge:
So - what can we agree upon in this issue? How and when should vector swatches be updated?

MrStevns:
This has been sitting for too long, so I'll join in. 
Here are my thoughts, after having a discussion with Jose on Discord.

I believe the functionality we want is:
* A vector stroke should not change unless it's been selected or the swatch "replace" command is used.
* The swatch should not change unless the "replace" command is used

Any thoughts?

MrStevns:
After taking a deeper look into this, it's probably out of scope to change the stroke only if selected, because the colors in the palette and the vector strokes are so attached to each other and there's no easy way to change that currently as far as I can see.

My proposal:
We remove the the relation between the the color palette widget and the two other widgets, this means that only the front color will be updated when interacting with the inspector or color wheel. Which will affect the stroke color on the bitmap layer but not on the vector layer.

The swatch will not change unless you use the replace command, which will then affect vector strokes.

With this change, you won't be able to see immediate color changes to a stroke on the vector layer anymore, you'd have to use the replace command specifically for that to happen.

Here's how it would work:
vector strokes

is that what we want?

Jose-Moreno:
> is that what we want?

@MrStevns Sorry!!! I honestly never saw this before until @J5lx pointed out today 🤦‍♂️ AFAIK part of this has already been implemented (the _not_ updating the vector strokes) The behavior you presented is exactly what we ended up discussing after some time, so with the proper updating to the current master I think it can be merged in the way you explained.

MrStevns:
The proposed changes has been added to the PR, it's ready to be reviewed.

J5lx (approved):
LGTM, please merge at your own discretion. Thank you for taking care of this!

Issue #1158, opened 2019-01-18
Implement blending modes [raster + vector]


Jose-Moreno:
### Issue Summary
After using Pencil2D for more illustrative tasks, there is a definite need to have the minimum blending modes for both raster and vector engines.

- For bitmap blending modes we can reference implementations in Krita, GIMP or even MyPaint
- For vector blending modes we can reference Inkscape as they have such implementation for Layer blending modes, which could be adapted to Coloring Tools Blending Modes (this would work on the filled shape entity)

And these capabilities would be further separated in:

+ Layer Blending modes (image compositing blend methods)
+ Coloring tools Blending modes i.e Brush and Fill Bucket

On the subject of coloring blending modes, this is mostly used for effects and enhancing the image (e.g using add mode to create lighting effects). This would be "baked" or applied into canvas image over the colored shapes when painting or applying a fill.

On the contrary on the subject of Layer blending modes, these would exist for a "non-destructive" workflow allowing you to rapidly iterate over a design or to create effects (e.g making and animating a  spotlight which is just a circle mask). 
I've seen Synfig, Blender and other software implement the blending modes / methods that take the alpha from layers and layer "groups" into account, so that using compositing blending modes such as "subtract" (Blender Grease Pencil), "erase" (krita), onto / straight onto (synfig) and so on, can be used as a primitive alpha masking feature similar to what can be found in Flash / Animate CC (nested layers + mask mode), toonboom harmony (masking node) or TVPaint (Stencils, renamed as Masks in TVP11)

PR #1267, opened 2019-09-21
Painting improvements

scribblemaniac:
This is a change I've wanted to do for a while now. I was playing with the feather rendering when I thought to myself, "hey I can probably do this now", and it turns out I could! When using the drawing tools, they don't always look the same as when you finish drawing. The eraser acts like a white brush and all brushes draw over top of everything, even if you are drawing on a lower layer. These were necessary tradeoffs related to how we cached the frame while drawing. However, now I've improved these behaviors and have adapted the caches to work as well as possible (which should be fairly close to the original performance). Take a look:

Live eraser preview revealing the layer below:
Painter

Drawing a stroke on "cur layer" with "pre vec" and "pre bit" below and "post vec" and "post bit" above the active layer:
Ordering

These changes also open up the opportunity to use different blend modes with our brushes. For instance, we could easily add alpha locking and burn/dodge tools with this. See #1158

Will conflict with #1160. The preview while drawing with a tablet will not work, but that will be fixed in #1269 which will be merged before this anyway.


Jose-Moreno:
@scribblemaniac First off, thank you for taking the time to work on this! Although I'm not sure I understand the first GIF, does this mean we can now paint with alpha as well as erase? if we can use fully transparent alpha values we can for example, use the fill tool to color a shape and replace the currently solid color instead of painting over it. Or is that a totally different thing?

If it is a different thing, could that idea be considered a blending mode then? Don't know if you're familiar with Blender's Add Alpha / Erase Alpha blending modes for example, where adding alpha makes the color fully transparent (but somehow it keeps the color instead of erase it) and erasing alpha makes the color fully opaque (modified by pressure of course) so the color is revealed.

Since we're using PNG under the hood and PNG's use straight alpha (that means the color comes RGBA'd instead of premultiplied through transparency masks), we should be able to use that where the color itself is transparent, this of course would count as part of the surface bounding box.

Also is the second GIF what I think it is? could we finally get a way to paint behind the lines? if so that would honestly be a most desired feature!!! and it would also solve a LOT of issues for users have when coloring in the same layer (i.e white halos and incomplete fills due to threshold with anti-aliased strokes)

If we couple this with the upcoming coloring features, Pencil2D users will seriously have no need to use other programs the coloring workflow.

Basically I think this GIF is appropriate for your PR's :grin: 

Daniel_Bryan_YES

scribblemaniac:
> Although I'm not sure I understand the first GIF, does this mean we can now paint with alpha as well as erase? if we can use fully transparent alpha values we can for example, use the fill tool to color a shape and replace the currently solid color instead of painting over it. Or is that a totally different thing?

That's a different thing by the sounds of it, although maybe you could think of the eraser tool as working that way if you wanted. The first gif is of the eraser tool. There's a layer of red, and then a layer with the text below. The red is being erased in real-time, rather than after you finish your stroke. Painting with alpha will add the color on top of the canvas like before. A sort of "clear mode" for the bucket fill was possible before this and is possible after too, it just needs to use a different composition mode.

> could we finally get a way to paint behind the lines?

Again, not really. There are no changes to the tools in this, only in how the are rendered while in use. So you can paint behind lines, but you will need to be on a lower layer, just like right now. The difference is that now you can actually see what you're doing :yum:

The Add/Erase Alpha and drawing behind lines on the same layer should both be possible with different blending modes for the brushes.

I appreciate the enthusiasm :grin:

CandyFace:
It would be nice to have this merged, so we can begin looking at #1160

@scribblemaniac would it be possible for you to fix the remaining conflicts soon?

scribblemaniac:
@CandyFace Yes I'll try to fix the conflicts soon. Just keep bugging me if I don't :wink:

scribblemaniac:
@CandyFace Alright, ready to be merged now. Sorry for the delay.

PR #1269, opened 2019-09-21
Miscellaneous bugfixes

scribblemaniac:
Here are a bunch of bugfixes, almost all for unreported bugs. The commit messages describe what each of the changes are for, feel free to ask for clarification on any of them if needed. I recommend reviewing commit-by-commit as it will make much more sense than looking at all of the changes at once.

### DIFF
--- a/core_lib/src/canvaspainter.cpp
+++ b/core_lib/src/canvaspainter.cpp
@@ -470,7 +470,7 @@ void CanvasPainter::paintCameraBorder(QPainter &painter)
 
         QTransform camTransform = cameraLayer->getViewAtFrame(mFrameNumber);
         mCameraRect = camTransform.inverted().mapRect(mCameraRect);
-        rg2 = camTransform.map(rg2);
+        rg2 = camTransform.inverted().map(rg2);
     }
 
     painter.setOpacity(1.0);
--- a/core_lib/src/canvaspainter.cpp
+++ b/core_lib/src/canvaspainter.cpp
@@ -454,12 +454,14 @@ void CanvasPainter::paintCameraBorder(QPainter &painter)
     QRect boundingRect;
     mCameraRect = cameraLayer->getViewRect();
 
+    QRegion rg2(mCameraRect);
     if (isCameraMode)
     {
         painter.setWorldMatrixEnabled(false);
         QTransform center = QTransform::fromTranslate(viewRect.width() / 2.0, viewRect.height() / 2.0);
         boundingRect = viewRect.toAlignedRect();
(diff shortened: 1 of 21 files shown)


core_lib/src/canvaspainter.cpp
```diff
@@ -454,12 +454,14 @@ void CanvasPainter::paintCameraBorder(QPainter &painter)
     QRect boundingRect;
     mCameraRect = cameraLayer->getViewRect();
 
+    QRegion rg2(mCameraRect);
```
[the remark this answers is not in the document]
scribblemaniac: The camera border has always rotated with the *view* but never with the *camera*. Let me demonstrate.
Before 82e6837:
Peek 2019-09-28 14-27

After 82e6837:
Peek 2019-09-28 14-29

Based on your picture it looks like it might be applying the view transform twice or something like that, which I did encounter in my testing, but I thought I fixed. I will look into it again.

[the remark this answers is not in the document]
scribblemaniac: Alright, it should be fixed now. The zoom now should behave exactly as it did before, except with the rotation fixed like above. Note that the zoom will be affected by both the current view and the camera, I just wasn't inverting the transformation. Here's another gif showing how it should behave when zooming out and rotating on the camera layer and in the view (featuring more quality artwork from yours truly :wink: ):

border_behavior

core_lib/src/interface/editor.cpp
```diff
@@ -929,15 +929,25 @@ KeyFrame* Editor::addKeyFrame(int layerNumber, int frameIndex)
...
-        frameIndex += 1;
+        if (layer->type() == Layer::SOUND && static_cast<SoundClip*>(layer->getKeyFrameAt(frameIndex))->fileName().isEmpty()
+                && layer->removeKeyFrame(frameIndex))
+        {
+            break;
+        }
+        else
+        {
+            frameIndex += 1;
+        }
```
CandyFace: I'm not sure it's a good idea to move to the next available keyframe implicitly, I'm thinking that we already have complains with the scrubber moving on its own.

scribblemaniac: This isn't anything new, it already does this. You obviously can't add a new keyframe where one already exists, so it makes some sense to find the next available space to insert a keyframe. I think people would also complain about the add new frame action doing nothing if we did not do it this way.

CandyFace: True, what I meant is that we had a popup that would appear when trying to create on the same frame for sound layer but I guess this is more consistent..

Issue #1194, opened 2019-03-18
Feather widget slider has different max value than input / spinnerbox


Jose-Moreno:
### Issue Summary

The feather option, widget slider has a different value than the spinner / input box (currently max set at 80).

### Expected Results
Both slider and input / spinner box should have the same value.

Related to #1193 It is important that the maximum value for the feather do not stop the user from applying pixels on a surface. At most it should allow a very faint application of "paint" but this probably is only possible by modifying how the feather calculation is applied.

As per discord discussion it is recommended to cap the feather range from 0 to 100. This will be issued as a separate request as well.

### Steps to reproduce

1. Open Pencil2D
2. Select the brush tool
3. Type the current max feather in the input box (80)
4. Notice the left over on the feather slider widget
5. Move the slider beyond the current position to the right.
6. Notice the input box does not change.

### System Information
+ Pencil2D Version:

Version: 0.6.3
Operating System: Windows 7 SP 1 (6.1)
CPU Architecture: x86_64

scribblemaniac:
Due to the long time with no activity, I am unassigning you from this issue @ElectricVersion. You are still welcome to submit a PR for this.

davidlamhauge:
I have a PR almost ready for issue #1193, and that PR covers this issue as well. For this reason I assign this to myself.

CandyFace:
The PR resolving this issue has been merged, closing.

PR #1274, opened 2019-10-01
#1193 Changing feather range to 1-99

### LINKED ISSUE #1193: High feathering values do not allow pixel application on canvas
### Issue Summary

Any feathering value above 96 (tested with Qt by modifying ui files) will clamp the feather and not draw any pixels at all. This affects particularly the Brush Tool as it is the only active tool that paints with variable feathering but this is a global issue, so the Pencil tool, the Smudge tool or any other tool that would be set to sue feathering would suffer from this issue.

### Expected Results

I mentioned before a value fo 200 might be good, but what I meant was that a value that amtched the maximum SIZE value, would work, however it seems the calculations say otherwise.

On Discord it was suggested to clamp the feathering values from 0 to 100, but the maximum feathering value should never stop the user from painting, at most it will give a very faint and smoothed out paint feel.

### Steps to reproduce

1. Open Pencil2D
2. Select the brush tool and move the slider to the right most edge. The input box will say 80, but this is another bug reported here: #1194
3. Try to draw anything.
4. Move the slider to the left little by little until you can draw and compare.

### System Information
+ Pencil2D Version:
Version: 0.6.3
CPU Architecture: x86_64

davidlamhauge:
This closes #1193 and #1194 .
Since the value indicates the alpha-value of the feathering, 100 = invisible. For this reason it could be argued that we should have 99 as maximum. Right now it is set to 100.

### DIFF
--- a/app/src/tooloptionwidget.cpp
+++ b/app/src/tooloptionwidget.cpp
@@ -49,7 +49,7 @@ void ToolOptionWidget::initUI()
     ui->sizeSlider->setValue(settings.value("brushWidth", "3").toDouble());
     ui->brushSpinBox->setValue(settings.value("brushWidth", "3").toDouble());
 
-    ui->featherSlider->init(tr("Feather"), SpinSlider::LOG, SpinSlider::INTEGER, 1, 100);
+    ui->featherSlider->init(tr("Feather"), SpinSlider::LOG, SpinSlider::INTEGER, 1, 99);
     ui->featherSlider->setValue(settings.value("brushFeather", "5").toDouble());
     ui->featherSpinBox->setValue(settings.value("brushFeather", "5").toDouble());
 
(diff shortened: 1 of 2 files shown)


CandyFace:
I'd say that it should be 1-99, the brush shouldn't be able to go invisible since the user could consider that a bug.

Jose-Moreno:
I agree with CandyFace, and it's actually written in the expected results for 1193. Feathering should never impede the user from painting because it's only meant to lighten the amount of color that is applied on the canvas so strokes look softer.

davidlamhauge:
I agree, but as I remember the discussion we had, it was undecided, so I had to choose something. I'll change it later today.

davidlamhauge:
Maximum feather set to 99.

PR #1276, opened 2019-10-01
#177 Save and Load shortcut list

### LINKED ISSUE #177: A way to save the Custom shortcuts list for future needs.
This would be an enhancement.
I was wondering if there a simple way to save our preference shortcut´s list?
Every time there is a new build, the shortcut list is showed as a default. So it would be a saving time, just that, for re-loading our preferences

Thank you all developers. You are kind of Inspiration for all of us. Really!
So far Pencil2D is a very nice software for making our life more interesting :) with its animations :)

davidlamhauge:
This PR addresses a user request from 2013.
The more you use Pencil2D, the more you know what you need from shortcuts. It is important that you can save your favourite shortcuts, so they can easily be re-instated, if they are messed up, or someone accidentally pressed 'Reset to default'.
The shortcuts are save as a text-file, and I've given them the suffix '*.p2shortcut' . It could be altered if you want to, but I prefer special and descriptive suffixes for special files.
This closes #177 .


app/src/shortcutspage.cpp
```diff
@@ -117,6 +121,56 @@ void ShortcutsPage::keyCapLineEditTextChanged()
     qDebug() << "Shortcut " << strCmdName << " = " << strKeySeq;
 }
 
+void ShortcutsPage::saveShortcutsButtonClicked()
+{
+    QSettings settings( PENCIL2D, PENCIL2D );
+    QString fDir = settings.value("LastSavePath/Animation").toString();
```
CandyFace: As far as I can see you never set this anywhere so the string will always be empty.

scribblemaniac: It's set here I think: https://github.com/pencil2d/pencil/blob/220a27a06756bc524230211c7b80a9183eac6d23/app/src/filedialogex.cpp#L123-L125

Having said that, it definitely could still be empty, which should be checked so that it doesn't try to write to the root directory.

CandyFace: Hmm that looks like a different value to me, I'd expect that what you posted should equal a string "Animation" not LastSavePath/Animation

scribblemaniac: As I understand it, if you think of the LastSavePath/Animation, then beginGroup is like cd (or more accurately pushd).

Looks like it's set alright for me. if you're looking directly at the settings file, it will look like this:
```
[LastSavePath]
Animation=<some_path>
```

Take this with a grain of salt as I have not actually tested this branch yet.

Jose-Moreno:
Great work. I understand the need for descriptive names, but If possible let's try not to exceed the 3~6 character range for filename extensions. I know it's not 1980 but it's better to make it short for mnemonics and simplicity. We already have *.vec, *.pcl, *.xml, etc so hopefully we can continue being coherent. Nevertheless here are some varied proposals: 

- *.key
- *.ksc (**k**eyboard **s**hort**c**ut)
- *.pclkey
- *.kbdmap
- *.keymap

davidlamhauge:
I have tested it thoroughly, and it finds the last animation path as expected.
I can see that I forgot these lines
`if (fDir.isEmpty())
    fDir = QDir::homePath();`
in the loadShortcutsButtonClicked, so the path can be empty if nothing has been saved before. I'll correct that later today.
About the name, the shortest I can think of is *.pclscut, but for the sake of the user I still would prefer *.pclshortcut. I have no issues with long suffixes. Especially not when they are helpful.
I just remebered one odd thing: I've tested it on MacBook, win10 and Ubuntu18.04, and works as expected, except when I want to load a shortcut file in Ubuntu. Here it finds the last saved animation path, but the filter is not working. It shows all files, and the button that should show the filter is empty. When I doubleclick this button, the filter starts working, and everything is ok again. Can any of you tell me why it's like that? It's work fine in iOS and Win.

chchwy:
The file is actually in [INI format](https://en.wikipedia.org/wiki/INI_file), so call it *.ini is legit, too.

davidlamhauge:
Yes, it is the INI format, but it's content is not a typical ini-file, so I would not support that. If the others support it, iI will of course follow the majority.
This reminds me of the palette files that have the *.xml suffix. Even if the content is xml, I would never have chosen that name. I would go with *.pclpalet or something. When I export a palette file, I start the filename with 'PAL', as in PALgeorge.xml, so I can see that it is a palette file. There are so many xml-files around that I have to distinguish it somehow.

app/src/shortcutspage.cpp
CandyFace: It works as expected from loadShortcutsButtonClicked but in saveShortcutsButtonClicked I never get a path. I've also debugged it a bit because the code is slightly different.

this in the load method will work:
```
    QSettings settings( PENCIL2D, PENCIL2D );
    QString fDir = settings.value("LastSavePath/Animation").toString();
```

While this in the save method will not.
```
settings.beginGroup( "LastSavePath" );
    QString fDir = settings.value("LastSavePath/Animation").toString();
```

If you however change the above to 
```
settings.beginGroup( "LastSavePath" );
QString fDir = settings.value("Animation").toString();
```

Then I get the expected behaviour.

davidlamhauge: Maybe that's what is not working in Ubuntu, but is working on Win and iOS. As explained in another comment, I have som odd behavior on my Ubuntu. I'll test it when I get home.

davidlamhauge:
With some help I go the path problem solved. Everything works as expected in my tests now.
Since I was making a commit anyway, I changed the suffix to *.pclshortcut. I still think a suffix to such a file should be descriptive, but I'll change it again if necessary, to whatever we can agree upon.

Jose-Moreno:
I think the same as @CandyFace that it should be related to the software using pre-existing format keywords . Although I would certainly like it to be shorter, if it's not possible or usable then just go with `*.pclkey[s]` (plural might work better) or for more descriptive power try with `*.pclkeylist` or `*.pclkeymap` Whatever works is fine with me. Good work and thank you for doing this @davidlamhauge ! :blush:

davidlamhauge:
I can't understand why *.pclkeymap, *.pclkeys or *.pclkeylist is better than *.pclshortcut, when it comes to a file that contains nothing but information on shortcuts.
Of the three suggestions, I prefer *.pclkeys, but I'm glad that I know it's a shortcut file, because I would never have guessed it by the name.
Can we agree on *.pclkeys ?

Jose-Moreno:
Sorry, I just wanted to understand and suggest an appropriate name, so let me analyze this last piece here then. 

Is `shortcut` more universal than `keymap`? is it more common? We do use "keyboard shortcuts", keymap is perhaps more technical jargon. Key can be confused with keyframe, so there's that.

In that sense I honestly don't really mind if it's renamed as `*.pclshortcuts` , but I must also apologize for the misunderstanding, I was working under the assumption that it was still proposed to be `p2shortcut` instead :man_facepalming: :bowing_man: 

Either way I see there is no possibility we can have both a descriptive name and short name for the extension, so, let's go with the _shortcuts_ one and let this PR be finalized under the assumption that:

a. This is a file type that will be directly manipulated by the end user and read by Pencil2D
b. The extension is honestly only for cosmetic purposes
c. We are assuming the user has filename visibility enabled in their operating system (which most people don't)
d. In a common example, I assume everyone will rename the shortcut list to something along the lines of "_Pencil2DShortcutList_Animation.pclshortcuts_" but most people won't see the extension. and those who do will adjust the name accordingly to avoid longer names. As I can see filenames, I'd rename them like "_Drawing.pclshortcuts_" or "_Coloring.pclshortcuts_" if that were the case

scribblemaniac:
The code looks good aside from what I have already noted/fixed. I don't like how long `.pclshortcuts` is but I like shortcuts over keymap. I personally would have gone with `.pcls` for shortcuts and `.pclp` for palettes, but I don't care too much. Whatever we settle on, it would be a good idea to add it into to the mime type lists (ex. info.plist, desktop file), but only after we support opening these files from the program arguments.

davidlamhauge:
I haven't tested it @scribblemaniac , but it looks good. What now? I can't see that I can merge it, since it's not a PR.
About the suffix, I think *.pcls is a good alternative to the *.pclkey*** combinations, which I honestly don't like. What do you say @Jose-Moreno ?
Next thing is to alter palette suffixes from *.xml to *.pclp, but I guess that's too late...

scribblemaniac:
> I haven't tested it @scribblemaniac , but it looks good. What now? I can't see that I can merge it, since it's not a PR.

You don't have to do anything if you're good with it since it was already push to your repository. You'll want to run `git pull` on your local repository to get the changes if you haven't already of course. Please do test it when you get a chance, I'm only able to test it on linux right now, and as you know things don't always work the same on other systems.

> Next thing is to alter palette suffixes from *.xml to *.pclp, but I guess that's too late...

We could still change it. We would still have to allow opening .xml files but we could write all new files with a different extension at least.

Jose-Moreno:
I don't mind `*.pcls` +1 :+1: 

Regarding `*.xml` that is a bit more trickier for me, I feel this format, and others similar to it, should be used as a data exchange format. 

I've always entertained the idea of allowing software like Harmony, Synfig, GIMP, Krita, etc to import the Pencil2D palette back and forth. Changing this file format name to something specific might confuse other people trying to get the information out of pencil2d.

That said, if we leave the capability of having both, the `*.pclp` for internal purposes and the `*.xml` for exchange purposes, with a more universal node structure, I'm fine with that too.

Don't know what the others think about the latter idea, but if you guys want to change the palette format either way you have my blessing :yum:

davidlamhauge:
I've tested it @scribblemaniac , and it works fine, except line 152 and 154. With these lines in the code, the odd behaviour explained 4 days ago, and solved by @CandyFace , comes back, so I've deleted these lines. It's tested on Ubuntu, so I wonder why didn't have the same odd behaviour?
About the *.xml for palettes, I think we should discuss it elsewhere, but I think it should be discussed.

scribblemaniac:
> Regarding *.xml that is a bit more trickier for me, I feel this format, and others similar to it, should be used as a data exchange format.

I think basically the opposite, that it should primarily be a internal format and a more widely recognized format like the GIMP palette format should be used for interchange. Anyway, as @davidlamhauge said, this can be discussed elsewhere.

> I've tested it @scribblemaniac , and it works fine, except line 152 and 154. With these lines in the code, the odd behaviour explained 4 days ago, and solved by @CandyFace , comes back, so I've deleted these lines.

Huh? Line 152 was added by @CandyFace to fix the issue in the first place, and 154 is after the setting is retrieved so it should not affect that at all.

davidlamhauge:
@scribblemaniac  It still yielded some strange behaviour on my Ubuntu, so now I've done what I probably should have done from the beginning. I've added Shortcuts to LastSavePath in Preferences.

CandyFace:
I think that *.pcls is too close to the project file format, even if it's for internal use only I still think the extension should be somewhat descriptive and separated from other potential name clashes. I'd rather have *.pclkmap, *.pclkeymap or *.pclkeys

davidlamhauge:
My favourite office suite, LibreOffice, uses the odf format. Writer documents are called *.odt, Spreadsheets are called *.ods and Presentations are called *.odp. I have never heard anyone complain about mixing them up. 
Word has used *.doc for many years. Now they use *.docx, and *.dotx if it's a template, without problems afaik.
As mentioned before, I like the idea of having longer and more descriptive extensions, but *.pclkmap, *.pclkeymap and *.pclkeys confuse me more than the enligth me.
*.pclshortcut is my preferred, if we can't agree on *.pcls.

PR #1283, opened 2019-10-02
#944 Reorder palette swatches with drag and drop

### LINKED ISSUE #944: [Request] Allow to re-order color swatches in either list or grid mode
### --Issue Summary--
With the idea of customizing swatches to export them later and be re-used by other project members in any production, this is another important behavior for creating custom palettes from your color model.

Ideally you should just be able to click select & drag any swatch to change it's default order.

If the color palette has some sort of absolute index (similar to what layers had) it might be wise to think ahead and also re-issue the index once the move / re-order operation is completed.

davidlamhauge:
A much needed feature, is the ability to drag and drop palette swatches.
You can select as many swatches as you want. For every drag and drop, the links to possible vector strokes are updated and re-indexed.
Tested on Ubuntu 18.04, Win10 and Mac.
This closes #944


app/src/colorpalettewidget.cpp
```diff
@@ -262,6 +264,54 @@ void ColorPaletteWidget::onItemChanged(QListWidgetItem* item)
...
+void ColorPaletteWidget::onRowsMoved(const QModelIndex &parent, int start, int end, const QModelIndex &destination, int row)
+{
+    Q_UNUSED(parent)
+    Q_UNUSED(destination)
+    Q_UNUSED(end)
+
+    int startIndex, endIndex;
+    if (start < row)
+    {
+        row -= 1; // TODO: Is this a bug?
```
CandyFace: Not sure... all documentation points towards row and column being zero based (at least for QModelIndex) so in theory when selecting the first item, the row should be 0 but the actual result suggests that it starts at 1.

row in this context might not be the same as row in QModelIndex context but there seems to be no info that should suggests otherwise.

core_lib/src/structure/layervector.cpp
```diff
@@ -50,6 +50,14 @@ void LayerVector::removeColour(int colorIndex)
     });
 }
 
+void LayerVector::moveColor(int start, int end)
+{
+    foreachKeyFrame( [=]( KeyFrame* pKeyFrame )
```
CandyFace:
```suggestion
    foreachKeyFrame( [=] (KeyFrame* pKeyFrame)
```

core_lib/src/structure/layervector.cpp
```diff
@@ -50,6 +50,14 @@ void LayerVector::removeColour(int colorIndex)
...
+    {
+        auto pVecImage = static_cast< VectorImage* >( pKeyFrame );
```
CandyFace:
```suggestion
        auto pVecImage = static_cast<VectorImage*>(pKeyFrame);
```

core_lib/src/structure/layervector.cpp
```diff
@@ -50,6 +50,14 @@ void LayerVector::removeColour(int colorIndex)
...
+        pVecImage->moveColor(start, end );
```
CandyFace:
```suggestion
        pVecImage->moveColor(start, end);
```

core_lib/src/structure/object.cpp
```diff
@@ -308,6 +308,25 @@ void Object::addColour(QColor colour)
...
 }
 
+void Object::movePaletteColor(int start, int end)
+{
+    mPalette.move(start, end);
+}
+
+void Object::moveVectorColor(int start, int end)
+{
+    // movePaletteColor _must_ be made before calling this function!
```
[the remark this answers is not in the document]
davidlamhauge: That comment was more like a reminder to myself, and should have been erased before commit.
The 'logic' in ColorPaletteWidget::onRowsMoved won't work unless you first call movePaletColor, and then some other bogus - but it works!

app/src/colorpalettewidget.cpp
davidlamhauge: I just know that that the default palette has 24 colors, with indexes 0-23. If I drag the first color, and places it in the end, then the 'row' parameter is 24(?). It works if i deduct 1 from row here...

CandyFace: Yeah I think it's fine, seems to work as expected from what I've tested.

davidlamhauge:
I have made the proposed changes - I think.
Does drag and drop work in grid mode? I haven't been able to make it work, If it's a must, then I'll try again when I have better time.
I had a function that disabled drag 'n drop if multiple swatches were selected, but I disabled it here 
c69ac35 and re-enabled it here https://github.com/pencil2d/pencil/pull/1283/commits/8614e0fc01cc1a38b2e6ff42067bf97b048afc6c because I tested it with multiple selection, and it worked. It doesn't seem to work any more, so maybe I should reinstate this function? (Annoying...)

app/src/colorpalettewidget.cpp
```diff
@@ -262,6 +264,54 @@ void ColorPaletteWidget::onItemChanged(QListWidgetItem* item)
...
+        editor()->object()->addColour(editor()->object()->getColour(startIndex));
+        editor()->object()->moveVectorColor(startIndex, editor()->object()->getColourCount() - 1);
+        for (int i = startIndex; i > endIndex; i--)
+        {
+            editor()->object()->moveVectorColor(i - 1, i);
+        }
+        editor()->object()->moveVectorColor(editor()->object()->getColourCount() - 1, endIndex);
+    }
+
+    editor()->object()->removeColour(editor()->object()->getColourCount() - 1);
```
CandyFace: You call editor()->object a lot, consider creating a local Object* variable so that repeated reference lookup is avoided.

davidlamhauge: For some odd reason I had to make both Editor and Object local vars.
When I tried
`mObject = editor()->object();`
it crashed. It works fine now, with the usual setCore method.

davidlamhauge:
More issues/comment to this one?

CandyFace:
Merging this one, knowing that the functionality doesn't work for grid view.

Issue #1177, opened 2019-02-12
Imported image position modified by canvas view center


Jose-Moreno:
### Issue Summary

The problem here is that when importing any image, these images are not positioned relative to the canvas absolute center, but relative to the view center, so if you pan outside the camera viewfinder the image will be placed exactly in the center of the view on canvas, but outside the initial camera view state, which for all intents and purposes represents the actual canvas that will be seen (unless there is camera motion).

I'd like to propose that the images are placed based on a selected camera layer viewspace (remember we can have more than 1 camera) to import these drawings.

This option should require enhancing the image import dialogs to allow the canvas placement of any given image be relative to any selected camera layer, provided we have more than one. This also to maintain conceptual consistency and flexibility in the same way image and movie export dialogs are currently based on the camera finder (with camera selector and all). 

This is not really a critical bug, but more like something that can throw people off if it's not understood. The current workaround is simply to reset the view every time you want to import a still image or image sequence.

### Expected Results

When having only one camera, the image should be appropriately placed in the middle of and relative to the active camera view. If the camera is hidden, then place the image in the absolute center of the canvas.

If there is more than one camera, allow the user to select which camera should be used to base the image location. In a sense the camera becomes a parent of the image only for import purposes.

### Steps to reproduce

1. Create a new pencil2d file
2. Import any image on the default state for comparison. For my tests I used TIFF image files.
3. Pan the view away from the current image
4. Import another image, can be the same as used before. Note the difference in placement as the previous one.

### System Information
+ Pencil2D Version:

Version: 0.6.2
commit: fe450b794aca754768f9e5dfe02e6abcdcc29189
date: 2019-02-05_13:58:04
Development build
Operating System: Windows 7 SP 1 (6.1)
CPU Architecture: x86_64

davidlamhauge:
@Jose-Moreno  I think it is confusing if you can "...allow the user to select which camera should be used to base the image location."
The images should at all times be imported to the absolute center of the canvas. 
We should on the other hand give the user the possibility to move selected drawings to another position, whether or not they are imported or drawn in Pencil2D.

Jose-Moreno:
@davidlamhauge I understand what you mean, and I agree that the import should be done relative to the camera, but bear in mind that you can have multiple cameras in a single pencil2d document. So that particular suggestion was for the possibility that the active camera (or one that the user actively choose) will import the image relative to the center, and the active camera may not necessarily be the one pointing to the canvas center.

[...]

PR #1285, opened 2019-10-04
#1177 Import drawings on wanted position

davidlamhauge:
This addresses #1177, but I won't say it closes it.
There are 3 kinds oh image imports:
(1) image - (2) image sequence - (3) predefined set.
With this PR, you can select whether you you want to import (2) and (3) relative to: 
(a) current view - (b) canvas centre - (c) current camera view. 
The selection is done in a combobox on the filedialog.
You can't select where you want to import a single image (1), because we use the ordinary fileopen dialog in that case, and it's not possible to (for me at least) to place a combobox on that dialog. It could be solved by agreeing that all single image imports are done to canvas centre. Or - what do you think?
You can't select camera either, if you have multiple cameras. The importExportDialog doesn't know of layers or editors, and I'm not convinced that this possibility is necessary.

### DIFF
--- a/app/app.pro
+++ b/app/app.pro
@@ -35,6 +35,7 @@ INCLUDEPATH += \
 
 HEADERS += \
     src/importlayersdialog.h \
+    src/importpositiondialog.h \
     src/mainwindow2.h \
     src/predefinedsetmodel.h \
     src/pegbaralignmentdialog.h \
@@ -67,6 +68,7 @@ HEADERS += \
 
 SOURCES += \
     src/importlayersdialog.cpp \
+    src/importpositiondialog.cpp \
     src/main.cpp \
     src/mainwindow2.cpp \
     src/predefinedsetmodel.cpp \
@@ -100,6 +102,7 @@ SOURCES += \
 FORMS += \
     ui/importimageseqpreview.ui \
     ui/importlayersdialog.ui \
+    ui/importpositiondialog.ui \
     ui/mainwindow2.ui \
     ui/pegbaralignmentdialog.ui \
     ui/timeline2.ui \
--- a/app/src/importexportdialog.cpp
+++ b/app/src/importexportdialog.cpp
@@ -28,8 +28,13 @@ ImportExportDialog::ImportExportDialog(QWidget* parent, Mode eMode, FileType eFi
     ui = new Ui::ImportExportDialog;
     ui->setupUi(this);
     m_fileDialog = new FileDialog(this);
+    ui->cbImportPosition->addItem("Current view and position");
+    ui->cbImportPosition->addItem("Canvas center");
+    ui->cbImportPosition->addItem("Active camera");
 
     connect(ui->browseButton, &QPushButton::clicked, this, &ImportExportDialog::browse);
+    connect(ui->cbImportPosition, QOverload<int>::of(&QComboBox::currentIndexChanged),
+            this, &ImportExportDialog::setPosIndex);
 
     Qt::WindowFlags eFlags = Qt::Dialog | Qt::WindowTitleHint | Qt::WindowCloseButtonHint;
     setWindowFlags(eFlags);
--- a/app/src/importexportdialog.cpp
+++ b/app/src/importexportdialog.cpp
@@ -28,13 +28,7 @@ ImportExportDialog::ImportExportDialog(QWidget* parent, Mode eMode, FileType eFi
     ui = new Ui::ImportExportDialog;
     ui->setupUi(this);
     m_fileDialog = new FileDialog(this);
-    ui->cbImportPosition->addItem("Current view and position");
-    ui->cbImportPosition->addItem("Canvas center");
-    ui->cbImportPosition->addItem("Active camera");
-
     connect(ui->browseButton, &QPushButton::clicked, this, &ImportExportDialog::browse);
-    connect(ui->cbImportPosition, QOverload<int>::of(&QComboBox::currentIndexChanged),
-            this, &ImportExportDialog::setPosIndex);
 
     Qt::WindowFlags eFlags = Qt::Dialog | Qt::WindowTitleHint | Qt::WindowCloseButtonHint;
     setWindowFlags(eFlags);
--- a/app/src/importexportdialog.h
+++ b/app/src/importexportdialog.h
@@ -42,6 +42,7 @@ public:
     QString getFilePath() const;
     QString getAbsolutePath();
     QStringList getFilePaths();
+    int getPosIndex() { return mPosIndex; }
 
 signals:
     void filePathsChanged(QStringList filePaths);
@@ -60,6 +61,7 @@ protected:
 
 private slots:
     void browse();
+    void setPosIndex(int index) { mPosIndex = index; }
 
 private:
     Ui::ImportExportDialog* ui = nullptr;
@@ -69,6 +71,7 @@ private:
 
     FileType mFileType = FileType::ANIMATION;
     Mode mMode = Import;
+    int mPosIndex = 0;
 };
 
 #endif // IMPORTEXPORTDIALOG_H
--- a/app/src/importimageseqdialog.cpp
+++ b/app/src/importimageseqdialog.cpp
@@ -158,8 +158,18 @@ void ImportImageSeqDialog::setSpace(int number)
     uiOptionsBox->spaceSpinBox->setValue(number);
 }
 
-void ImportImageSeqDialog::importArbitrarySequence()
+void ImportImageSeqDialog::importArbitrarySequence(int index)
 {
+    QPointF currentView = mEditor->view()->translation();
+    if (index == 1)
+    {
+        mEditor->view()->resetView();
+    }
+    else if (index == 2)
+    {
+        mEditor->view()->translate(mEditor->view()->getCameraView());
+    }
+
     QStringList files = getFilePaths();
     int number = getSpace();
 
@@ -219,6 +229,8 @@ void ImportImageSeqDialog::importArbitrarySequence()
                              QMessageBox::Ok,
                              QMessageBox::Ok);
     }
+
+    mEditor->view()->translate(currentView);
 
     emit notifyAnimationLengthChanged();
     progress.close();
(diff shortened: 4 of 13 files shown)


CandyFace:
While I think that adding a combobox to each different import dialog is an okay solution that would work for some cases, it's limited because it's tied to a dialog and won't work for normal image import.

My own take on this would have been to import the images as you normally would and then add a post process step afterwards where you can drag the imported image around on the canvas, so you're 100% in control of where you want to import the image/s. There could still be a floating window with a combobox of where you want to position for some easy adjustments but overall I think that the ability to position exactly where you want the image is crucial.

This would not be tied to a dialog and the same logic could work for importing a sequence of images too, because you would use the first image as pinpoint of where you want the other images.

My suggestion to make this PR work for all imports (disregarding my specific position comment above), would be to trigger another custom dialog specific for re-positioning after clicking "ok"/open on the import dialog, where you can select the combobox options you show in this PR.

This is better because:
1. There's going to be less repetitive code
2. It can be added to any future dialog, no matter what it looks like
3. It will be easier to maintain.

davidlamhauge:
I had that in mind @CandyFace , but I thought it was somewhat clumsy. For the reasons you give, I can see it's in fact a better solution.
Your comments on a post process where you could reposition the drawings, reminds me of a feature I worked on half a year ago or something. I couldn't get it to work then, but maybe I got wiser since...? Not only for importing, but also for re-using animation, it's important that you can select a series of drawings and re-position them. That will be another PR.

davidlamhauge:
With good help from @CandyFace , I've made a new commit to this requested feature. I hope it fulfills the purpose.
You can now import images, image sequences, predefined image sets and gifs, relative to your view position or the camera view. Please feel free to test it.

chchwy:
Better to add to the milestone whenever merging a PR, it would be a lot of easier for us in preparing the next release note.

CandyFace:
Yes of course, forgot about that Matt 👍

scribblemaniac:
I know I'm very late to the party here. I don't think that making another dialog was the best solution. It's quite annoying to have to go through 3 dialogs just to import an image sequence. Code duplication is a valid concern, but I think a more appropriate solution would have been to make a custom widget which could be added to each import dialog.

What happened to the center of currernt view option? I think that could be useful for situations where you want to import to a specific position, but don't want to modify the camera to do so.

Finally, importing with the "Center of active camera" option selected imports the images to the center of the camera at the first frame, but I would expect it to be centered on the camera at the time it is imported. Here's a demo of what I mean, because it is a bit difficult to describe:
Peek 2019-11-11 11-36
I would expect the imported images to remain in the center of the camera as it moves.

CandyFace:
I agree that we could have make the flow less tedious by for example having a checkbox to avoid being asked every time, however I don't think having another widget in every import dialog is a better idea.

Ideally I'd want the import to position to be much more visualized and not be bound to a dialog with the expectancy that one understand what either "center of canvas" or "center of active camera" means, so you would be able to freely move the image to where you want the image to be, with the addition of a floating widget that could be used to move the image to defined positions like what it currently does.

As for the discovery you made, that's a bug it seems, I never considered testing an import where a camera transition ongoing :P

scribblemaniac:
> I agree that we could have make the flow less tedious by for example having a checkbox to avoid being asked every time, however I don't think having another widget in every import dialog is a better idea.

To me that is saying that either you have to deal with the second dialog each time, or you you have to go to the preferences or wherever each time you want to change it. I don't know how often users will want to change this setting. If they want to change it often, it's better off in the same dialog as other options they might want to change (ex. frame spacing). And if they don't want to change it often, then it's better off being just a preference from the start. I think we should figure out what use cases are likely and choose the most appropriate option for it.

> Ideally I'd want the import to position to be much more visualized and not be bound to a dialog with the expectancy that one understand what either "center of canvas" or "center of active camera" means, so you would be able to freely move the image to where you want the image to be, with the addition of a floating widget that could be used to move the image to defined positions like what it currently does.

Yes I do like this idea. Or we could allow it to be moved after importing similar to what's been proposed for a single frame except that moving the one frame affects all of them.

CandyFace:
Just a reminder: what was merged is not the final solution but it's one iteration that's usable and better than what was currently implemented. There's definitely room for improvement though.

I don't think options regarding import/export should be hidden in preferences, there are too many steps back and forth, it should definitely always be available, either through another widget in the import dialog or another dialog that shows later.

More buttons/settings in a dialog vs spreading it out through multiple tabs, windows or whatever is UI/UX and I'd say that it's better and easier to understand things where there are less of them, even if it requires more steps, however that's very subjective...

However none of these things matter if we rework the solution and allow moving the image/s in realtime :P

Jose-Moreno:
@scribblemaniac @CandyFace It's not a bug it's a feature! :laughing: That's actually pretty cool, to import the files while the camera is moving! This can be used when you are animating a moving shot so you could load your layout or storyboard panels in order to roughly give it a sense of space, similar to this video here (notice the actual camera frames that are drawn have the effect of moving spatially, but all of it is 2D). 

Watch the video by scrubbing frame by frame and you'll see what I mean regarding the camera framing:
https://www.sakugabooru.com/post/show/17755

I won't be opposed it it's "fixed", but just consider it :wink: 

Regarding the center of the canvas...I think we should at least consider the default canvas position with the camera frame middle point as the relative center of the "world".

image

In that sense the camera would then be moving in relation to that. This would be aligned with how actual traditional animators used to use the field chart, where the center of the screen was (0,0) and they moved North (-Y in CG) South(+Y) East (+X) & West (-X) to specific coordinates.
image

However related to this I recall you guys mentioned that the camera wasn't actually moving, but the canvas was. I guess this was meant originally to simulate a camera stand where the backgrounds were meant to be moved instead while taking the pictures (a la Disney), so if that's true, then it would still make sense that the camera was stationary while we better represent that what you're actually moving is the canvas itself? But if this is the case, then the center of the "world" should related to the camera, and not to the canvas :thinking: 

What do you think?

davidlamhauge:
I agree with @Jose-Moreno that it's not a bug but a feature. If you @scribblemaniac  want your drawings to follow the camera movement, that should be another option in the menu/dropdown.
@CandyFace asks "what is 0,0 in an infinite canvas?". I think Jose explains and shows it in his comment. For 20+ years that fieldguide was good help for me in my professional career, and it's natural for an animator to think of 0,0 as the center.
I also agree that there is room for improvement. I worked on a feature some time ago, where you could move a series of drawings to wherever you wanted it. I couldn't make it work then, but maybe we should discuss (under another issue) how it should be implemented, the design of the dialog etc, so it could have another go?

CandyFace:
@davidlamhauge, @Jose-Moreno I think you misunderstand what I mean

You have center of active camera 
image


Then you have center of viewport (what I consider canvas... but viewport is not a layman thing)
image


This is fixed to the coordinate system of the application window

Then there's the area inside the unlimited drawing area, this has no height or width, therefore no visual border, therefore no center.

What my point is that we could rephrase the options to something like:
* Centre of active camera
* Centre of canvas (some arbitrary point)
* Centre of viewport (what I consider canvas) 

if you guys think that's better.

davidlamhauge:
That's not how I see it @CandyFace .
The coordinate system inside the unlimited drawing area has indeed no height, width or visual border, but is has a center, and this center i (0,0). 
I rely on it, @Jose-Moreno relies on it, and my old colleague Jakob Koch relies on it. He has told how he sometimes must re-import drawings, because they were not 'centered', and by centered he means according to (0,0).
Even the camera has a function called 'reset', that translates the camera to (0,0).
For me, and others, (0,0) is the absolute center of the infinite canvas, and thus a very important point.

Jose-Moreno:
@CandyFace Thank you for taking the time to clear up the definitions, I didn't know we had to consider the viewport as well like that.

As of 0.6.4 the images are being imported at the center of the "viewport". Unfortunately this is not consistent with how people experience import in other programs, and the reason behind # 1177 

So the proposal was to choose between: 
- Active camera center: The averaged center of the camera point you obtain at (width / 2, height /2)
- Paper canvas center: The "arbitrary" point on the surface that both the camera and the viewport are on top of when the zoom / rotate operator is used, or when a new file is created.

Having the center point either in the paper canvas or in the active camera center would initially fulfill the same role, to give a sense of virtual space for the artist.

I always think of it as If I had a drawing on my table and looked at it through a lens. I can move freely but the drawing will remain in the "same" place. The user is what moves virtually.

If I move the view I expect the images to be imported at the center of the camera / paper canvas always, because that represents my drawing area.

The caveat with having the origin at the center of the camera though, seems to be that if you "move" the camera you will create the nice effect Scribblemaniac showed. Moving the view should not change this though.

If we left the origin in the "paper" canvas however, moving the view or the camera should not effect the import location whatsoever. I can live with having the center origin at the center of the camera, but the "paper canvas" option seems to offer the most consistent behavior for me.

chchwy:
There is a center (0, 0) in the infinite canvas actually. It's the center of the world space.

The top-left point of a bitmap image sits in the world space. The image was transformed to camera space (with the camera position/rotation). and then was transformed to screen space (with the viewport) and finally be drawn on the window.

CandyFace:
@chchwy True, From a programming or matrix perspective there's a center but that's a technical knowledge that the user can't visually see.

Consider this example: You've created a preset or startup project where your camera always starts 1000 pixels to the left in the world space, would you still consider the point that is 1000 pixels (eg. the actual world space center) to the right to be the center and thus where to import the images, rather than where you currently are?

To me I'd expect the starting pos to be the center of canvas, since that's where I start drawing every time but maybe that's just me.. I still see center as an arbitrary coordinate in world space, even if that's technically 0,0 since there are no boundaries.

Whatever the case, let's just add the "0,0" as center of canvas again, your points are valid.

scribblemaniac:
The center of the canvas, from a user perspective can be viewed as the center of the viewport when the view is reset. Although I agree that it is quite arbitrary.

Anyway, I think all of this is missing the point I was trying to make. I'm not saying that having an import that is fixed relative to the canvas is a bad idea, but it absolutely should not be called "Center of active camera". That implies that well...the pictures will be in the center of the camera, and they're not right now! If you want to import a background that is fixed to the camera or something like that, this is a must.

Personally I think that an option to import to the center of the viewport is far more useful than the center of the canvas. First off you can import to the center of the canvas with it by simply resetting your view as users have to do in v0.6.4. Secondly, it allows you to fix the images to the canvas on any other location, not just 0,0, which allows for much more flexibility for the users to take advantage of than just a center of canvas option. I'm not opposed to having both of these options, but I am opposed to not having an option that follows the active camera.

Issue #1047, opened 2018-08-18
Request a toolbar so can do Copy/Paste/Undo without keyboard


shakin1:
I can not do anything from the keyboard, nothing happens. I need to do anything with the toolbar. For example - copy, paste, undo a last action. All these and more are not working on the keyboard
what can i do?

Jose-Moreno:
@shakin1 Hi, first of all, there is a report template laid out for a reason. Without knowing what operating system, and what version of Pencil2D it becomes very difficult for developers to understand why the problem might be happening. 

This is a bug tracker for confirmed bugs, but if you're not sure how to solve something it doesn't mean it's a bug. For that we recommend that you first use the software forum and the discord channel to ask about the problem you're having before actually submitting a bug report. 

You can find the other channels here http://pencil2d.org/community

So, please let us know what is the pencil2d version and what is your operating system version, so we can at least begin from there.

Jose-Moreno:
[2019] This was discussed to become the implementation of an actual file operations toolbar that had the most common file operations from the File and Edit menus in button form, with their respective icons, such as:
- [ ] New File
- [ ] Load File
- [ ] Save File
- [ ] Save As
- [ ] Copy
- [ ] Cut
- [ ] Paste
- [ ] Undo
- [ ] Redo
- [ ] The toolbar icons should use of the current tooltip system
- [ ] This tool bar should have a visibility toggle, so it should work similar to panels (if not even be based on the same code) in order to preserve screen real-state when the user does not need the toolbar.


Example of a file operation toolbar:
![image](https://user-images.githubusercontent.com/4912907/65920804-c5412280-e3a5-11e9-88d2-ee9431003f73.png)

davidlamhauge:
I would put those icons in a dockable window. Then it would take up less space, than a toolbar across the screen, and I (and probably many others) could close that dock, since I always use shortcuts.
That said, I still think it's a good idea, and maybe it should even be visible as default?

chchwy:
A few Microsoft Surface users have asked this as well. The toolbar will help the users who are working on a touch screen (usually without a keyboard).

Qt's toolbar is very flexible. You can drag it around like a dock widget, also can show/hide it very easily.

CandyFace:
On the flexible note I might add that Qt layout docking system is limited if you want to optimize the space of the toolbar as well as keep it possible to dock it, with our current dock setup I might add.

[...]

PR #1289, opened 2019-10-31
Implementation of operations widget

nickbetsworth:
Addresses #1047

Apologies it took so long, only managed to start working on it yesterday evening.

Let me know if you'd like you change any design decisions.

### DIFF
--- a/app/app.pro
+++ b/app/app.pro
@@ -59,6 +59,7 @@ HEADERS += \
     src/app_util.h \
     src/errordialog.h \
     src/aboutdialog.h \
+    src/operationswidget.h \
     src/toolbox.h \
     src/tooloptionwidget.h \
     src/importexportdialog.h \
@@ -91,6 +92,7 @@ SOURCES += \
     src/exportmoviedialog.cpp \
     src/errordialog.cpp \
     src/aboutdialog.cpp \
+    src/operationswidget.cpp \
     src/toolbox.cpp \
     src/tooloptionwidget.cpp \
     src/importexportdialog.cpp \
@@ -105,6 +107,7 @@ FORMS += \
     ui/importimageseqpreview.ui \
     ui/importlayersdialog.ui \
     ui/mainwindow2.ui \
+    ui/operationswidget.ui \
     ui/pegbaralignmentdialog.ui \
     ui/timeline2.ui \
     ui/shortcutspage.ui \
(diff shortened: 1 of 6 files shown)


nickbetsworth:
Does anyone know why the CI build has failed with the following error?

```
if [ "$TRAVIS_BRANCH" != "release" ]; then bash $TRAVIS_BUILD_DIR/util/checkchanges.sh; fi
fatal: Invalid symmetric difference expression origin/master...ecc645d39769f73c06aaf7d4a7fb5293b0c91171
The command "if [ "$TRAVIS_BRANCH" != "release" ]; then bash $TRAVIS_BUILD_DIR/util/checkchanges.sh; fi" failed and exited with 128 during .
```

davidlamhauge:
@nickbetsworth I don't know why Travis failed, but I know it does it all the time, and I'm always told to not worry. So - you shouldn't worry...

CandyFace:
Alrighy here's a few comments. 
My observations are that the UI widget is taking up a considerable amount of space, considering the space the buttons actually needs, this seems too much.
I also think that the icons could be a tad smaller, 16x16 or 18x18 maybe.. not quite the size of a toolbox icon. 
image

The redo icon should only be enabled when it's actually usable, same with undo.

Jose-Moreno:
@CandyFace Two questions:
1) Would it be possible to take the icons, modify them and then commit them to this PR?
2) Can we make the panel fluid enough to work in a single horizontal & vertical line as well as a "grid" layout? This would be similar to how the toolbox widget works when the size is constrained to a single column, though it doesn't work for a horizontal single row.

CandyFace:
@Jose-Moreno 
1. Yes it's possible, you should be able to push directly to his repo, granted that you have the appropriate permissions.
2. That depends... it makes use of FlowLayout so it should already behave the same, with the same limitations.

I don't think they should have configurable size if that's what you mean, but a bit smaller would be preferred. They're imo. not as important as the toolbox therefore they should not require the same sizes.

chchwy:
Hey @nickbetsworth. Thanks for your contribution & sorry for the late reply.

I had a look at this PR. Is there any reason you pick QDockWidget rather than QToolbar? Just ondering because this widget is very similar to a toolbar.

chchwy:
@nickbetsworth unfortunately I am closing this PR as similar features can be made by the `QToolbar` and I think it's the better way to do this. I appreciate your time in helping pencil2d. Please feel free to send more PR if you have any ideas. Thanks.

PR #1290, opened 2019-10-31
Miscellaneous compiler warning fixes

nickbetsworth:
Just a fix of some compiler warnings (unused variable/arguments).

The only remaining warnings are in `miniz.h` and `miniz.cpp`, but these look to be third party.


CandyFace:
LGTM, Merging.

Yes miniz is a third party dependency we're using, although we have already modified the source to fix some problems, I personally don't think it's a good idea to do so, instead it would be better to send a pull request to the owner of Miniz instead.

Thanks for the contribution 👍

PR #1293, opened 2019-11-11
Onionskin loop [WIP]

4v0v:
onionloop

A usefull feature to work on looping animations


CandyFace:
This looks like a nice addition to our onion skin feature 👍 
It would be nice though if the grey bars represented the same behaviour too, ie. they should wrap around the beginning and end of the frames.

PR #1294, opened 2019-11-14
Rewrite the windows after-build tasks by Powershell

chchwy:
I rewrote the Appveyor after-build tasks by Powershell.
The behaviour is exactly the same as the previous packing process:
1. Download FFmpeg
2. Pack and zip binaries
3. Uploading to Google drive

A few benefits of using PowerShell scripts:
1. I can run and test the script locally on my machine without AppVeyor environment.
2. It's easier to implement features with a decent script language.
3. Built-in zip & wget, so we don't need 3party programs like`curl.exe` and `7z.exe` to pack binaries.

The next step I will implement internal PR builds for windows, which was once implemented and then was reverted.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -25,7 +25,6 @@ skip_commits:
   files:
     - '*.md'
     - docs/*
-    - util/*
 
 install:
   - set PYTHON3="%PYTHON%\python.exe"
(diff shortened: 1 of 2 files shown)


chchwy:
I am going to merge it since it's working and it's not part of the main app.

PR #1295, opened 2019-11-16
1174 extended import position possibilities

davidlamhauge:
I've made this extension to the much discussed import feature.
Here you can import relative to four positions, as shown in the screenshot below.
Whether my naming for the four possibilities are precise enough is up to you. Let's discuss it.
importpos
When importing to follow camera, we have to set the viewManagers importView for each image, which called for 6 lines of code in the Editor::importImage() function. If any of you know a smarter way, please tell me how.
Finally, this only addresses the urgent problem of the merged feature, that could be more precise. We should have a feature where we could import to whatever position we want. That must be another time.


davidlamhauge:
I found a more efficient way to check import choice.
Still, improvements are welcome :smiley:

davidlamhauge:
I've used the wrong branch as basis. I close this and opens another

Issue #1224, opened 2019-05-31
[ISSUE] When Exporting a PNG image, the PNG extension is not added when selected


Kaikogf:
### Issue Summary
When wanting to export an image as a PNG file the extension is not added when it is first selected
 

### Actual Results
You have to select another extension and came back to PNG in order for it to be added.
 

### Steps to reproduce
Open Pencil2D as default
make a stroke
Go to File > Export > Image > FILE (browse to the path you will like the image to be saved) > Select the name of the image > Click Save (it will bring back the Export-Image window) > Go to the FORMAT bar and after opening all formats by clicking over the bar, select PNG by hitting it > PNG extension is not added to the image name on the File bar.
Continue the last step but this time selecting JPG > it is correctly added to the image name > now go and hit PNG > the extension is finally added to the name

 

### System Information
+ Pencil2D Version:
Compiled Code from 31/May2019

 

+ Operating System:
Windos 8.1 x 64bits

 

+ RAM Size:

 

+ Graphics Tablet:
Huion Tablet

Jose-Moreno:
@Kaikogf This for some reason has been happening to Windows users that upgrade to any version between 0.6.2 and 0.6.3+. This also happens when exporting movies.

The workaround and individual (per user) fix is that you have to manually type the filetype (e.g PNG) along with the name when the file explorer opens (e.g my_image.png), if you do that the PNG extension will continue to show properly in new work sessions.

Please try that, and let us know if it solves this issue for you, but even if it works I have to ask you not to close this Issue. I personally want to investigate this problem to see if after deleting all the Pencil2D preferences this continues to happen in Windows 7. If not then it might be a problem that is particular to Windows 8 & 10 instead.

scribblemaniac:
I do not believe this is a Windows specific issue. The program will remember whatever filename you saved as before, and if that doesn't include a filename, then it doesn't add it. It could be argued that this is not even a bug as long as the initial default filename has an extension, because the only way for it to lose that extension should be for the user to explicitly remove it at some point. Having said that, I think that it should add an extension if there is none when the dialog is opened. Perhaps an extension should even be required.

CandyFace:
^I'd still classify it as a bug, it should not be possible to export without an extension.

Jose-Moreno:
[...]

PR #1299, opened 2019-11-24
Remove extensions from translations and tweak some strings

CandyFace:
This removes extensions from translations, only text that could be interpreted otherwise is translatable now.

I've also tweaked the strings.
This means that it's not up to the translator now to write the extension, thus we should see no more of those bugs where an extension is not added because of wrong format.

Should not be merged before #1262

### DIFF
--- a/app/src/filedialogex.cpp
+++ b/app/src/filedialogex.cpp
@@ -79,17 +79,38 @@ QString FileDialog::saveFile( FileType fileType )
     QString strFilter = saveFileFilters( fileType );
     QString strSelectedFilter = getFilterForFile(strFilter, strInitialFilePath);
 
-    QString filePath = QFileDialog::getSaveFileName( mRoot,
-                                                     strTitle,
-                                                     strInitialFilePath,
-                                                     strFilter,
-                                                     strSelectedFilter.isNull() ? Q_NULLPTR : &strSelectedFilter );
+    QString filePath = QFileDialog::getSaveFileName(mRoot,
+                           strTitle,
+                           strInitialFilePath,
+                           strFilter,
+                           strSelectedFilter.isNull() ? Q_NULLPTR : &strSelectedFilter);
+
     if ( !filePath.isEmpty() )
     {
         setLastSavePath( fileType, filePath );
     }
 
+    QFileInfo info(filePath);
+    if (info.suffix().isEmpty() || strSelectedFilter.isEmpty()) {
+
+        filePath += addDefaultExtensionSuffix(fileType);
+    }
+
     return filePath;
+}
+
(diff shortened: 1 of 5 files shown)


Jose-Moreno:
@CandyFace Regarding #1224 as we identified with scribblemaniac that the problem was on the translation strings being inappropriately spaced and thus impeding the program to get the extension name correctly. So I believe this specific issue was meant to be closed by commit https://github.com/pencil2d/pencil/commit/220a27a06756bc524230211c7b80a9183eac6d23 that @chchwy pushed before

core_lib/src/util/fileformat.h
```diff
@@ -26,8 +26,42 @@ GNU General Public License for more details.
 #define PFF_EXTENSION				    ".pclx"
 #define PFF_BIG_LETTER_EXTENSION	    "PCLX"
 
-#define PFF_OPEN_ALL_FILE_FILTER	QObject::tr( "All Pencil Files PCLX & PCL(*.pclx *.pcl);;Pencil Animation File PCLX(*.pclx);;Old Pencil Animation File PCL(*.pcl);;Any files (*)" )
-#define PFF_SAVE_ALL_FILE_FILTER	QObject::tr( "Pencil Animation File PCLX(*.pclx);;Old Pencil Animation File PCL(*.pcl)" )
+#define PFF_PROJECT_EXT_FILTER \
+    QObject::tr("Supported Pencil formats") + " (*.pclx *.pcl);;" + QObject::tr("Pencil Project") + " (*.pclx);;" + QObject::tr("Legacy Pencil Project") + " (*.pcl)"
```
chchwy: I think the work "Supported" is redundant? It makes no sense to choose an unsupported format.

app/src/filedialogex.cpp
```diff
@@ -238,7 +247,6 @@ QString FileDialog::defaultFileName( FileType fileType )
         case FileType::MOVIE: return "untitled.mp4";
         case FileType::SOUND: return "untitled.wav";
         case FileType::PALETTE: return "untitled.xml";
-        default: Q_ASSERT( false );
```
chchwy: Any reason to remove this default tag?

CandyFace: Yes because all cases are handled, so having the default would cause a warning, saying it's obsolete.

core_lib/src/util/fileformat.h
CandyFace: Yes you're right, will fix that today.

CandyFace: Should be fixed now.

chchwy:
LGTM, merged.

PR #1301, opened 2019-12-12
Issue1082 camera overlay system

davidlamhauge:
safe_areas
Adding four options for usefull overlays:
Center, which makes a cross in the center of the field.
Thirds, which divides the field in thirds.
Golden Ratio, which divides the field according to the golden ratio (1:1,618)
Safe Areas, which shows the safe areas for action and titles. Each can be turned on/off separately, and can have values between 1 and 25 %.
All settings are saved in config  file.


core_lib/src/canvaspainter.cpp
```diff
@@ -552,6 +615,30 @@ void CanvasPainter::renderGrid(QPainter& painter)
     }
 }
 
+void CanvasPainter::renderOverlays(QPainter &painter)
+{
+    if (mOptions.bCenter)
+    {
+        painter.setWorldTransform(mViewTransform);
```
CandyFace: If you need to call setWorldTransform in each condition, then you could probably move that into the paintOverlays method.

core_lib/src/canvaspainter.cpp
```diff
@@ -543,6 +543,69 @@ void CanvasPainter::paintGrid(QPainter& painter)
...
+        }
+        if (mOptions.bTitleSafe)
+        {
+            int title = mOptions.nTitleSafe;
+            QRect safeTitle = QRect(rect.x() + rect.width()*title/200, rect.y() + rect.height()*title/200, rect.width()*(100-title)/100, rect.height()*(100-title)/100);
+            painter.drawRect(safeTitle);
+            painter.drawText(safeTitle.x(), safeTitle.y(), tr("Safe Title area %1 %").arg(title));
+        }
+        break;
+    }
```
CandyFace: This could be split up via private methods.

app/src/displayoptionwidget.cpp
```diff
@@ -139,3 +155,28 @@ void DisplayOptionWidget::toggleMirrorV(bool isOn)
 {
     editor()->view()->flipVertical(isOn);
 }
+
+void DisplayOptionWidget::toggleOverlayCenter(bool isOn)
+{
+    editor()->view()->setOverlayCenter(isOn);
+    editor()->preference()->set(SETTING::OVERLAY_CENTER, isOn);
+//    editor()->getScribbleArea()->settingUpdated(SETTING::OVERLAY_CENTER);
```
CandyFace: Remove this

core_lib/src/canvaspainter.cpp
davidlamhauge: Yes, it could, but the four overlays share 20 lines of code that must be repeated in every function, if they are split up.
On the other hand, I've thought of using if(...) in stead of a switch, and call from renderOverlays, if minimum one overlay is active. In this case, it would only be with painter as parameter.

core_lib/src/canvaspainter.cpp
davidlamhauge: You're right

app/src/displayoptionwidget.cpp
davidlamhauge: Forgot that...

core_lib/src/canvaspainter.cpp
davidlamhauge: I have optimized the code, but kept it in one function. Should I split it in four?

davidlamhauge: Split up in four functions. Done.

Jose-Moreno:
Alright I managed to test it :smile: 

Well I like the icons and I like the simplicity of the overlays. I think it works. I also liked that you can change the safety area area % in the preferences and is found in the general tab in the grid area so it makes sense.

The center overlay can sometimes feel a bit too large but it doesn't bother me personally. I can see it's set about 50% horizontally and vertically from center to edges, so maybe a unified customization slider for that could help similarly to the safe areas.

The design itself of the safe areas could also work with a dashed line instead of a solid line, but that's a stylistic choice I've only seen in harmony. In other software it's always solid so that's okay as well.

Maybe in a future update something i'd really like is to be able to change the color of the overlays to have better contrast. If i'm working on a very heavy lineart or dark colored animation, for example, black over dark colors doesn't tend to work that well, but right now I think it works for most user needs so I'm thinking it could be implemented later as an enhancement once we start getting feedback.

The functionality itself is straightforward so I doubt there's going to be a crash or anything due to mishandling, and at least I didn't experience any sort of issue while testing it.

CandyFace:
> The design itself of the safe areas could also work with a dashed line instead of a solid line, but that's a stylistic choice I've only seen in harmony. In other software it's always solid so that's okay as well.

@Jose-Moreno I agree that a dashed line would be better, also to distinguish the these lines from the grid lines. 

As for the colors, we could apply composition with xor that would almost always result in a line that's easy to see, the downside of using this is that the text will be not have anti-aliasing, but if that's okay then we could do that.

it would basically look like this.
lines-inverted

would that suffice or would you still need a way to tweak the line colors?

Jose-Moreno:
@CandyFace Yes, as long as the guidelines are always clearly visible regardless of the backdrop I think that's quite acceptable.

davidlamhauge:
If you prefer dashed lines, that's no problem. 
I don't know how to use xor in the composition, but maybe you could send me some information @CandyFace , or send a PR to my branch?

CandyFace:
@davidlamhauge Whoops.. I thought I had a local branch of the PR but instead accidentally pushed it directly to your branch.

davidlamhauge:
No problem @CandyFace , as long as it works ;-)

CandyFace:
Alright I'll review it later today again and merge it if I find no issues :)

CandyFace:
@Jose-Moreno  I've created four alternative overlay icons, making them pop a bit more and color coded them to easily be able to distinguish them from each other.
image

What do you think of that, compared to the ones @davidlamhauge created?
Does it convey the meaning of the overlays better or should we stick with the although quite ambiguous "GR" and "SAFE" text ?

Jose-Moreno:
@CandyFace Hey. Well they look more colorful 😄. I like them as well. 

If you're going to hace the reticule size to cover the entire icon space (like the green one does) then the blue one should also fill the entire icon, as "thirds" cover the entire screen, otherwise it looks like "pound" sign.

The orange one does remind me of the safe area, I think having the icon will help to easily remember what it is, so if given the choice i'd say it's ok to upgrade them to these icons.

As an additional note the orange icon could probably benefit to have these little mid section lines that intersect the boundary stroke on the outside to give visual relationship to the camera registration marks of **animation layout templates** (not necessarily drawn as shown though)

image

But other than that you have my blessing 😉

davidlamhauge:
One of your small changes has been to move the icons into a grid in stead of horizontal alignment. Couldn't it, please, be two rows with four icons, in stead of four rows with two icons? Just like the tools panel.
On a laptop, the unnecessary rows just take up extra space, and with Jose's new onion skin window, the upcoming bitmap coloring window, and maybe other new dockable windows, the last thing we want is unused space.

PR #1311, opened 2020-02-26
Apply more saturated color scheme to layers and highlight more distinct

CandyFace:
In correlation to #1308 which was made to get more distinct highlight, I've come up with this change. The highlight is always based on the color behind.
layer colors


davidlamhauge:
I can fully support this elegant solution, for a more visible highlight for the active layer.
I'm still puzzled and confused over the fact, that I was more than a year to see that there indeed was a faint white shadow on the active layer. We really need a more visible highlight for the active layer.

CandyFace:
I've applied gradient to tracks and labels that are not selected now too, now there's definitely no doubt that a track is selected.
gradients

PR #1317, opened 2020-03-29
Layout tweaks

scribblemaniac:
This PR tweaks a few parts of the display and tools dock widgets' appearance. These include:
- Updating the minimum height of the widgets which use FlowLayout on resize so that they cannot be resized in such a way that hides tools or other items.
- Use FlowLayout instead of GridLayout for the display options widget
- Add support for horizontal centering to FlowLayout. I currently only have this enabled for the display widget as it looks a little weird to me when the tools are not left-aligned like in all other programs.
- Added a spacer to the bottom of the onion skins dock widget just so it behaves like the tool options widget

And here are some visuals

FlowLayout Resizing:
Toolbox Resizing 1
Toolbox Resizing 2

Horizontal centering:
Display Centering

Note that if you move a widget with FlowLayout from floating to a dock with a smaller width, it may overflow into the widget below. I have not found a fix for this, but I don't think it is very high priority since interacting with the widget in any way will trigger some update which fixes the issue.

Also note that this has only been tested on Linux. Given the slightly hackish nature of this solution, it is possible that it will not work on some other operating system. We should test Mac and Linux before applying these changes.

### DIFF
--- a/app/src/displayoptionwidget.cpp
+++ b/app/src/displayoptionwidget.cpp
@@ -43,7 +43,14 @@ DisplayOptionWidget::~DisplayOptionWidget()
 void DisplayOptionWidget::resizeEvent(QResizeEvent *event)
 {
     BaseDockWidget::resizeEvent(event);
-    int minHeight = ui->innerWidget->layout()->heightForWidth(event->size().width()) + layout()->margin()*2;
+    int margins = layout()->margin()*2;
+#ifdef __APPLE__
+    // For some reason the behavior of minimumSize and the margin changes on mac when floating, so we need to do this
+    if (isFloating()) {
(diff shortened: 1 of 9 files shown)


scribblemaniac:
> Would be nice to move that resize chunk that's repeated inside BaseDockWidget but I can see we use a non generic ui element each time... couldn't we pass the height in as a param to the BaseDockWidget? that's a bit better than having this repeated per time we need to use the flowLayout.

Yeah it would be a good idea to make a helper function for that in BaseDockWidget. I'll try that soon.

scribblemaniac:
Alright, this is refactored and ready to go now hopefully.

PR #1318, opened 2020-03-29
Fix MinGW windows compiling issues

CandyFace:
Now that I've got a windows dev environment up and running I was able to reproduce those obscure compiler errors that @Jose-Moreno has been mentioning for a while.

The first part of the problem is that mingw does not use release and debug destination folder by default, so I've restructured the project file to use only that when it's msvc and none when it's g++

The second problem came from an earlier change we've made to the miniz library to fix some encoding issues.. that's good and all but it didn't compile with mingw64... so I looked online and found out that for g++, the ```<locale>``` header is needed. Please verify that it comples on both linux and windows still. I've tested it on my windows machine with MinGW but not MSVC.

### DIFF
--- a/app/app.pro
+++ b/app/app.pro
@@ -185,17 +185,26 @@ linux {
     INSTALLS += bashcompletion zshcompletion target mimepackage desktopentry icon
 }
 
-
-
 # --- core_lib ---
-win32:CONFIG(release, debug|release): LIBS += -L$$OUT_PWD/../core_lib/release/ -lcore_lib
-else:win32:CONFIG(debug, debug|release): LIBS += -L$$OUT_PWD/../core_lib/debug/ -lcore_lib
-else:unix: LIBS += -L$$OUT_PWD/../core_lib/ -lcore_lib
 
 INCLUDEPATH += $$PWD/../core_lib/src
+DEPENDPATH += $$PWD/../core_lib/src
 
-win32-g++:CONFIG(release, debug|release): PRE_TARGETDEPS += $$OUT_PWD/../core_lib/release/libcore_lib.a
-else:win32-g++:CONFIG(debug, debug|release): PRE_TARGETDEPS += $$OUT_PWD/../core_lib/debug/libcore_lib.a
-else:win32:!win32-g++:CONFIG(release, debug|release): PRE_TARGETDEPS += $$OUT_PWD/../core_lib/release/core_lib.lib
-else:win32:!win32-g++:CONFIG(debug, debug|release): PRE_TARGETDEPS += $$OUT_PWD/../core_lib/debug/core_lib.lib
-else:unix: PRE_TARGETDEPS += $$OUT_PWD/../core_lib/libcore_lib.a
+CONFIG(debug,debug|release) BUILDTYPE = debug
+CONFIG(release,debug|release) BUILDTYPE = release
+
+win32-msvc*{
+  LIBS += -L$$OUT_PWD/../core_lib/$$BUILDTYPE/ -lcore_lib
+  PRE_TARGETDEPS += $$OUT_PWD/../core_lib/$$BUILDTYPE/core_lib.lib
+}
+
+win32-g++{
+  LIBS += -L$$OUT_PWD/../core_lib/ -lcore_lib
+  PRE_TARGETDEPS += $$OUT_PWD/../core_lib/libcore_lib.a
+}
+
+# --- mac os and linux
+unix {
+  LIBS += -L$$OUT_PWD/../core_lib/ -lcore_lib
+  PRE_TARGETDEPS += $$OUT_PWD/../core_lib/libcore_lib.a
+}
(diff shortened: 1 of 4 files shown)


CandyFace:
Sigh.. so fixing the issue for gcc makes it complain in msvc.. I've withdrawn the miniz change again. Any suggestions how we can fix this?

chchwy:
What makes MSVC complain?

CandyFace:
As soon as I include the <locale> for the __MINGW64__ macro, it the MSVC compiler can't find any of the following:
```
miniz.cpp
Using fopen, ftello, fseeko, stat() etc. path for file I/O - this path may not support large files.
..\..\core_lib\src\miniz.cpp(3243): warning C4100: 'pArray': unreferenced formal parameter
..\..\core_lib\src\miniz.cpp(3930): error C3861: 'ftello': identifier not found
..\..\core_lib\src\miniz.cpp(3934): error C3861: 'fseeko': identifier not found
..\..\core_lib\src\miniz.cpp(3960): error C3861: 'fseeko': identifier not found
..\..\core_lib\src\miniz.cpp(3966): error C3861: 'ftello': identifier not found
..\..\core_lib\src\miniz.cpp(4006): error C3861: 'ftello': identifier not found
..\..\core_lib\src\miniz.cpp(4010): error C3861: 'fseeko': identifier not found
..\..\core_lib\src\miniz.cpp(4013): error C3861: 'ftello': identifier not found
..\..\core_lib\src\miniz.cpp(5738): error C3861: 'ftello': identifier not found
..\..\core_lib\src\miniz.cpp(5742): error C3861: 'fseeko': identifier not found
..\..\core_lib\src\miniz.cpp(5817): error C3861: 'ftello': identifier not found
..\..\core_lib\src\miniz.cpp(6696): error C3861: 'ftello': identifier not found
..\..\core_lib\src\miniz.cpp(6698): error C3861: 'fseeko': identifier not found
..\..\core_lib\src\miniz.cpp(6731): error C3861: 'fseeko': identifier not found
..\..\core_lib\src\miniz.cpp(6732): error C3861: 'ftello': identifier not found
..\..\core_lib\src\miniz.cpp(6733): error C3861: 'fseeko': identifier not found
qminiz.cpp
activeframepool.cpp
```

chchwy:
Probably it's just a missing underscore. it's `_MSC_VER` not `MSC_VER`

CandyFace:
Weird... it definitely has an underscore in my code
```
#if defined(_MSC_VER) || defined(__MINGW64__)
```

chchwy:
This is my code and it compiles in VS2019
```cpp
#if defined(_MSC_VER) || defined(__MINGW64__)
#include <codecvt>
#include <string>
#endif

#ifdef __MINGW64__
#include <locale>
#endif

#if defined(_MSC_VER) || defined(__MINGW64__)
static FILE *mz_fopen(const char *pFilename, const char *pMode)
{
  ...
```

chchwy:
And MSVC has no issue with an extra `#include<locale>` so this compiles too:

```cpp
#if defined(_MSC_VER) || defined(__MINGW64__)
#include <codecvt>
#include <string>
#include <locale>
#endif

#if defined(_MSC_VER) || defined(__MINGW64__)
static FILE *mz_fopen(const char *pFilename, const char *pMode)
{
  ....
```

CandyFace:
Weird, you're right... I must missed an underscore then..

CandyFace:
Agh there seems to be another problem as well, on my dev setup running windows 10, the project files because it doesn't expect a release or debug folder when compiling with mingw... however when david runs the project on his PC, the mingw compiler expects a release and debug folder for the libraries to be moved correctly. I've inspected his code i see nothing different, the only difference is that he's using Qt 13 and I'm using Qt 14.1...

chchwy:
If you need to update include paths of `app.pro`, it's very likely you need to update the paths of `tests.pro`, too.

app/app.pro
```diff
@@ -185,17 +185,26 @@ linux {
...
+CONFIG(release,debug|release) BUILDTYPE = release
+
+win32-msvc*{
+  LIBS += -L$$OUT_PWD/../core_lib/$$BUILDTYPE/ -lcore_lib
+  PRE_TARGETDEPS += $$OUT_PWD/../core_lib/$$BUILDTYPE/core_lib.lib
+}
+
+win32-g++{
+  LIBS += -L$$OUT_PWD/../core_lib/ -lcore_lib
+  PRE_TARGETDEPS += $$OUT_PWD/../core_lib/libcore_lib.a
```
chchwy, line 203: Don't know why, but MinGW does need the `$$BUILDTYPE` in lib paths on my PC. It turns ou no need to differentiate MSVC and G++. I am using Qt 5.12.7 and MinGW 7.3.0.

CandyFace: And that's the thing I found weird.. it seems to be inconsistent, because on my windows setup, it doesn't. so there's still a problem as far as I can see.. maybe we should just force release and debug folders allover and be done with it

CandyFace:
Should compile for Qt 14 and lower now.

CandyFace:
@chchwy will you review when you have time? I think it's ready to be reviewed and otherwise merged.

chchwy:
Sure thing. Will review it today.

PR #1323, opened 2020-03-30
Update pencil_sv.ts

eson57:
Update and complete Swedish translation from Transifex


chchwy:
Hey eson57! Thanks for your contribution.
Is there any reason that you have to send a PR here rather than do it in Transifex?

eson57:
Afaik transifex don't accept strings that don't match source file.

chchwy:
Haha, yes. I just updated the source strings on transifex today. But anyway thanks heaps.

PR #1324, opened 2020-04-02
Increase visibility of active layer keyframes

### LINKED ISSUE #1322: Increase visibility of active layer keyframes
### Issue Summary

This is a UI / UX request after having problems myself with the rapid location of the frames I was working on. 

After the color optimizations made in #1311 the active layer frames become a bit too dark to see, particularly on the bitmap layer (considering blue has a low lightness value in gresycale) and get lost even more when they are selected (color turns to dark grey).

My requests for the actively selected layer are:
- Increase the lightness of the base keyframes on the ACTIVE layer by a 25%, or change the color slightly towards white when the layer is selected.
- Change the selected keyframes color to a bright color or if possible a complementary color instead of dark grey.

### Video or Image Reference
**Before**
image

**After**
image

This image has the same gradient overlay as Pencil2D, but it goes to show just how potent it is for a bright complementary color to appear as the keyframe selection color. 

image

However i'm aware it might not be visually appealing to just make all layer types have different selection colors, rather it could create confusion, and it would make it difficult to harmonize the concept of having elements selected on the timeline in the first place.

Here I'm trying a lighter grey instead of the dark grey.
image

Thoughts?

CandyFace:
Includes:
+ current keyframe is highlighted with white
+ All layer keyframes are more visible

colors2

closes #1322

### DIFF
--- a/core_lib/src/interface/timelinecells.cpp
+++ b/core_lib/src/interface/timelinecells.cpp
@@ -181,6 +181,8 @@ void TimeLineCells::drawContent()
         }
     }
 
+    mCurrentFrame = mEditor->currentFrame();
+
     QPainter painter(mCache);
 
     Object* object = mEditor->object();
@@ -203,6 +205,7 @@ void TimeLineCells::drawContent()
             continue;
         }
         Layer* layeri = object->getLayer(i);
+
         if (layeri != nullptr)
         {
             switch (mType)
--- a/core_lib/src/interface/timelinecells.h
+++ b/core_lib/src/interface/timelinecells.h
@@ -62,6 +62,7 @@ public:
     void clearCache() { if ( mCache ) delete mCache; mCache = new QPixmap( size() ); }
     void paintLayerGutter(QPainter& painter);
     bool didDetatchLayer();
+    int getCurrentFrame() { return mCurrentFrame; }
 
 Q_SIGNALS:
     void mouseMovedY(int);
@@ -90,6 +91,7 @@ private slots:
     void loadSetting(SETTING setting);
 
 private:
+
     TimeLine* mTimeLine;
     Editor* mEditor; // the editor for which this timeLine operates
     PreferenceManager* mPrefs;
@@ -106,6 +108,8 @@ private:
     int mLayerHeight = 20;
     int mStartY = 0;
     int mEndY   = 0;
+
+    int mCurrentFrame = 0;
 
     int mFromLayer = 0;
     int mToLayer   = 1;
--- a/core_lib/src/structure/layer.cpp
+++ b/core_lib/src/structure/layer.cpp
@@ -340,16 +340,18 @@ void Layer::paintTrack(QPainter& painter, TimeLineCells* cells,
         }
         else
         {
+            painter.save();
             QLinearGradient linearGrad(QPointF(0, y), QPointF(0, y + height));
             linearGrad.setColorAt(0, QColor(255,255,255,150));
             linearGrad.setColorAt(1, QColor(0,0,0,0));
             painter.setCompositionMode(QPainter::CompositionMode_Overlay);
             painter.setBrush(linearGrad);
             painter.drawRect(x, y - 1, width, height);
-            painter.setCompositionMode(QPainter::CompositionMode_SourceOver);
+            painter.restore();
         }
 
-        paintFrames(painter, cells, y, height, selected, frameSize);
+        paintFrames(painter, col, cells, y, height, selected, frameSize);
+
         painter.restore();
     }
     else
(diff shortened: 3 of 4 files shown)


scribblemaniac:
Something that was overlooked in this PR is overlapping sound frames. They are now nearly impossible to distinguish.

Before:
before

After:
after


In my opinion, the second sound frame here is too difficult to see.

Jose-Moreno:
@scribblemaniac Ouch. I haven't yet tested these as I was waiting for the other sound PR's (but then I got busy sorry). Thanks for looking into it! I agree it's not easy to see now, although I do wonder if the playback header lands on a sound keyframe, does the whole container lit up? or only the frame, or it doesn't work for sound frames? 🤔 (Well I guess i'll know soon enough)

In this specific case I think the overlapped keyframe container should become darker, right now it seems to be slightly lighter actually 🤔

CandyFace:
MMmm... I've personally never been a fan of the ability to overlap keyframes on the same track but i can see the issue. I'll make a PR to fix it. If you need overlapping soundtracks, then you should create more tracks, just like in audacity.

davidlamhauge:
I agree with @CandyFace . Overlapping sounds on the same track is not supported in any other applications I use, kdenlive, shotcut, audacity, to mention some.
I also think that it will solve some of the audio problems we struggle with. My sound scrub feature would be perfect, if sounds were not allowed to overlap.
I also recognize that there would be problems if we changed it. 
First there is backward compatibility, because there must be many files where sounds overlap.
Secondly you could argue that you'll need many sound layers, if we didn't allow overlapping sounds. My argument here is that you should never have any sounds but lip sync sounds, in your animation file. All other sounds (music, effects, background noise etc) should be added in the video editor.

Jose-Moreno:
@CandyFace @davidlamhauge Not disagreeing with not wanting the overlapping sounds, but additionally to using more sound layers, this should be also corrected by having the ability to clip the sound containers frame length, which is a very old part of the roadmap and part of the original Pencil / Pencil2D vision.

Just as in video editors, Ideally we want to keep the ability to have multiple sounds on a single track, however those should not be allowed to be overlapped in a single track, just clipped manually, as it makes no sense otherwise (even though it may feel intuitive for a beginner to stack up sounds like lego blocks 😋 ).

Well that's wishful thinking for the future.

On the comment of not having any sounds other than lipsync, I respectfully disagree due to the following: As mentioned briefly, part of the vision for the software is to have limited video editing capabilities, but animating to pre-recorded foley or soundtracks is not strange to animators as you all know. 

The old masters would surely need to have prior musical knowledge and they'd pre-time the animation while using the musical score to map the beats on the x-sheet, but kids or even entry-level animators might not have those talents or assumed knowledge in advance, so having the soundtrack in order to animate to the musical beats is a helpful tool for animating music videos and rhythmic scenes (like a dance showdown),and I feel this is valid example of why all manner of sounds should be allowed, not just lip-sync.

It's one thing to finalize sound timing in the video editor as a director or author, and another to use pre-existing soundtracks (and of course dialogue) as guidelines to animate the action, as such I believe we shouldn't close those possibilities to our users.

As for the camera, I agree there needs to be a better indication of the camera motion, but this and having keyframes for every frame is a different conversation, as well as unrelated to the sound layer visibility.

Let's make a new issue to discuss this (i'll add it later today if that's ok)

davidlamhauge:
I have already opened a new issue for the discussion @Jose-Moreno .
As for which sounds should be added to the Sound layer, it is entirely up to the user, and when I talk about lip sync sounds, I think of any sound that you need to know the position for, to do your animation. Sounds like coughs, clapping, hammering, rain and so forth will most likely be ok to add in editing, but if they are a part of something that you need to animate in sync, then you should of course add it to the sound layer.
I use lipsync, for things that you have to have in sync while animating. Is there a better word?

Jose-Moreno:
@davidlamhauge I see I use lipsync in relation to dialogue only mainly because the word is self explanatory (lip - synchronization) Didn't know you used it in a broader context.

However terminology aside, summarizing the main point of my reply, scrubbing should work for ALL the existing sound tracks, like a video editor or sound editor does. This is the simplest (for the user) and most effective way to go about it because people also expect this.

If you want to probe a single sound you should be prompted to use the "isolate" / "solo" visibility mode and only hear that particular sound, and in CandyFace's timeline mockup there is a mute icon included so you'll be able to selectively "mute" any sound layer as well (right now we should be able to do this by turning the visibility off)

I mean we could consider ways to categorize sounds per "clip", like placing sound tags to differentiate dialogue from foley and what not, but why make this more difficult for us too?

What I was thinking is that the different sound layer data streams need to be combined and read, but actually doesn't the playback already does this? why can't we "scrub" on top of that resulting "preview" sound when moving the playback header manually? For the latter maybe @scribblemaniac and @CandyFace may have a better idea regarding this assumption of mine 🤔

PR #1325, opened 2020-04-03
Reposition canvas content for selected keyframes

davidlamhauge:
This is a feature that lets you reposition many selected frames in one go, even from several layers at the same time.
The repositioning is done on existing keyframes in the scene/project.
It is a feature that has been discussed before, but I don't think it is one of our issues.
It works as expected, I think, but is still a WIP. Please feel free to comment, and give suggestions to improvements and the like.
Here is a small video, showing how it works.
https://youtu.be/qCXFtOi1sos


davidlamhauge:
Oh Dear! I just realized something that may/may not be a problem.
I've previously worked on a feature that lets you check (a checkbox) whether you want information on size and position for your selection, when you use the Select or Move tool.
Unfortunately, that feature branch is the base of my reposition-branch... - which means that, either you will have to review two new features in this branch, or I have to divide them in two.
I could argue that the two features are related, but please tell me what you think.

davidlamhauge:
Since I look at the two features as one, I've made some enhancements, so the original selection is saved properly, and the information on the screen is consistent.

core_lib/src/managers/layermanager.h
```diff
@@ -25,6 +25,7 @@ class LayerBitmap;
 class LayerVector;
 class LayerCamera;
 class LayerSound;
+class SelectionManager;
```
CandyFace, line 28: Managers should not be aware nor depend on each other.

core_lib/src/selectionpainter.cpp
```diff
@@ -5,6 +5,8 @@
 #include "layermanager.h"
```
CandyFace, line 5: Managers should not exist inside painter methods, the paint method should only take in parameters requires.

core_lib/src/selectionpainter.h
```diff
@@ -19,7 +19,12 @@ class SelectionPainter
 public:
     SelectionPainter();
 
-    void paint(QPainter& painter, const Object* object, int layerIndex, BaseTool* tool, TransformParameters& transformParameters);
+    void paint(QPainter& painter, const Object* object, int layerIndex, BaseTool* tool, TransformParameters& transformParameters, QPolygonF original);
+
+    void setCurrentSelectionNotMapped(QPolygonF polyF) { mCurrentSelectionNotMapped = polyF; }
```
CandyFace, line 24: Preferably only the paint method should be accessible from outside.

core_lib/src/selectionpainter.cpp
```diff
@@ -67,6 +70,19 @@ void SelectionPainter::paint(QPainter& painter,
                                                width, width);
         painter.drawRect(bottomLeftCorner);
 
+        if (tool->properties.showInfo)
+        {
+            int diffX, diffY;
+            diffX = static_cast<int>(mCurrentSelectionNotMapped.boundingRect().x() - original.boundingRect().x());
+            diffY = static_cast<int>(mCurrentSelectionNotMapped.boundingRect().y() - original.boundingRect().y());
```
CandyFace, line 77: Don't declare then init, simply declare and init on the same line.

core_lib/src/tool/selecttool.cpp
```diff
@@ -35,6 +35,7 @@ void SelectTool::loadSettings()
     properties.feather = -1;
     properties.stabilizerLevel = -1;
     properties.useAA = -1;
+    mPropertyEnabled[SHOWSELECTIONINFO] = true;
```
CandyFace, line 38: Value also needs to be stored and added to properties, so the user doesn't have to enable/disable every time.

core_lib/src/managers/layermanager.cpp
```diff
@@ -16,6 +16,7 @@ GNU General Public License for more details.
 */
 
 #include "layermanager.h"
+#include "selectionmanager.h"
```
CandyFace, line 19: Same problem as before, managers should not need nor be aware of each other.

core_lib/src/managers/layermanager.cpp
```diff
@@ -173,6 +175,34 @@ QString LayerManager::nameSuggestLayer(const QString& name)
...
+        LayerBitmap* layer = static_cast<LayerBitmap*>(currentLayer());
+        QRect reposRect = layer->getFrameBounds(frame);
+        editor()->select()->setSelection(reposRect);
+        QPoint point = reposRect.topLeft();
+        point += transform;
+        layer->repositionFrame(point, frame);
+        editor()->backup(layer->id(), frame, tr("Reposition frame"));
+    }
+}
+
```
CandyFace, line 205: This doesn't belong here, it has nothing to do with layer management. Should probably be placed in Editor for now.. but maybe another manager is needed.. a CanvasContentManager maybe.. not sure right now...

davidlamhauge: I could move it to Editor, but what about Scribblearea? It is pretty much a CanvasContentManager, as it is now...

core_lib/src/structure/layerbitmap.cpp
```diff
@@ -46,6 +44,18 @@ BitmapImage* LayerBitmap::getLastBitmapImageAtFrame(int frameNumber, int increme
...
+    BitmapImage* image = getBitmapImageAtFrame(frame);
+    image->moveTopLeft(transform);
+}
+
+QRect LayerBitmap::getFrameBounds(int frame)
+{
+    BitmapImage* image = getBitmapImageAtFrame(frame);
+    return image->bounds();
+}
+
```
CandyFace, line 58: Missed this in the review...

repositionFrame is misleading, as I'd expect the implementation to reposition a keyframe and not the content of the image. A better name would be repositionImage. The parameter called transform is also misleading, as transform is often related to QTransform. Call it pos or point to relate it to a qpoint.

davidlamhauge:
I think I've addressed all requests by no.

core_lib/src/tool/basetool.h
```diff
@@ -50,6 +50,7 @@ class Properties
     int   stabilizerLevel = 0;
     qreal tolerance = 0;
     bool  useFillContour = false;
+    bool  showInfo = true;
```
CandyFace, line 53: This name is very ambiguous, please make it describe what the property does.

davidlamhauge: I've changed it to showSelectionInfo, which I call it in settings as well. Should be understandable.

CandyFace: That's fine 👍

Jose-Moreno:
Oh wow I never saw this! Excellent tool! 😄 

I like everything about it. I think in this case I think having a dialog is a nice choice so people can focus on the task of moving the layers contents.

In the far future, after this is merged, i'd love to see only needing to select all the frames on the layers we want directly, and while using multi-layer onion skin, to be able to move the selected frames content with or without the "search" selection area, this could even become activated as a tool mode for the move tool like GIMP / Krita, in the options panel, or by pressing a modifier key ✌️🙂

Anyway good job! Hoping this gets reviewed / merged soon if that's the plan 🙌 🎉

chchwy:
Will be reviewing this PR soon.

chchwy:
Generally looks good. I like the selection info on the top-left corner. 
Thank you for this useful feature @davidlamhauge

Issue #1298, opened 2019-11-16
inconsistent sound sync


DogBoyMark:
when I pay the movie, the audio is never in the same sync. If it was consistent then fine, I can sync the video when finished, but I cannot complete movie because sometimes the audio moves a few seconds before or after the video

Version: 0.6.4
Operating System: Windows 10 (10.0)
CPU Architecture: x86_64

length 00:03:40 (hh:mm:ss)
it was an Mp3
converted to .wav and seems to be better

Jose-Moreno:
@DogBoyMark Hi. This is actually a known issue although it didn't specifically had a ticket open. 

Unfortunately we might need a sound specialist programmer to fix this because right now even though there have been improvements done over the past versions to fix sync, there's still a gap between playback and exported sound, sometimes it ranges from 1-3 frames up to 20 frames depending on the user hardware.

Also the sound playback tends to have difficulties depending on these hardware restrictions so we understand that getting an accurate sound replay at any given time frame is sometimes not possible.

For now please update / edit your initial comment to properly write out part of the original template...we need that info to help developers at least know where they have to test things out, as well as:

- Pencil2D version
- Operating System
- Type of sound used (mp3 vs wav)
- Length of the sound (seconds, minutes, etc)
- If possible share the sound or the Pencil2D project and the movie export as a zip file (_you can use a cloud file storage account you already own or use www.wetransfer.com and look for the **anonymous link option**_)

rebe83:
Not the above person, but I am also having this issue. 
Version: 0.6.4
Operating System: Windows 10 (10.0)
CPU Architecture: x86_64
Sound used: mp3
audio length: 14:29

What seems to be happening is, for every frame, the moment you pause on isn't the same as what that frame is loaded to start on, and so the audio that is played is from seconds beforehand.
So for example, if I'm starting the animation at frame 1, everything will be in sync the whole way through, but if I ever pause, the sound will end up off by a some amount. (Unsure if it increases in offset the farther into the sound the pause happens)
You can see what I mean by starting the animation, and pausing when it gets to the second keyframe(frame 32). When you press play, it will play the audio that originally played at frame 22. So by that point, the "resume" audio is already 10/9ths of a second behind.

I was keyframing the project when I noticed this problem, so the screens I have weren't changing at the right time with respect to the audio. Therefore some frames and words written on them are accurate to "resume" times where i was going keyframe to keyframe and not start-to-keyframe.
[...]

Issue #1227, opened 2019-06-19
Audio almost always lags when when pressing play. i have to go to the start of the animation and press play a few times and only then the audio works.


DWDemon:
### System Information
+ Pencil2D Version:

 v.0.6.4? i just downloaded the newest one

 

+ Operating System: 
windows 10
 

+ RAM Size: 

8.00GB

 

+ Graphics Tablet: 
wacom intuos pro

Jose-Moreno:
@DWDemon Well this depends on many variables and actually the behavior you're describing is meant to happen like that for the animations previewed in the editor. In Pencil2D there's a feature called the "image cache pool" basically it helps compensate when your computer has lower resources to spare so it "pre-loads" the images. But the only way to do it is by playing back the animation so the "cache" is created.

The image cache is set to 200 images by default but you can increase this number, however we only recommend this if your computer has a LOT of RAM (e.g 8~16+ GB)

Now, the issue where the sound is lagging that's a new one. I'm not sure if the pool cache also effects the audio, and while it wouldn't surprise me, I'll ask the other devs to see if this is true or not. Depending on that we can further investigate this problem. 

However, please, edit the issue and fill it with the info we ask of you on the template. We at least need to know:
1. What is your current version of Pencil2D
2. What is your Operating system and it's version
3. And record a video of the problem to see how the sound lags, and upload it to streamable.com (which can be done anonymously) so everyone has a better chance to find the cause of this, otherwise it will take a long time to get work done.

SanoStar:
Hello this is my first time participating in debug forums.
I just ran into an issue where Audio won't properly play unless I play it from the beginning.
I was writing down the lyrics to the song to keep track of the timing and aligning the keyframes to the song beats, however when I'd play in the middle of the song it would add about +5 secs to the misalignment of sound and timeline every time I paused and played. (Pause +5 sec off)

Now I have to play the sound from the beginning every time if I want it to be accurate, and it's REALLY off otherwise.

This is my first time providing a video example of an issue, I hope it works.

[SoundLag.zip](https://github.com/pencil2d/pencil/files/3679489/SoundLag.zip)

Jose-Moreno:
@SanoStar A new release will be shipping soon with what the developers consider fixes this problem. I'm closing this issue and we ask you to test the new release next week. If you'd like you can also test a development build closer to the release here:

[...]

PR #1326, opened 2020-04-05
Fix playback audio sync issues

### LINKED ISSUE #1179: Sound Layer Audio Delay
I can easily import audio into my sound layer in Pencil, but when I press play the audio does not play until after 12-14 frames. I have tried both MP3 and WAV files. When I export the animation, the audio plays normally at the very beginning and makes my animation no longer sync up, as I have animated after 12 or so frames to sync with the audio.

EXAMPLES:
What it sounds like in the program 
https://youtu.be/Vbi-x8-K44Q

What it sounds like exported 
https://www.youtube.com/watch?v=jFYCEKImgIY

Pencil2D Version: 0.6.0 
I downloaded 0.6.2 and had the same issue 

Operating System: Windows 10

RAM Size: 4 GB

scribblemaniac:
This PR loads and keeps audio files in memory, and also fixes two small coding errors which were also causing audio delays. I am perhaps overly optimistic that this will completely fix #1179, #1227, #1298, and the many user complaints we've had everywhere else. This code still needs to be tested on Windows where the issue seems to be worst.

Assuming this works as I expect it to, the only remaining known issue with playback audio synchronization will occur when Pencil2D is not scrubbing at the target frame rate (either because it's set too high or the computer cannot process the frames fast enough).

### DIFF
--- a/core_lib/src/managers/playbackmanager.cpp
+++ b/core_lib/src/managers/playbackmanager.cpp
@@ -101,6 +101,7 @@ void PlaybackManager::play()
     if (frame >= mEndFrame || frame < mStartFrame)
     {
         editor()->scrubTo(mStartFrame);
+        frame = editor()->currentFrame();
     }
 
     // get keyframe from layer
@@ -117,21 +118,10 @@ void PlaybackManager::play()
         }
     }
 
-    // check list content before playing
-    for (int pos = 0; pos < mListOfActiveSoundFrames.count(); pos++)
-    {
-        if (key != nullptr)
-        {
-            if (key->pos() + key->length() >= frame)
-            {
-                mListOfActiveSoundFrames.takeLast();
-            }
-        }
-        else if (frame < mListOfActiveSoundFrames.at(pos))
-        {
-            mListOfActiveSoundFrames.clear();
-        }
-    }
+    mListOfActiveSoundFrames.clear();
+    // Check for any sounds we should start playing part-way through.
+    mCheckForSoundsHalfway = true;
+    playSounds(frame);
 
     mTimer->setInterval(static_cast<int>(1000.f / mFps));
     mTimer->start();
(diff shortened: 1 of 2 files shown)


CandyFace:
Exciting! I'll review this.

scribblemaniac:
A follow up on this issue: I have tested this on Windows and it does improve audio playback. Unfortunately, there are still two issues (both exclusive to Windows):
1. There still a ~400-800 ms delay for me upon playing a clip for the first time. I have no idea what could be causing this because the resource manager clearly shows that the sound is already in memory.
2. There is a subtle fade-in when playing a sound clip, whether from the start or from the middle of the clip.

davidlamhauge:
Two things:
1) How do you measure the delay to be 400-800 ms, or is it guesswork?
2) I found out that your solution has the same flaw as my limited sound scrub feature. 
I have a scene with two sounds on the sound layer. One covers frame 8-57 and the other frame 17-122. If I start playback when on frame 1-16, it plays both sounds, but if I start playback from frame 17 and up, it just plays the last sound, and skips what remains of the first.
Tested on Linux (Ubuntu 18.04) and Win 10.

scribblemaniac:
1. It was just an educated guess based on how long it appeared to be.
2. The issue is that we use getKeyFrameWhichCovers in PlaybackManager, which returns at most one keyframe. This issue was "hidden" prior to this PR by mListOfActiveSoundFrames which acted as a cache so that if you were playing multiple sound and then paused, it would remember the sounds that were previously playing when starting to play again. However this was a bad solution because if you started playing at frame 17 and up (in your example), without first playing a frame in the range 8-16, it would not play the first sound. The simplest solution would be to iterate over all previous keyframes checking for ones which are active over the current keyframe. A more complicated solution may be necessary for performance reasons, depending on how often we need to do a full check like that.

PR #1328, opened 2020-04-05
New layer is selected after creation

### LINKED ISSUE #1327: New layers should become automatically selected after creation
### Issue Summary

This is something that I've been thinking for a long time. Seems like a minor issue but this should lead the user to not second guess themselves during an intense workflow.

The idea is that any action should always put the user in a state where they are aware of the UI and how to use it.

Creating a new layer should make this new layer selected because you are creating it with the purpose of using it next. Even if you create many layers (e.g bitmap) at the same time to use them later, the last one that was created should be your starting point to work on your animation.

Currently creating a layer leaves the selection focus on the original layer, and when you start drawing you expect that it's actually the new one that's selected.

davidlamhauge:
New layer is selected when created. Only tested on Win10, so it should work on all platforms ;-)
closes #1327

### DIFF
--- a/core_lib/src/managers/layermanager.cpp
+++ b/core_lib/src/managers/layermanager.cpp
@@ -181,7 +181,6 @@ LayerBitmap* LayerManager::createBitmapLayer(const QString& strLayerName)
     const QString& name = nameSuggestLayer(strLayerName);
     layer->setName(name);
 
-    Q_EMIT layerCountChanged(count());
     setNewLayerAsSelected();
 
     return layer;
@@ -193,7 +192,6 @@ LayerVector* LayerManager::createVectorLayer(const QString& strLayerName)
     const QString& name = nameSuggestLayer(strLayerName);
     layer->setName(name);
 
-    Q_EMIT layerCountChanged(count());
     setNewLayerAsSelected();
 
     return layer;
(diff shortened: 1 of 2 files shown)


davidlamhauge:
Since I don't like the same code repeated in many (four) functions, I now also moved the Q_EMIT to the common function setNewLayerAsSelected().
I'm not sure it's the right decision, and maybe we should omit the common function all together? Two lines repeated in four functions, is not that big a crime...

core_lib/src/managers/layermanager.cpp
```diff
@@ -180,7 +181,7 @@ LayerBitmap* LayerManager::createBitmapLayer(const QString& strLayerName)
     const QString& name = nameSuggestLayer(strLayerName);
     layer->setName(name);
 
-    Q_EMIT layerCountChanged(count());
+    setNewLayerAsSelected();
```
chchwy, line 185: Probably not here. `load()` is called after a Project is loaded. It looks a bit weird to see aything related to NewLayer here.

CandyFace: It's not being called in load(), github is showing a slightly confusing image.. the method is only added in createLayer methods.

chchwy: Ah, got it. sorry my bad.

davidlamhauge:
After thinking about it I decided to create a getLastLayerIndex() function, since it may come in handy in other cases.

CandyFace:
Tested and works as expected, well done 👍

PR #1331, opened 2020-04-06
LayerVisibility updated in settings. 

### LINKED ISSUE #1319: Opacity slider resetting visibility mode
### Issue Summary

For the re-implemented "light table mode" or "relative layer opacity" system, when using the layer visibility slider in the preferences, the visibility mode is always reset to opaque.

### Expected Results
The visibility slider should not reset the current visibility mode. It should keep the one it's currently active and the opacity should only effect the layers in "relative" mode (grey dot presentation)

### Video or Image Reference

image

### Steps to reproduce

1. On a new Pencil2D file
2. Draw a simple stroke on the bitmap layer
3. Create a new bitmap layer on top and draw on it to notice the effect
4. Click on the visibility mode button on the left of the timeline or press the new command `ALT + 2` to switch visibility mode to relative
5. Go to `Edit > Preferences > Timeline` and scroll down to the `Layer Visibility` section
6. Move the slider to a very different opacity
7. Notice the visibility "mode" is set to opaque
8. Set the visibility again to relative
9. Note the opacity is read properly

### System Information
+ Pencil2D Version:

Nightly build

commit: 7b0d1134d2853e4371c3cc8e3dd021ed00f2cf64
date: 2020-03-19_18:05:15

Operating System: Windows 10 Pro (v1909)
CPU Architecture: x86_64
Qt Version: 5.12.6

davidlamhauge:
This PR updates the layer visibility in settings at all times.
In the Preferences, the slider and spinbox are disabled, when not in RELATIVE mode.
closes #1319

### DIFF
--- a/core_lib/src/interface/scribblearea.cpp
+++ b/core_lib/src/interface/scribblearea.cpp
@@ -1478,34 +1478,17 @@ void ScribbleArea::setLayerVisibility(LayerVisibility visibility)
     updateAllFrames();
 }
 
-void ScribbleArea::updateLayerVisibility()
-{
-    switch (mLayerVisibility) {
-    case LayerVisibility::ALL:
-        mEditor->preference()->set(SETTING::LAYER_VISIBILITY, 0);
-        break;
-    case LayerVisibility::RELATED:
-        mEditor->preference()->set(SETTING::LAYER_VISIBILITY, 1);
-        break;
-    case LayerVisibility::CURRENTONLY:
-        mEditor->preference()->set(SETTING::LAYER_VISIBILITY, 2);
-        break;
-    default:
-        break;
-    }
-}
-
 void ScribbleArea::increaseLayerVisibilityIndex()
 {
     ++mLayerVisibility;
-    updateLayerVisibility();
+    mPrefs->set(SETTING::LAYER_VISIBILITY, static_cast<int>(mLayerVisibility));
     updateAllFrames();
 }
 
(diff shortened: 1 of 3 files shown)


core_lib/src/interface/scribblearea.cpp
```diff
@@ -1478,15 +1478,34 @@ void ScribbleArea::setLayerVisibility(LayerVisibility visibility)
...
+    case LayerVisibility::RELATED:
+        mEditor->preference()->set(SETTING::LAYER_VISIBILITY, 1);
+        break;
+    case LayerVisibility::CURRENTONLY:
+        mEditor->preference()->set(SETTING::LAYER_VISIBILITY, 2);
+        break;
+    default:
+        break;
+    }
+}
```
CandyFace, line 1496: There's no need for a switch here, simply static cast the enum to int and set preference using that.. eg..
mPrefs->set(SETTING::LAYER_VISIBILITY, static_cast<int>(mLayerVisibility));

Also notice, we already hold a reference to PreferenceManager, so there's no need to call editor to get it.

davidlamhauge:
I didn't know you could static cast an enum. Good to know!

core_lib/src/interface/scribblearea.cpp
davidlamhauge: Done. Thanks for the advice.

CandyFace:
Can confirm it's been fixed, merged.

PR #1333, opened 2020-04-08
Sound scrubbing, using shortcuts or mouse

davidlamhauge:
This PR will **not** solve the issue 1072 entirely.
I would go so far as call it a poor mans sound scrub. A limited version.
If you have a file with a sound layer and a sound loaded, you can scrub and hear the sound by using the shortcuts for frame-forward and frame back.
It is a scraped feature. If you have loaded two sounds on the layer, and no. 1 covers frame 5-50, and no. 2 covers frame 30-100, then you will be able to sound scrub no. 1 from frame 5-29, and no. 2 from frame 30-100. Only one clip at a time.
If there is a fast way to merge 2 or more sounds together, then I'm all ears.
I made this feature because I need something for my private, bootleg version of Pencil2D, and if you think other users could benefit from it, then lets use it.


Jose-Moreno:
This is great in the sense that you are making Pencil2D your own. I feel that is amazing 😄 I even recall your xsheet implementation had sound scrubbing too 🙌

With that said, I'd honestly prefer this to be much more fleshed out. I know the lack of sound scrubbing is a pain point without a doubt, but giving the users a limited feature is going to bite us in the behind later. 

Just recall how much damage the accidental loss of copy/paste did for 0.6.2. It was a disaster. And In a similar fashion to how we got crowded with reports despite announcements and crowd control support, we don't have the manpower to explain every single user that doesn't read the news (or simply has no access to the media platforms), how the partial sound scrubbing is meant to work. 
I'm positive many will come to report that the scrubbing is defective and what not.

I'm also worried that having a partial sound scrubbing will be affected by the sound delay problem that is still unsolved amplifying both issues. And even if scribble's latest commit helps with the latter, we are still not clear how much work is left to do in order to finally fix the problem for audio playback.

So I'd rather ask the following:
1. Keep this commit open, so you or anyone else who can help can continue to add commits to it until it's sufficiently fleshed out to work as people expect it to.
2. Do not merge until the long-standing sound delay issues are fixed. As this will certainly effect the outcome of sound scrubbing at some point.

I'd like to know what the others think as well in case there can be another solution like compartmentalize the improvements made here and separate them into various commits or something else.

davidlamhauge:
Thanks for the comments @Jose-Moreno .
I will of course keep it open, and if the team thinks it has potential, it will become a real feature.
As it is now, it is an improvement, and, as I write, if I can be told an efficient way to include all sounds on a layer, it will be (almost) full-fledged.
The sound delay problem is not affecting this sound scrub feature. This feature takes the last sound clip that covers the current frame, and plays 0,15 secs from that point. No more - no less. I don't use the play() function at all.
That is why it is fairly efficient, and that is why I dared to make the PR.

davidlamhauge:
I have (a little boldly) stricken the 'limited version' from this PR.
In the Preferences, you can check whether you want sound scrubbing to be active or not.
If it is active, you use it by pressing the shortcuts for 'Next frame' and 'Previous frame'. It then plays the sounds on that  frame, for a period of between 100 and 250 milliseconds. You can set the playing length in the Preferences,

Jose-Moreno:
@davidlamhauge Alright I tested your branch. At first it didn't work for me, for unknown reasons no sound was coming out after using the shortcuts. It only succeeded after changing the msec value on the preferences.

The way I got the problem was:
1. Compile & Run the program
2. File > Import > Sound > Create Layer > browse for WAV (I first imported a long BGM and then a shorter dialogue file by creating a new sound layer manually; I tested both by playback and by scrubbing then remembered i had to turn it on)
3. Edit > Preferences > Timeline > Turn Scrubbing ON (it was off by default)
4. Try scrubbing the sound, no sound.
5. Change the msec value from the default 180msec to 100 msec
6. Scrubbing worked with shortcuts.

It seems this was a first time thing however because after that it kept working as expected no matter what I did.

Now regarding workflow, I was also expecting that we could scrub by dragging the playback indicator, not in vain it is called scrubber in other apps. Actually my first reflex was to try moving this and I kept wondering why it wouldn't give sound (this was after turning sound scrubbing on and forgetting you mentioned the shortcuts) because, at least in Flash and Harmony, you can't scrub sound using shortcuts, you have to drag the playback indicator for a sort of freehand scrubbing.

When using this in (Video Editors), if you click on the playhead the sound preview starts off from the current frame it is positioned on. Basically it **previews on click.**
In other software (e.g Harmony, Flash, AEFX) you have to drag the playhead at least up until to the next frame so it plays the current frame's sound preview, so it **previews on click&drag**

So far I think it works, but even so due to the change between frames the sound is a bit choppy, however I think it is a problem of using the shortcuts and not because of the implementation. Considering this, If possible I'd like to ask you to consider one last request of adding the sound scrubbing via dragging the playhead.

This really helps a lot when gauging the relationships between the phonemes in a more organic way and it allows you to control the speed of the scrubbed sound better than pressing the shortcuts because these are limited by the keyboard input filter and will always cut off slightly so you can't get a good impression of any given portion of the sound (and this beats having to setup range loops for different parts when you are quickly measuring the phonemes).

Also I think the scrubbing activation toggle should be located on the timeline next to the volume speaker icon (maybe a speaker with the letter "S"?) and not on the preferences since it's hard to find (the msec bar is fine to keep in the prefs though as it's not necessary to access all the time)

That's about it. Excellent work so far, it really doesn't seem far until the day we'll be able to finally use this feature in Pencil2D. 

Thanks! 😄

Jose-Moreno:
Ah well one thing I'd like to ask as a general curiosity and something that I've never understood, is why scrubbing backwards always gives me the preview of part of the waveform that actually comes after the current frame and if one keeps scrubbing the stream keeps offset until one starts scrubbing forwards.

I suppose it's because of how the buffer is done or something? I've always wondered if this after effect of scrubbing could be improved as it's kind of tedious, but i'm not even sure it can be.

@scribblemaniac @CandyFace While this is not a request for this PR, do you guys know why this could happen and what could be a good way to fix it? I mean it's not exclusive of this PR but a problem i've seen with just about every other implementation out there.

davidlamhauge:
I've made the changes you suggested @Jose-Moreno .
I must say that scrubbing sound with the mouse is no help for me. I'll use the shortcuts every time, and twice on Sundays, but now the user has the choice.
The sound scrub toggle is on the timeline, and the mSec range is altered to between 40 and 200 msec. Be aware that while one scrub is played, another cannot start.

Jose-Moreno:
@davidlamhauge That's excellent David. I think now that you mention it, in Flash the sound previews per frame were overlapped, maybe that's why it gave that cohesive sense when scrubbing.

That said I tested it a bit on discord having some people listen to the lipsync exercise, and I think it works as it is right now, though you still need to get used to it to properly discern the phonemes.

One thing I noticed is that when grabbing the playhead, moving the mouse cursor on top of the playhead, over the same frame would preview that frame everytime the cursor was moved on top. In general I think that works. If anyone has complaints we can sort them out afterwards.

I also loved that you put the actual speaker icon already. So far it seems to be working (it activated /deactivated the scrubbing properly).

I'll continue testing a bit later this week and let you know if i run into any additional issues. Thanks again! 🙌

davidlamhauge:
To be able to scrub on the timeline, I have to react on mousePress and mouseMove events, so if you move the mouse slowly across one frame, it will play that frames sound a number of times.
I tried to avoid it, but it yielded some errors that were impossible for me to get rid of.
And then again - why would you scrub slowly across one frame?

Jose-Moreno:
@davidlamhauge I don't. I was expecting the preview to occur only once per scrubbed frame and on the current frame when clicking on the playhead, but instead I noticed that clicking over the playhead active area and slightly moving the cursor inside the area would activate the preview repeatedly per mouse motion.

Just thought I would mention it so we could be aware, but to me it's not a problem as one would be preoccupied scrubbing back and forth through the timeline; it's not like it would inconvenience the user either.

davidlamhauge:
I've done slight changes, so now it only plays the scrub for each frame once.
I have experienced, that it sometimes plays the rest of the sound while scrubbing, even if that not should be possible. See if you can reproduce that.
Meanwhile, I'll look further into it.

CandyFace:
Is it ready to be reviewed?

davidlamhauge:
I'd say it's ready to be reviewed. I have no more wishes for the feature, but maybe we should ask @Jose-Moreno , when morning dawns in his place.
The only thing I've thought of, is that it would be fine to be able to write down what you hear, when you scrub the sound. The animator uses it for his animation, but I think that must come later - if ever.

Jose-Moreno:
Good day.

I've tested the latest changes made available recently under Windows 10 v1909 with a 5.14.1 Kit using the 64bits Release profile.

So far everything has worked as expected. I livestreamed a bit with the version from https://github.com/pencil2d/pencil/pull/1333/commits/ccc8723e2f5ca9e9f264b5bb2daf9ff44de45ef6 and found no major problems.

After this it was reported that there were issues on other systems, but despite testing it could be reproduced only under select conditions.

After continued testing and through the enhancements proposed by David and CandyFace so far the latest commit for this branch https://github.com/pencil2d/pencil/pull/1333/commits/331980c04af88d44a645aa828f4a951c10cb1182 the functionality works great.

I will note that this should be stress-tested on a wider scale, probably using a large file with a longer sound (5+ min) to weed out corner cases and potential contributions to crashing and file saving (which are major points paints in every version we've had), but other than this, from a user experience perspective, I believe it's ready to be used by more people.

core_lib/src/managers/playbackmanager.h
```diff
@@ -93,14 +102,19 @@ class PlaybackManager : public BaseManager
     int mFlipRollInterval = 100;
     int mFlipInbetweenInterval = 100;
     int mFlipRollMax = 5;
+    int mMsecSoundScrub = 100;
+    bool mSoundScrub = false;
 
+    SoundClip* clip = nullptr;
```
CandyFace, line 108: This is unused.

davidlamhauge: Removed...

CandyFace:
Alright I will merge this now, my inquire still stands that I think we should look into how we can make the compatibility, stability and consistency better across all three platforms.

it will be a known issue for now. 

Well done @davidlamhauge!

scribblemaniac:
@CandyFace FFmpeg is not really suitable as a player. I think that the best way forward to consistent cross-platform playback is to use wavs for everything. This means always using the movie audio import behind the scenes so we can continue to support other audio formats, and converting existing mp3 audio files in projects created with older versions of Pencil2D.

CandyFace:
Alright that's fair, yeah you're probably right about that, I just thought it would be worth mentioning, since FFmpeg is a multimedia library and I've seen other projects using it as such.

PR #1340, opened 2020-04-21
Preferencemanager updated when fps changes

davidlamhauge:
While working on the sound scrub feature, I accidentally found out, that the preference manager is not updated when the fps changes. The playback manager is updated, but the preference manager get the fps from the file that loads with Pencil2D, and is not updated until next time it is opened.
It has had no impact on Pencil2D, but it prevented me from an enhancement I wanted, and thus it was discovered.


CandyFace:
LGTM, merging.

Issue #1127, opened 2018-11-21
The 'install' target doesn't install anything


yurivict:
It is supposed to install files into ```$(DESTDIR)$(PREFIX)``` or ```$(INSTALL_ROOT)$(PREFIX)```.

Observed this problem while trying to create the FreeBSD port.

J5lx:
When I originally added install support I made it Linux-only because I
wasn't sure how much of it was truly *nix agnostic or whether part of it
might be Linux specific (see
https://github.com/pencil2d/pencil/pull/671#issuecomment-301484555) –
though looking back I have no idea which parts that would've been.
Still, while you're here would you mind taking a quick look at the linux
section towards the end of app/app.pro and check whether all of it
applies to FreeBSD as well? If that's the case we might as well turn it
into a unix:!macx section.

scribblemaniac:
@J5lx Beat me to it, but I'll post this anyway. We do use `$(INSTALL_ROOT)$(PREFIX)` with `PREFIX` being defined during `qmake` and `INSTALL_ROOT` during `make install`. It appears to be working fine in our travis builds if you want to use those for reference.

The code that @J5lx was talking about which you may have to modify for FreeBSD can be found [here](https://github.com/pencil2d/pencil/blob/b5ae22c8d1ed0fbe28c8b580406d028616f9efd6/app/app.pro#L138-L157).

yurivict:
@J5lx This linux section perfectly matches FreeBSD. You can just assume that it would match it for all ```unix``` systems, IMO.

J5lx:
Alright then. I’m still not entirely sure about mac support, though, so @scribblemaniac could you tell me:

- Are Desktop Entries, MIME packages and icons contained within icon themes (as detailed in their respective fdo specifications, which I note originate in X11-based software) useful on mac as well?
- Bash and Zsh completion need to be placed in specific paths. Are these [prefix]/share/bash-completion/completions and [prefix]/share/zsh/site-functions on mac as they are on Linux (and other unices I suppose) or does mac use other default paths?

scribblemaniac:
Well there is no such thing as desktop entries in mac. Mime types and icons are specified in the Info.plist file and the icons themselves are contained in the .app package. I don't think here would be any use for these on mac.

[...]

PR #1347, opened 2020-05-05
Add install support for other Unixes

J5lx:
This is an extremely trivial change, but I would nonetheless like a quick review of the macOS-specific part because I’m not an Apple user.

### DIFF
--- a/app/app.pro
+++ b/app/app.pro
@@ -164,7 +164,7 @@ win32 {
     RC_FILE = data/pencil2d.rc
 }
 
-linux {
+unix {
     target.path = $${PREFIX}/bin
 
     bashcompletion.files = data/pencil2d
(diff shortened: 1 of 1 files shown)


app/app.pro
```diff
@@ -164,7 +164,7 @@ win32 {
     RC_FILE = data/pencil2d.rc
 }
 
-linux {
+unix {
```
CandyFace, line 167: I'm fairly certain this should be
```suggestion
unix:!macx
```

Because Mac OS uses other conventions which are setup above.

J5lx: @scribblemaniac I thought so too, but scribblemaniac suggested doing it this way, see https://github.com/pencil2d/pencil/issues/1127#issuecomment-440585025.

CandyFace: Hmm.. ah maybe that's true... alright then, no further comments. Looks fine to me.

J5lx:
I got more reviews than I asked for and no unresolved discussions, so I’m merging this.

PR #1348, opened 2020-05-06
Appimage sound followup

scribblemaniac:
This is a followup to #1341 with the changes that I have recommended. The biggest changes are the refactoring of some of the code in main to PlatformHandler and adding the apetag gstreamer plugin. Hopefully this will fix the issue with some mp3s not playing :crossed_fingers:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,7 @@ addons:
 
 before_install:
   - echo "are changes related to source code?"
-  - bash $TRAVIS_BUILD_DIR/util/checkchanges.sh || test "${FORCE_NIGHTLY_UPLOAD}" = "yes" || travis_terminate 0;
+  - bash test "${FORCE_NIGHTLY_UPLOAD}" = "yes" || $TRAVIS_BUILD_DIR/util/checkchanges.sh || travis_terminate 0;
   - |
     if [ "$TRAVIS_OS_NAME" == "osx" ]; then
         brew update;
@@ -88,4 +88,4 @@ script:
   - "$TRAVIS_BUILD_DIR/build/tests/tests"
 
 after_success:
-  - '"${TRAVIS_BUILD_DIR}/util/after-build.sh"'
+  - "${TRAVIS_BUILD_DIR}/util/after-build.sh"
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,7 @@ addons:
 
 before_install:
   - echo "are changes related to source code?"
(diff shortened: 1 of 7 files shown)


app/src/main.cpp
```diff
@@ -343,15 +319,12 @@ int main(int argc, char* argv[])
     // Force dot separator on numbers because some localizations
     // uses comma as separator.
     std::setlocale(LC_NUMERIC, "en_US.UTF-8");
+
+    PlatformHandler::initialise();
+
```
J5lx, line 324: I’d prefer having this after the resource initialisation for core_lib, since technically PlatformHandler is from core_lib and might use its resources (even though it currently does not). That said, it’s a cosmetic thing so I won’t block the PR over it.

.travis.yml
```diff
@@ -50,7 +50,7 @@ addons:
 
 before_install:
   - echo "are changes related to source code?"
-  - bash $TRAVIS_BUILD_DIR/util/checkchanges.sh || test "${FORCE_NIGHTLY_UPLOAD}" = "yes" || travis_terminate 0;
+  - bash test "${FORCE_NIGHTLY_UPLOAD}" = "yes" || $TRAVIS_BUILD_DIR/util/checkchanges.sh || travis_terminate 0;
```
J5lx, line 53: To avoid a build failure, this needs to be
```suggestion
  - test "${FORCE_NIGHTLY_UPLOAD}" = "yes" || bash $TRAVIS_BUILD_DIR/util/checkchanges.sh || travis_terminate 0;
```

scribblemaniac: Oops :flushed:

app/src/main.cpp
scribblemaniac: Good point. I just put it above because I wasn't sure if the gstreamers stuff needed to be set up before core_lib. As long as it's not an issue, I agree it's best to be after.

scribblemaniac:
I have tested your build and it does play the audio files with APE tags successfully :slightly_smiling_face:  A music program I use stores metadata with APE tags so I have no shortage of files to test this with.

J5lx:
Actually, I just noticed that the Mac build failed. Looks like it fell through the cracks the first time because that bash issue made the build fail silently. I’ll take a closer look as soon as I find the time.

J5lx:
It was only a typo, already fixed it myself. I’m merging this now (build already passed on my fork).

PR #1351, opened 2020-05-11
Add an action to open Pencil2D's temporary directory

scribblemaniac:
Implemented as discussed. It is an action in the Help menu, and when clicked, it produces a warning message before opening the user directory. Works on Ubuntu, should be tested on other platforms.

### DIFF
--- a/app/src/actioncommands.cpp
+++ b/app/src/actioncommands.cpp
@@ -911,6 +911,16 @@ void ActionCommands::checkForUpdates()
     dialog.exec();
 }
 
+// This action is a temporary measure until we have an automated recover mechanism in place
+void ActionCommands::openTemporaryDirectory()
+{
+    int ret = QMessageBox::warning(mParent, tr("Warning"), tr("The temporary directory is meant to be used only by Pencil2D. Do not modify it unless you know what you are doing."), QMessageBox::Cancel, QMessageBox::Ok);
+    if (ret == QMessageBox::Ok)
+    {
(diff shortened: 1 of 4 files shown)


davidlamhauge:
I've tested it on Win10, and it works as expected.
I've also looked at the code and it looks good to me. Only thing is the 'Opening!' debug message in ActionCommands. I'd replace that line with '// TODO we must find a better solution for this'.
I'll leave the review to @CandyFace .

chchwy:
Sorry I missed the discussion. What's the purpose to provide this?

CandyFace:
@chchwy The reason for the implementation is that we often end up explaining how to get to the temporary workspace, when someone has lost their project through corruption or similar. When trying to recover the project we lead them here in any last resort to extract the last data that was saved to the temporary workspace. On mac OS you'll have to delve through "private/var" to find the folder, which is not somewhere one should be looking unless they know what they're doing... so instead of only providing the information necessary for them to get there, we provide a simple shortcut.

It's not meant to be a permanent solution, it should be replaced with a proper recovery mechanism in due time.

scribblemaniac:
The qDebug statement was left in unintentionally. It has been fixed now.

@CandyFace Summed it up fine. There's not really much to this, it's an action that takes you to the temporary directory. This will simplify providing support to users who want to try and recover their project after a crash. This has previously been particularly difficult on macOS where the directory is different for everyone, but even Windows users had difficulty navigating to a hidden directory. It was extensively discussed in #feature-requests so if you have some complaint about it, it has probably already been complained about.

PR #1352, opened 2020-05-17
1082 perspective grids

davidlamhauge:
This PR is a simple feature, that gives you drawing aid for 1, 2 or 3 point perspective drawing.
I don't think it is in our issue list, but I thought it could be a good thing to have, so I made it.
This video shows how it works.
https://youtu.be/PsKO0ASrN1M


Jose-Moreno:
Great work! I might need to test it myself. From the video I can only say that it looks and feel great

I would personally only have buttons that activate the actual mode / tool (perspective grid) instead of having several buttons for each individual function. This idea also applies for the camera overlays, however I haven't found a way to reduce the options themselves, so I have no proper proposal on how to change it, though I do feel that the angle menu could be a solution for this.

For example allowing a single overlay button to select between various other overlays and change the icon once selected, this also would work for the perspective grid button to let you select between 1, 2, 3, x points mode.

However to be honest the perspective grid behaves more like a tool than a display mode, due to its inherent interactivity, and if made a tool the angle selector could be part of the options panel for that tool freeing space in the display panel.

As you can see my comments are only related to the UI, I personally have no issues with the functionality shown, if anything i've always admired how the free windows application **Carapace** from an Epic Games developer, implemented these concepts with a decent UI / UX, here are some reference videos you can look at, but please note i'm not suggesting to copy it, it's more like an inspirational reference.

**Carapace : Placing Vanishing Points**
https://www.youtube.com/watch?v=YQ6CEMEDxqY

**Carapace : Using Trace Lines**
https://www.youtube.com/watch?v=2xMkpCPlDhk

**Carapace V2.0 : Source Code Clean Up**
_(He talks through the code a bit, download in the description)_
https://www.youtube.com/watch?v=kgwDpLWbyvU

chchwy:
Will be reviewing this soon.

core_lib/src/overlaypainter.cpp
```diff
@@ -0,0 +1,313 @@
...
+{
+    QRect rect = mOptions.mRect;
+
+    painter.save();
+    initPerspectivePainter(painter);
+    QPainter::RenderHints previous_renderhints = painter.renderHints();
+    painter.setRenderHint(QPainter::Antialiasing, false);
+
+    qreal degrees = static_cast<qreal>(mOptions.nOverlayAngle);
+    if (degrees == 7.0) { degrees = 7.5; }
```
chchwy, line 196: Any reason the degree cannot be 7.0?

chchwy:
I have a question. Is it a common case that an animator could enable 2 or more perspective grid at the same time? Just wonder if it can be a perspective "mode" that can be switched between.

chchwy:
I have looked through the code. It's generally good at the core logic part and some relatively minor issues. 

I wonder if we can improve the UI.

1. Adjusting angles by numbers rather than a fixed list?
2. 3 buttons so you can turn each one on/off individually, or having a "perspective mode" to switch between 3 modes?
3. Probably better icons to indicate the purpose clearly?
4. Add these options to the menu, too?

Jose-Moreno:
@chchwy 

> I have a question. Is it a common case that an animator could enable 2 or more perspective grid at the same time? Just wonder if it can be a perspective "mode" that can be switched between.

It depends. By traditional experience normally you had to use a reference grid called a field chart to determine distances and sometimes you could have an actual "pixel" grid. 

In the case of digital domains, the perspective grid should allow simultaneous vanishing points to allow the artist to construct up to 5-point perspectives used in curvilinear perspective or the so called "fish-eye" view. In that sense the answer is yes, you should be able to have multiple perspective grid planes active at the same time.

image
_image taken from graphicstackexchange.com_

Aside from the ability to have these guidelines "stacked" on top of each other there's also the important snapping functionality to assist in the drawing of difficult angles or radiating lines towards a vanishing point.

> I wonder if we can improve the UI.

I was hoping we could do that actually.

> 1. Adjusting angles by numbers rather than a fixed list?

I agree that this should be a single spinner box instead of a list

> 2. 3 buttons so you can turn each one on/off individually, or having a "perspective mode" to switch between 3 modes?

I think a mode switch would be better, but also we should consider there are at least 5-point perspectives as I mentioned above, so it should be scalable.

> 3. Probably better icons to indicate the purpose clearly?

I think we can review these once we zone in the modifications on the UI behavior itself

> 4. Add these options to the menu, too?

Hmm, the problem is that a perspective aid is more like a tool, instead I think we could consider making this a tool so we could use the tool options to control its various settings, and use the display panel to enable / disable visibility of the grids themselves.

regarding the widget itself I personally proposed to have a more WYSIWYG approach similar to the software CAPARACE along other references in discord through this link https://discord.com/channels/342369662710972417/402984217702563871/578299474267602978, but even if we reference Software like Adobe Illustrator's perspective helpers, it should be enough to have a cleaner UI.

image

chchwy:
Thanks for your comment @Jose-Moreno.
I am a bit confused about the "perspective tool". Is it like a "Pen tool" or "Hand tool" sort of things?
what should this tool do when a user pick this tool?

core_lib/src/overlaypainter.cpp
davidlamhauge: In theory - no.
For simplicity and ease of use, I only use angles that are divisors to 180. Right now 2, 3, 5, 7.5, 10, 15, 20 and 30 degrees. We could add 1, 4, 6, 9, 12 and 18.
I we allow the user to pick a random degree, let's say 9.1 degrees, we would have to draw 39.56 angles in 1 point perspective mode. How do we do that?
The same goes for 2 and 3 point perspective modes. When the numbers doesn't add up, it just complicates things.
If we should leave it to the user to decide, we could have a integer spinbox, where the user could enter the number of angles he wanted. If we go for this solution (that I'm no fan of), I think we should have a maximum of 360 angles in 1 point pespective and 180 angles in 2 and 3 point perspective.

davidlamhauge:
Thanks for your comments!
I see the perspective grids as no mora than an overlay, like the Center overlay, Golden Ratio overlay and so on.
As for having 5 point perspective - Well, if someone makes a PR to my branch, I can merge it, but I'm not going to code it myself.
I certainly count on someone (Candyface?) to redesign the buttons. At least, he should be consulted in the UI improvements you talk about.
Thanks again for your comments. I look forward to see it merged.

Jose-Moreno:
> Thanks for your comment @Jose-Moreno.
> I am a bit confused about the "perspective tool". Is it like a "Pen tool" or "Hand tool" sort of things?
> what should this tool do when a user pick this tool?

@chchwy Well the way David's implementation currently work is that when you activate the 1,2 or 3 point perspective visibility button, you will see that many points on the canvas. But you need 3 different buttons for 3 different points, so having a 5 point perspective would mean having 5 different buttons.

In the current implementation you can move these pre-existing points on the canvas normally once you activate the display buttons, but the idea of having a tool would be to centralize these options.

Additionally when people click on the canvas, these points should be generated with the specified amount of attributes that would appear in tool options such as the amount of lines according to the input angle among other things. And I would argue those points should be stacked on top of each other rather than having 3 to 5 separate "modes". This would contribute to simplify the UI and the user experience and to still have a deterministic system.

Jose-Moreno:
@davidlamhauge Hi David!

> I see the perspective grids as no mora than an overlay, like the Center overlay, Golden Ratio overlay and so on.

I agree they are an overlay, but I'm asking for a tool to allow the user to clearly manipulate these overlays at will. Currently the move tool serves this purpose, but scaling the overlays in functionality would mean that having perspective related options should also need to appear in the tool options (I'm talking about the grid lines angle selector, and perhaps color tinting options for the lines among other things), however in order to avoid clutter and to be direct I'm suggesting to have an additional tool that handles this, but that's a point of discussion, not something i'm asking to implement right now.

> As for having 5 point perspective - Well, if someone makes a PR to my branch, I can merge it, but I'm not going to code it myself.

Don't worry I was answering to Matt's questions, I didn't mean it for you to code the rest. In fact what you already did I believe is a great foundation that can be used to build the remainder on top of, at least in regards to functionality, our resident devs will let us surely know if there needs to be changes in the code, but even then it won't go away. 

We do need to keep the UI clear and flexible, so once we can agree to a UI improvement as well as a way for the functionality to scale in the future so it doesn't need full rewrites, i'm hopeful it will be merged. You know how it is every time we're discussing about these features, it's a bit of push and pull so we can provide a good experience for the majority😄

chchwy:
I watched **Carapace** and **FireAlpaca** videos, very interesting! I have some ideas about the new UI in my mind. 

Besides the UI part, I would like to merge this because of the Overlay manager, which can help the internal code structure and make the drawing process simpler. 

will make another PR soon for the perspective grid new UI. 
Thanks for your opinion @Jose-Moreno and Thanks for your work @davidlamhauge

PR #1353, opened 2020-05-20
Make shortcut names translatable

J5lx:
This change makes the shortcut names in the preferences translatable. The relevant changes are all in the first commit, the second one is only cosmetic.

Note, to see existing translations in action, it is necessary to update the translation files. I didn’t add that to this PR to avoid noise.


chchwy:
@J5lx I sent a PR to this branch at https://github.com/J5lx/pencil2d/pull/2

J5lx:
Sorry, that kinda flew under my radar because my email rules didn’t cover my fork. It’s merged now.

PR #1354, opened 2020-05-20
added invalid sound frame test

mc-csharpclass:
just added a new test to the end of the test_filemanager file.


chchwy:
Good job @mc-csharpclass

PR #1359, opened 2020-05-22
added test_colormanager.cpp

mc-csharpclass:
added test_colormanager.cpp and updated tests.pro to include the new test file


chchwy:
Good one.

PR #1364, opened 2020-06-01
issue #1254 vector eyedroppertool fix 

### LINKED ISSUE #1254: Vector color picker not working on line strokes
### Issue Summary

Candyface first noticed an issue today while we were discussing some features. It seems the vector layer color picker is not working for strokes (but it is working for fills) on 0.6.4 and it has not been working ever since the inception of mankind (or at least ever since pencil v0.4.4b)

### Expected Results
At a minimum and due to vector workflows the color picker should not only pick the color but select the current swatch that corresponds to that color in the palette for both stroke path and fill.

### Steps to reproduce

1. Select a color in the palette (e.g red)
2. Draw on a vector layer with said color using brush or pen. You can fill it too.
3. Select a different swatch to notice the change.
4. Color pick the stroke. Notice the color does not change.
5. Color pick the fill. Notice the color fill you picked changes appropriately in the palette.

### System Information
+ Pencil2D Version:
Version: 0.6.4
CPU Architecture: x86_64

mc-csharpclass:
updated eyedroppertool to work with vectors.  Added getCurvesColor to vectorImage.h/vectorImage.cpp to implement this fix.

closes #1254


CandyFace:
closes #1254

PR #1369, opened 2020-06-03
Make onion skin widget scrollable

CandyFace:
I recall seeing an issue about the new onion skin widget being too big and I agree that it is a problem, especially when working on a small monitor, so an immediate quick fix for the release could be to make it scrollable. 

scrollable

### DIFF
--- a/app/ui/onionskin.ui
+++ b/app/ui/onionskin.ui
@@ -13,7 +13,7 @@
   <property name="minimumSize">
    <size>
     <width>200</width>
-    <height>100</height>
+    <height>150</height>
    </size>
   </property>
   <property name="floating">
@@ -24,6 +24,18 @@
   </property>
   <widget class="QWidget" name="dockWidgetContents">
    <layout class="QVBoxLayout" name="verticalLayout">
+    <property name="leftMargin">
+     <number>0</number>
+    </property>
+    <property name="topMargin">
+     <number>0</number>
+    </property>
+    <property name="rightMargin">
+     <number>0</number>
+    </property>
+    <property name="bottomMargin">
+     <number>0</number>
+    </property>
     <item>
      <widget class="QScrollArea" name="scrollArea">
       <property name="sizePolicy">
@@ -46,8 +58,8 @@
         <rect>
          <x>0</x>
          <y>0</y>
-         <width>215</width>
-         <height>331</height>
+         <width>233</width>
+         <height>349</height>
         </rect>
        </property>
        <layout class="QVBoxLayout" name="verticalLayout_3">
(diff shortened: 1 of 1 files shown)


scribblemaniac:
This is an alright solution for now. Personally I wanted to change the default layout, but I could see that not fixing the issue for everyone. I have not tested this yet, but based on the preview, can you remove the margin between the widget and the scroll area?

CandyFace:
Right, yes should be done now.
image

chchwy:
I moved the min% and max% into one row. Does anyone think it helps?

onions01

I can do further as well by moving the prev frames & next frames into one row:
onions02

scribblemaniac:
I like that, but my only concern would be what affect it has on the minimum width of the widget. If it could split onto separate lines if the width was small enough, that would be nice.

chchwy:
Then it's called responsive layout, which doesn't really exist in Qt layout system.

Here is how it looks like when I minimize the side widgets:
onions03

I suppose it's fine as today most people use widescreen monitors. Vertical spaces are more precious than horizontal space, yea?

CandyFace:
I prefer the previous layout as both vertical and horizontal space matters and it fits in between.
The min width is already pretty wide, so I'd prefer not having it wider.

chchwy:
Alright, I think the fact is there are too many options in this widget. So the widget just can't fit in a small space.

So either
1. we reduce the number of options (or make a show/hide toggle)
2. or we allow it uses more space (like make it wider or scrollable)

Why I said the vertical space is precious is because we already use up all the vertical space.

scribblemaniac:
> Then it's called responsive layout, which doesn't really exist in Qt layout system.

Qt is definitely capable of this, look at how the toolbox behaves now. All that that should need to be done is to put two QHBoxLayouts in a FlowLayout. I think that would compromise between horizontal and vertical space.

chchwy:
@scribblemaniac Oh thanks. I checked the `flowlayout` again. Seems Qt has fixed an issue that `flowlayout` couldn't handle minimum height very well when I integrated it previously. Yeah, now I think it's definitely doable.

chchwy:
Just wonder why Qt doesn't want to add flowlayout into Qt SDK and leave it as an official *Example*. :/

CandyFace:
I think the solution to make it scrollable is fine, as it will also happen to other widgets in the future... eg. in the mypaint branch the tooloptions widget is scrollable too. 200-300 tall widgets is not a good idea if it can be avoided. There are certain widgets that shouldn't be scrollable though... the tool widget, color inspector and color wheel to name a few.

chchwy:
Ok, Let's merge it for a temporary solution.

PR #1371, opened 2020-06-04
Checking for nullptr to prevent crash with Ctrl+A and Ctrl+D

### LINKED ISSUE #834: Segfault with the move tool
### --Issue Summary--
Pencil2d crashes with a segmentation fault when clicking on the canvas under certain circumstances.

### --Steps to Reproduce--
1. With a bitmap layer selected, create a new keyframe
2. Go to the first frame and delete it.
3. Click on the canvas with the move tool

### --System Information--
+ Pencil2D Version: 0.6.0

+ Operating System: macOS 10.13.1

+ RAM Size: 16 GB

### --For Code Contributors--
The dispatch queue when the crash occurred was:
```
BitmapImage::copy(QRect)
BitmapImage::transformed(QRect, QTransform, bool)
ScribbleArea::applyTransformedSelection()
TimeLineCells::mousePressEvent(QMouseEvent*)
```

davidlamhauge:
Pencil2D was crashing if there was no frame on frame 1, and you tried to Ctrl+A on that frame.
Checking for nullPtr fixes that.

Does this close #834?

### DIFF
--- a/core_lib/src/interface/editor.cpp
+++ b/core_lib/src/interface/editor.cpp
@@ -958,11 +958,15 @@ void Editor::selectAll()
         // Selects the drawn area (bigger or smaller than the screen). It may be more accurate to select all this way
         // as the drawing area is not limited
         BitmapImage *bitmapImage = static_cast<LayerBitmap*>(layer)->getLastBitmapImageAtFrame(mFrame);
+        if (bitmapImage == nullptr) { return; }
+
         rect = bitmapImage->bounds();
     }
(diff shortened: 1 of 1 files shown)


scribblemaniac:
This looks fine. There are a couple places where it would be better to run the code that uses the BitmapImage/VectorImage in an if statement it is not a nullptr rather than returning the whole function as it skips over other code. I will fix this and add more fixes for issues similar to the ones fixed here soon. I will merge this for now.

PR #1390, opened 2020-06-20
Check for vector layer, before toggle showInvisibleLines

davidlamhauge:
When drawing, the setting INVISIBLE_LINES was set, no matter what type layer was drawn onto.
Now it checks for Layer::VECTOR


chchwy:
Reviewed, Thank you @davidlamhauge

PR #1395, opened 2020-07-10
Position new palette colors at bottom, and renaming immediately

davidlamhauge:
This fixes a 'bug' that has often bothered me.
1) New colors were placed in front of the selected color, and that made sense before the colors could be dragged and dropped. I always expect to find the new colors at the bottom of the palette, because they logically should be appended to the palette. This PR places new colors at the bottom of the list.
2) The reason for having names for the colors is to use them. If we don't use the names, we might as well just have them numbered. My colors get names like 'Lisa skin', 'Lisa eyes', 'Lisa Hair' etc, and not 'Magenta', 'Vivid Reddish Purple', or some of the other suggested names. For that reason, I've put the rename color dialog up front, so you're naming your color when you add it. This is more logical, and if you indeed want the name to be 'Vivid Reddish Purple', then just press enter.

### DIFF
--- a/app/src/colorpalettewidget.cpp
+++ b/app/src/colorpalettewidget.cpp
@@ -139,13 +139,6 @@ void ColorPaletteWidget::addItem()
     ColorRef ref(newColor);
 
     mObject->addColorAtIndex(colorIndex, ref);
-    refreshColorList();
-
-    if (mFitSwatches)
-    {
-        fitSwatchSize();
-    }
-
     bool ok;
     QString text = QInputDialog::getText(this,
                                          tr("Color name"),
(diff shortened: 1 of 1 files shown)


chchwy:
I made a change for this PR.

Rather than popping up a rename dialog. I just make the new color editable after clicking "add".
image

What do you think? @davidlamhauge @CandyFace

davidlamhauge:
Great idea. Sorry I didn't think of that.

CandyFace:
That's exactly how I would improve it based on my comment in Discord made the other day, I approve of such a change.

chchwy:
@davidlamhauge I made a PR in your repo
https://github.com/davidlamhauge/pencil/pull/13

scribblemaniac:
Do both views scroll down to the new color when added? If not, they should, especially the list view so you can see the newly added/editable color.

chchwy:
Yes it does scroll down to the new color.

PR #1397, opened 2020-07-17
Abort fill with a message, if fill extends mBounds

davidlamhauge:
I've been wondering why we let the floodFill fill the entire cameraRect, if there is a hole in the line/area you want to fill.
This PR cancels the fill process, if P2D fills beyond the mBounds, and tells you what to do to fix it.
In the picture you see two heads. In the right the line is closed, and it is filled as it should, but in the left, it extends the bounds, aborts, and tell me what to do. (You can see the small crosshair, where I pressed to fill)
fill_warning
I know some of you don't like message boxes, but in stead of filling all over the place, this solution aborts the filling, and tell you in a nice way, how you solve the problem, and get on with your life. Since you, in the old way, had to undo the fill, before you could repair the line, it's not more work with this PR, and it is more pleasent.


CandyFace:
I don't think this is the right way to go about the problem, it reminds me of the very old pencil2D where you would get a similar message because the algorithm couldn't fill because of a gap that was seemingly nowhere. This happened only for vector though but it was imo. still not a great solution.

The problem with the similarity here is that if the gap is only 2 pixels wide, then you're forced to figure out where the gap is before you can fill, that's bad ux imo. It's easier to undo and fill again if you see the spilling over. The spilling might even help you locate the error.

We want to fill within a given gap threshold, not cancel the whole fill operation.

davidlamhauge:
With the many minutes of real animation I've done this year, and the literally hundreds of times I've seen paint spilling over and fill my screen, my experience is that it's _not_ easier to undo and fill again. It's a PITA.
About filling within a given gap threshold, please be my guest. I saw how GIMP had done it last year at LibreGraphics Meeting, and it was a mess. I tried my own implementation when I came home from LGM, and that made me realize why it was a mess. From what I've seen, it makes more problems than it solves.

CandyFace:
Fill within a threshold can be done, Clip Studio Paint does it very well, although the implementation is not available so one can only guess how they've done it. Whether it's a mess technically is imo. not an issue as long as the user experience is smooth. 

csp

I'm not saying it's easy to implement, I have tried once and failed to do so but it can be done without making the process become a challenge. As certain as I am with people having issues with the obnoxious "remember to save" dialog, this too will cause an endless amount of "how do I turn this off, I just want to fill", especially if the user can't see why or where it goes wrong, they can only guess and try again until the popup disappears.

Message dialogs are for when things go wrong or there's something you must take action to solve but this is not a bug nor a problem, this is the expectation of the user being different than the result of the algorithm.

scribblemaniac:
Sorry for letting this sit for so long. I agree with @CandyFace, this is not a the right solution. The way I look at it is that when using the fill tool, you are either doing one of two things:
1. You're filling some enclosed object (like the first example).
2. You're filling the background.

Now compare what we currently do to this new the dialog approach for each of these scenarios. In scenario 1, you go from having to undo the fill (one keyboard shortcut) to having to close a dialog (one mouse click). To me, we don't really gain anything from that change, and I personally feel that dialogs would be far more of a nuisance than undoing. And in the second fill scenario, you go from it working exactly how you want it to, to it not working at all and a seemingly irrelevant dialog appearing. That's not good.

I'm am all for good gap filling algorithm option for the fill tool, but this is not that. I move that we close this PR. Thoughts @CandyFace @davidlamhauge?

davidlamhauge:
I use Pencil2D for real animation, and last spring I made two short films, with a running time of 4-5 minutes. In such a process it is really nagging to see the color spill all over maybe 50 times a day, and every time it is like the software says "Ha! Looser. Gotcha!". It wears you down, and sometimes I lost my patience, and had to walk around a minute or two to calm down, before I could continue - and I am normally a very patient man.
I can understand that most of you don't want the feature, but nevertheless I want to ask, if it could be a checkbox in the preferences, so you could activate the feature if you want it?

CandyFace:
But for every time you fill, you can simply undo and try again, is that really a huge problem? Why do you think it's better to be notified of the mistake without being able to get a hint of where the gap is? The only difference is that you don't get to see where it tries to fill but is simply being told you're not there yet. That's not useful if you ask me.

I do not think it should be toggle for those reasons, I think we should try to deal with problem by trying to fix it, not throw dialogs at the user and tell them they're doing it wrong.

I too think we should close this issue and then make a new discussion of how we can solve the gap problem.. I'm sure we can come up with a few ideas that could be tried out.

davidlamhauge:
I'd say that it's a huge problem, when you get so mad that you want to smash your 400$ xp-pen drawing tablet.
But - this is too much uphill. I'll close it, and integrate in my own private Pencil2D version instead.

PR #1400, opened 2020-07-21
Draw the timeline using palette colours (AKA dark theme support for the timeline)

J5lx:
In order to make some progress on #550, I modified the timeline rendering to use colours from the application palette instead of hardcoded values. With this change, the only remaining dark theme issues are those related to the icons.

Here is a before/after comparison using a random dark theme:

Before

After

In addition to the colour changes, I also simplified the design a little in order to accommodate the somewhat limited selection of colours available from QPalette and to make sure the timeline continues to look good regardless of which theme is used. I tested this with a wide range of different themes, both light and dark, and I’m happy with the results, but if you find a theme that “breaks” the timeline, let me know.

Grab a build here: [Linux](https://drive.google.com/file/d/1ykvkfqN0ikptV6fgY3hgK2BFLMpcANwp/view?usp=sharing) [macOS](https://drive.google.com/file/d/1LBQAJw5ff1aghMTgeMyKWd9rD3qMjJoQ/view?usp=sharing) [Windows x86_64](https://drive.google.com/file/d/15ZIy5YvM281UNQgtowb9WBVeuv5ATamQ/view?usp=sharing) [Windows x86](https://drive.google.com/file/d/1oooXl-P3NiAZyhC8qhy2VwarYsEHx1Bh/view?usp=sharing)

### DIFF
--- a/core_lib/src/interface/timelinecells.cpp
+++ b/core_lib/src/interface/timelinecells.cpp
@@ -17,6 +17,7 @@ GNU General Public License for more details.
 
 #include "timelinecells.h"
 
+#include <QApplication>
 #include <QResizeEvent>
 #include <QMouseEvent>
 #include <QInputDialog>
@@ -172,6 +173,8 @@ bool TimeLineCells::didDetatchLayer() {
 
 void TimeLineCells::drawContent()
 {
+    const QPalette palette = QApplication::palette();
+
     if (mCache == nullptr)
     {
         mCache = new QPixmap(size());
@@ -195,7 +198,7 @@ void TimeLineCells::drawContent()
 
     // grey background of the view
     painter.setPen(Qt::NoPen);
-    painter.setBrush(Qt::lightGray);
+    painter.setBrush(palette.color(QPalette::Base));
     painter.drawRect(QRect(0, 0, width(), height()));
 
     // --- draw layers of the current object
@@ -229,7 +232,7 @@ void TimeLineCells::drawContent()
     {
         if (mType == TIMELINE_CELL_TYPE::Tracks)
         {
-            layer->paintTrack(painter, this, 
+            layer->paintTrack(painter, this,
                               mOffsetX, getLayerY(mEditor->layers()->currentLayerIndex()) + getMouseMoveY(),
                               width() - mOffsetX, getLayerHeight(),
                               true, mFrameSize);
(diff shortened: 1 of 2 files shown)


chchwy:
Awesome!

CandyFace:
Nice work.

Simplifying the colors would be okay if they worked but imo. it doesn't quite work.

The background color is fine but the rest I think requires more refinement.

Specifically I don't think using qpalette colors works for the scrubber, as I don't see it as a highlighter in the same way a layer is highlighted as such, I'd rather keep the red color and then find a more appropriate red if it doesn't work for dark themes but maybe that's just me.

The visibility dots gets difficult to see and differentiate, they're practically invisible when the layer is hidden.
image

The background line splitters are almost gone when the layers are hidden
image

image

I am all in for proper dark mode support but I don't think simplifying to qpalette colors with the limited selection we have currently solves the problem well enough to use it for all of it.

J5lx:
Good points, thanks for giving it a try. I guess the scrubber is more of a preference thing and since it’s not hard to go back to red, I’ll just wait for what the others think about it for now. I can see the issue with the visibility dots, though, and the contrast of the “line splitters” is indeed a real problem.

Unfortunately it’s not as easy as just picking some custom colours for that; there’s a reason why I mostly just stuck with the QPalette colours, despite how limited they are. As soon as you start throwing custom colours into the mix or modifying the palette colours, it’s really easy to get into situations where it works with dark themes but not light ones, or it looks jarring with some themes (especially here on Linux where themes can be not just light or dark but also somewhat colorful), etc. Hence I kinda feel like sticking to the palette colours is actually the lesser evil in a way. That said, it’s time for some more experimentation! I have some ideas already…

J5lx:
Okay, I think I got something that works. Here are the relative layer visibility dots:

Screenshot_20200721_223150

Here are some hidden layers:

Screenshot_20200721_223302

I’m quite happy with this, but it’s pretty trivial to tweak the amount of dimming if necessary.

Builds: [Linux](https://drive.google.com/file/d/1cN4zA_IjK8gVFFKnGMa3mD1TfnC8GdTh/view?usp=sharing) [macOS](https://drive.google.com/file/d/1-dZJmWobBBcEKS_LKvPzcmUkb8_M14yp/view?usp=sharing) [Windows x86_64]() [Windows x86](https://drive.google.com/file/d/1A4sEEnAIRoMqcl0d0LbJxPsal-PhUtsO/view?usp=sharing)

I also gave the red scrub another shot, but IMO it just doesn’t look that good with some of the more colourful themes, so I’m still in favour of the highlight colour (current state of this PR) there.

CandyFace:
Yes that's better, I like the way you went about fixing the line splitting 👍 
Although from a technical pow, I find it a bit hacky to use "Text" for the palette color but whatever... I guess that's fine for now.

I did stumble upon a bug though... it seems you change the visibility mode when the global setting is set to hidden, yet you can't see the change on the dot... I'm not sure what would be correct here, since the global mode is technically "solo", so should all layers except for the layer you stand on be shown as hidden or should we simply ignore the button click on layers that are not the currently selected in this particular mode?

J5lx:
Using the text colour isn’t even *that* hacky… the documentation actually describes as a “foreground color [which] must provide good contrast with Window and Base”, and that’s exactly what’s needed here ;)

I’m not 100% sure which bug you mean there, but I think it’s been there even before this PR, no? I agree that the behaviour is kinda weird but tbh I’m not sure if there’s a way to make it truly intuitive :thinking: Might have to think about it some more…

CandyFace:
Ah sorry seems like I removed the part where I mentioned that the bug existed before the PR. Yes maybe it's not that important, if we're going to redesign the timeline anyway.

Ah okay if the documentation refers to it too, then that's fine with me.

J5lx:
Are we still waiting for something here or is this PR ready to be merged?

CandyFace:
Not that I know? Should I just merge?

J5lx:
Considering no one else voiced any concerns in the last 1½ months, I guess it should be fine.

CandyFace:
You're right, merged :)

J5lx:
Thanks!

chchwy:
timeline-win

Just tested this with Windows 10 light theme.

Timeline background is plain white. I expect it to be the light grey colour like other widgets. Maybe need some OS-specific settings?

J5lx:
Hmm, personally I think the white background actually kinda makes sense because that’s also what the list view in the Colour Palette uses, and the timeline is kind of a list view itself. And if you look at my original screenshot in the OP you’ll see the colours are matching there as well (although the actual colours are different). I tried changing the background colour to that background grey (in GIMP, no need to boot up Windows that way) but I’m not quite convinced by the result tbh:

Screenshot from above modified with a light grey background for the timeline

But in the end I’m not exactly an everyday Windows user, so if you think it looks better in grey then let’s do it that way.

PR #1404, opened 2020-07-24
Handle angle-based high resolution scrolling properly

J5lx:
For zooming, Pencil2D currently handles all wheel events that are not pixel-based as if they had the same resolution. This leads to sub-par zooming for devices which report high resolution wheel data. With this PR, I fixed that by taking the same approach already used for pixel-based scrolling and tweaking it so that the algorithm satisfies a few additional properties:

- After first scrolling forward by X degrees, scrolling backward by the same X degrees returns the zoom to its original state
- Scrolling forward by X degrees, then scrolling forward by another Y degrees is the same as scrolling forward by (X + Y) degrees
- Using a mouse with "standard" 15 degree resolution, it’s possible to go back to exactly 100% from min/max zoom without having to use "Reset zoom/rotate"
- Using a mouse with "standard" 15 degree resolution, the zoom sensitivity feels roughly the same as before (100x zoom per 180 degree rotation to be precise; though since it wasn’t linear before there’s no direct comparison)

It would probably be a good idea to modify the pixel-based code like that as well (at least when it comes to the first two bullet points), but unfortunately I don’t have the hardware and/or software to test that code, hence why I didn’t do that myself. If anyone has the means to test that (most likely MacBook users) and wants to give it a shot, feel free to commit directly to this PR.

The drawback of this change is that scrolling now works differently for mouse wheel and keyboard, but since this was already true for pixel-based scrolling anyway I don’t think it’s a problem.

[Here](https://streamable.com/rncav5) is an example of how high-resolution scrolling looks without this change. It’s choppy, and even though I’m only making very small movements it goes almost all the way from min zoom to max zoom and back. [Here](https://streamable.com/luklzm) is how it looks with this PR. The zoom doesn’t go crazy fast anymore and is also much smoother (though the video recording/encoding seems to have messed it up a little).

Builds: [Linux](https://drive.google.com/file/d/14NP6fnDh1Y615xQyeaRtvB5yrK3fErXh/view?usp=sharing) [macOS](https://drive.google.com/file/d/1F3YZJ9MV7FpWaRgLKcIX2UnMwqWRYUej/view?usp=sharing) [Windows x86_64](https://drive.google.com/file/d/1zuSirdjUJVZqg2aEKVSrrc7SFYHkqwjH/view?usp=sharing) [Windows x86](https://drive.google.com/file/d/1dSfxScxOPGSQ4_WQCV32nOXQFcCRaxuU/view?usp=sharing)

### DIFF
--- a/core_lib/src/interface/scribblearea.cpp
+++ b/core_lib/src/interface/scribblearea.cpp
@@ -398,11 +398,14 @@ void ScribbleArea::wheelEvent(QWheelEvent* event)
 
     const QPoint pixels = event->pixelDelta();
     const QPoint angle = event->angleDelta();
-    const QPointF offset = mEditor->view()->mapScreenToCanvas(event->position());
+    const QPointF offset = mEditor->view()->mapScreenToCanvas(event->posF());
 
     const qreal currentScale = mEditor->view()->scaling();
     if (!pixels.isNull())
     {
+        // XXX: This pixel-based zooming algorithm currently has some shortcomings compared to the angle-based one:
+        //      Zooming in is faster than zooming out and scrolling twice with delta x yields different zoom than
+        //      scrolling once with delta 2x. Someone with the ability to test this code might want to "upgrade" it.
         const int delta = pixels.y();
         const qreal newScale = currentScale * (1 + (delta * 0.01));
         mEditor->view()->scale(newScale, offset);
--- a/core_lib/src/interface/scribblearea.cpp
+++ b/core_lib/src/interface/scribblearea.cpp
@@ -398,20 +398,21 @@ void ScribbleArea::wheelEvent(QWheelEvent* event)
 
     const QPoint pixels = event->pixelDelta();
     const QPoint angle = event->angleDelta();
+    const QPointF offset = mEditor->view()->mapScreenToCanvas(event->position());
 
     const float currentScale = mEditor->view()->scaling();
     if (!pixels.isNull())
     {
         const auto delta = static_cast<float>(pixels.y());
         const float newScale = currentScale * (1.f + (delta * 0.01f));
-        mEditor->view()->scale(newScale);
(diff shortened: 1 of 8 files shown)


J5lx:
I just added another related change so that when using a pointing device to zoom, the canvas will now zoom towards the cursor rather than the center, like it does in other graphics applications as well.

J5lx:
Build failure is fixed.

core_lib/src/interface/scribblearea.cpp
```diff
@@ -398,26 +398,29 @@ void ScribbleArea::wheelEvent(QWheelEvent* event)
...
     const QPoint angle = event->angleDelta();
-    //qDebug() <<"angle"<<angle<<"pixels"<<pixels;
-
+    #if (QT_VERSION >= QT_VERSION_CHECK(5, 14, 0))
+    const QPointF offset = mEditor->view()->mapScreenToCanvas(event->position());
+    #else
+    // This is actually not the same as the above, hence why we use it only if necessary
+    const QPointF offset = mEditor->view()->mapScreenToCanvas(mapFromGlobal(QCursor::pos()));
+    #endif
+
```
CandyFace, line 407: The comment mentions that these doesn't give the same outcome, but as far as I can see they do... at least using my trackpad scroll? 

is there any particular reason why you use ```mapFromGlobal(QCursor::pos()))``` 
instead of event->posF() or event->pos()?

J5lx: THANK YOU event->posF() is exactly what I actually wanted but I just couldn’t find it because it is deprecated and hidden in the documentation 🤦. I’ve now removed the version check altogether like we did with the similarly deprecated byteCount() in MovieExporter.

chchwy:
@J5lx I made a PR to this branch at https://github.com/J5lx/pencil2d/pull/3

PR #1406, opened 2020-07-29
Pencil2D 0.6.5 Release Changelog [WIP]

Jose-Moreno:
Since Pencil2D last stable release was deployed over a year, due to the huge amount of commits the categorization style has not yet been implemented and instead it's being listed in chronological order from oldest to latest commit. 

- There are some commits that are difficult to describe as they contain several fixes of heterogeneous nature and have retained their merge-to-master automated title descriptions, but this will require some discussion as to how we want them represented.

- I've also attempted to directly link each and every commit via hyperlink, but the commits from the last month have not been updated and thus I'm pending finishing this task.

Please let me know how you would suggest categorization or any changes and i'll make them as soon as possible.


Jose-Moreno:
Closing due to force-push. Now I have to delete all my branches otherwise I'll be bringing unwanted baggage into master.

PR #1408, opened 2020-07-30
Pencil2D 0.6.5 Release Changelog [WIP]

Jose-Moreno:
Since Pencil2D last stable release was deployed over a year, due to the huge amount of commits the categorization style has not yet been implemented and instead it's being listed in chronological order from oldest to latest commit.

- There are some commits that are difficult to describe as they contain several fixes, patches and commits of heterogeneous nature, as such when they merge into master they retain the automated name of their branch as descriptions. Though In general I've added extra description where I saw prudent, I left it mostly as is.

- I've also attempted to directly link each and every commit via hyperlink, but the commits from the last month have not been updated and thus I'm pending finishing this task.

- I've trimmed many commits that were incorrectly added and only left those that were commited to master.

- I've categorized the commits following previous categories and added some new ones like **Continuous Integration** and **Tests**

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,3 +1,221 @@
+## Pencil2D 0.6.5 - 31 July 2019
+
+### Features
+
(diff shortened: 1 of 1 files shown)


chchwy:
Merged.

PR #1419, opened 2020-08-10
New layer naming #1414

### LINKED ISSUE #1414: [ISSUE] When first adding a new layer the system will add a wrong name
### Issue Summary
as example, there is NO sound layer on the timeline:
When importing an audio (File > Import > Sound...) we are asked to create a new sound layer. The windows property's panel will pop-up showing **Sound Layer** as its default name. After hitting OK, the system will add a sound layer with the name **Sound Layer 2**

### Actual Results

It is added a wrong default name

### Steps to reproduce
File > Import > Sound > Let its default name **Sound Layer** on the property panel and hit OK  
Then select a WAV sound track and hit enter. The layer will be created with the name **Sound Layer 2**

NOTE: this issue will also happen when creating a new bitmap or vector layer instead

### System Information
+ Pencil2D Version:
last version (v0.65) 

+ Operating System:
Windows 10 

+ RAM Size:
32GB

+ Graphics Tablet:
Huion HD 185

davidlamhauge:
A new sound layer was named 'Sound layer 2', even if one selected 'Sound layer'. The reason was that the nameSuggestLayer() function was called twice.
Fixes #1414

### DIFF
--- a/app/src/actioncommands.cpp
+++ b/app/src/actioncommands.cpp
@@ -186,7 +186,7 @@ Status ActionCommands::importSound()
         bool ok = false;
         QString strLayerName = QInputDialog::getText(mParent, tr("Layer Properties", "Dialog title on creating a sound layer"),
                                                      tr("Layer name:"), QLineEdit::Normal,
-                                                     tr("Sound Layer", "Default name on creating a sound layer"), &ok);
+                                                     mEditor->layers()->nameSuggestLayer(tr("Sound Layer", "Default name on creating a sound layer")), &ok);
         if (ok && !strLayerName.isEmpty())
         {
             Layer* newLayer = mEditor->layers()->createSoundLayer(strLayerName);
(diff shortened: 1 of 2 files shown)


davidlamhauge:
I think we should avoid layer names with the same name, but when I suggested it once, the argument was that if you have a bitmap layer called 'Hero', and you wanted to clean it up on a vector layer, it would be logical to have the vector layer called 'Hero' too.
I will at any time support a PR that prevents identical layer names.

chchwy:
Thanks @davidlamhauge. I confirmed it works, too.

For the issue of same name layer, I just checked three painting software (Photoshop, Krita and Paint.NET), they all allow multiple layers with the same name. I think it depends on whether we treat the layer name as a unique identifier or just a name displayed on UI. Internally, layer id should be used instead of layer name to avoid duplicated names. The camera selection process can be improved by displaying layer indexes plus names like `1: my camera`, `2: my 2nd camera`. (I made the problematic camera selection dropdowns ;) haha)

Personally I prefer to leave the layer name to the users so they can use whatever they want. We can suggest a unique name like what we've done in this PR, but don't have to force it.

PR #1420, opened 2020-08-10
Only creating keyframes that exists in *.pclx

### LINKED ISSUE #1372: Unwanted keyframe added, when opening file
### Issue Summary
It is perfectly legal to delete keyframe 1 on a layer, as long as minimum one keyframe exists on that layer.
 

### Actual Results
If I have a saved scene, where frame 1 is deleted, and I open it again, then Pencil2D re-instates frame 1, even if it not exists.

 

### Expected Results
I would expect that Pencil2D opens my file with only the existing keyframes
 

 

### Steps to reproduce
1) Open Pencil2D
2) Scrub to frame 5 on the bitmap layer, make a keyframe and draw something
3) Scrub to frame 1 and delete the frame
4) Save the file
5) Open another file
6) Re-open the first file, and you will notice that Pencil2D has re-instated the deleted frame 1

 

### System Information
+ Pencil2D Version:
Newest

davidlamhauge:
If you, for some reason, delete frame 1 on some layer, it doesn't appear in the *.pclx file. Nonetheless, Pencil2D uses to reinstate frame 1, when reloading the file.
This PR removes these non-existing frames.
If you can think of a smarter way to implement it, please come forward.
fixes #1372

### DIFF
--- a/core_lib/src/structure/layerbitmap.cpp
+++ b/core_lib/src/structure/layerbitmap.cpp
@@ -184,6 +184,7 @@ void LayerBitmap::loadDomElement(const QDomElement& element, QString dataDirPath
 {
     this->loadBaseDomElement(element);
 
+    QList<int> frameList;
     QDomNode imageTag = element.firstChild();
     while (!imageTag.isNull())
     {
@@ -199,10 +200,13 @@ void LayerBitmap::loadDomElement(const QDomElement& element, QString dataDirPath
                 int x = imageElement.attribute("topLeftX").toInt();
                 int y = imageElement.attribute("topLeftY").toInt();
                 loadImageAtFrame(path, QPoint(x, y), position);
+                frameList.append(position);
 
                 progressStep();
             }
         }
         imageTag = imageTag.nextSibling();
     }
+    if (!frameList.contains(1))
+        removeKeyFrame(1);
 }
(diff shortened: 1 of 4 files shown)


scribblemaniac:
Finding out why it is added in the first place and correcting that would be a better way to fix this. What have you found out so far about the cause of this issue?

davidlamhauge:
When Penci2D reads the object elements in the xml, it finds the layer type, and adds a layer of that type, and then adds the keyframes that are in the xml.
The problem is that when it adds the layer, it creates a keyframe on frame 1. If there is no keyframe 1 in the xml, the keyframe 1 still is there on the timeline, and that is a bug in my world.
I could only see two solutions:
1) Not to add the first keyframe, when the layer is created, and somehow add the mandatory keyframe later.
2) Check the xml for a keyframe 1, and remove the first keyframe if it wasn't present in the xml.

I was afraid that the first solution would give problems, since we at all times need at least one keyframe on each layer, so I chose the second solution, but maybe solution 1 is the way to go, or maybe there is a third way?

scribblemaniac:
The second way works clearly, but I think the first way would be better. When you call `new BitmapLayer` or whatever layer type, it should create a completely empty layer in my opinion. Then the initial frame can be added to the new layer action code and the default startup template code. There's already no first frame for the sound layer, so it must be already doing something special there.

davidlamhauge:
`Then the initial frame can be added to the new layer action code and the default startup template code. `
I'm not sure what you mean here, since it is in the 'new layer action code', that the first frame is created now.
I've tried to remove it and add it in the createDefaultLayers() function, but then it goes wrong when you add a new layer, that will be created without an initial keyframe.
I'm inclined to favor solution 2...

davidlamhauge:
Replaced by #1422

PR #1422, opened 2020-08-11
Only creating keyframes that exists in *.pclx: The Sequel

### LINKED ISSUE #1372: Unwanted keyframe added, when opening file

scribblemaniac:
This is an alternative solution to the one provided in #1420. Fixes #1372.

### DIFF
--- a/core_lib/src/structure/object.cpp
+++ b/core_lib/src/structure/object.cpp
@@ -87,7 +87,6 @@ bool Object::loadXML(QDomElement docElem, ProgressCallback progressForward)
     {
         return false;
     }
-    int layerNumber = -1;
 
     const QString dataDirPath = mDataDirPath;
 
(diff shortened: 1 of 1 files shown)


scribblemaniac:
@davidlamhauge The code is not that complex. Basically the addNew*Layer functions do three things:
1. Create a new layer object of the specified type.
2. Add that layer to the mLayers list
3. Add a frame at the beginning.

Currently all ways of creating layers (shortcuts, buttons, and file loading) used these functions. All I did was stop calling those functions during file loading, and instead run the code for just steps 1 and 2.

davidlamhauge:
Yes @scribblemaniac , after having time to look through the code, I understood most.
The only thing that puzzles me, is that newLayer, which is of type Layer, doesn't need to be cast as LayerBitmap etc., but that's a minor thing.
Good job!

chchwy:
Good job @scribblemaniac!

Just a suggestion. If the PR is small, having only one or two commits like this one, `Rebase and Merge` is a good choice. It makes git history clean and straight.
image

PR #1425, opened 2020-08-15
Open last project

davidlamhauge:
This PR makes it possible to open the most recent used file directly. I have missed this feature again and again, when I have to stop animating a scene, and then have to open P2d and go selecting from the last opened files.
It's a small amount of time you save, but it's a huge improvement of UX. It's a delight to have presets, but if you work on a project with many scenes/files, this feature will be highly praised, I'm sure.
David_DavidScreencast_20200815_16_42_41
The two checkboxes in the preferences can't be checked simultaneously, but of course they can both be unchecked.


J5lx (approved):
LGTM!

chchwy:
Good job. Thank you @davidlamhauge

PR #1430, opened 2020-08-17
1427 Pclx import crash

### LINKED ISSUE #1427: Import layers from PCLX keeping reference to the layer object instead of copying on successive import
### Issue Summary
When you use the `Import Layers from PCLX` feature if you want to import a certain layer more than once (for example when importing a walk cycle for looping) your first reflex is that when you open the import dialog and load the reference project file, you can select a layer and press "import layers" as many times as you want. However doing this will load the same object as a reference into new layers and when you move one of them, the others move as well.

While this could be interesting for a future development where we can finally have multiple instances of a file inside another file, in the current workflow it's not ideal as you want to manipulate the layer objects separately even if they are referencing the same resources.

If you import the layer once, close the dialog and invoke the import dialog again this problem will not occur and the layer will be imported as a new layer object..

### Expected Results

Importing the same layer multiple times in a single import action is valid and the layer objects that are created are independent form each other.

### Video or Image Reference

https://youtu.be/0e3CNoYIL3I

### Steps to reproduce

1. Create a file with any given amount of frames and save it as PCLX (PCL doesn't work at the moment)
2. Create a new file
3. Go to File > Import > Import Layers from PCLX...
4. On the dialog load the PCLX you created in step 1
5. The layer section will be populated. Select the layer that has the keyframes
6. Press the import button once but **don't close the dialog**! Notice that the layer has been imported into the Pencil2D project.
7. Now press import layers two more times (this is only to see the effect clearly, you can only import it once more if you want)
8. Close the dialog
9. Now select a keyframe in any of the imported layers, now you'll see that the other layers are selected on the same keyframe. If you select multiple keyframes this will be visible on all imported layers. And if you move them all the layer keyframes will move accordingly.
10. After some time, if you mess up and move the keyframes erroneously, Pencil2D might crash so be careful.

### System Information
+ Pencil2D Version:
Version: 0.6.5

date: 2020-08-01_01:32:15

CPU Architecture: x86_64
Qt Version: 5.12.9

davidlamhauge:
This PR fixes the crashes that happened, when same layer was imported multiple times.
If you import a layer called 'Sam', and you have a layer of that name already, the imported layers will now be named 'Sam 2', 'Sam 3' and so forth.
If you open another file, and also import a layer called 'Sam' from this file, the layer is called 'Sam 4'.
You can now import from both *.pcl and *.pclx files.
If you open a new file after importing layers, there is no crash.

David_DavidScreencast_20200817_22_26_26
Fixes #1427 and #1446


app/src/importlayersdialog.cpp
```diff
@@ -45,19 +45,24 @@ void ImportLayersDialog::getFileName()
     FileDialog fd(this);
     mFileName = QFileDialog::getOpenFileName(this, tr("Choose file"),
                                              fd.getLastOpenPath(FileType::ANIMATION),
-                                             tr("Pencil Animation file (*.pclx)"));
+                                             tr("Pencil Animation file (*.pclx *.pcl)"));
```
chchwy, line 48: Is it possible to use `PFF_PROJECT_EXT_FILTER` instead?

davidlamhauge: The function I need from FileDialog is openFileFilters(), and it is private, so I don't know how to retrieve it. Is there another way?

CandyFace: You can get PFF_PROJECT_EXT_FILTER from fileformat.h

davidlamhauge: Ok. Now I've got it. Include fileformat.h, and use PFF_PROJECT_EXT_FILTER.
Done

chchwy:
I am reviewing this. Will send some feedback later on.

chchwy:
Just realized that problem of importing a layer multiple times. It actually adds the same object (the layer pointer) several times into the Object. So even they look like separated layers, there is only one layer object.

We need to implement a proper layer clone function to deal with this issue, which could be hard due to many corner cases.

chchwy:
Layer duplication needs to do a lot of things. Looks like it's not just adding a name suffix like "Sam 2" and "Sam 3"

When you going to duplicate a layer, you need to
1. Duplicate all the keyframe objects. Because keyframes shouldn't be shared between two layers.
2. After duplicating keyframe objects, you have to duplicate the associated files. one `.png` /`.vec` file shouldn't be shared between two keyframe objects. (so they need a new file path, too)

davidlamhauge:
You're right it needs re-implementing. What I did was more fixing the bug (hoping for a bug-fix-release soon), than fixing the code.
After importing the layers, I call getLayers() again. Thus the object is set to nullPtr and reloaded, and it will not be a duplicate.
Duplicating the layer as you suggests is probably the best solution.
Another solution could be to have two functions: One that reads the xml, and returns a list of layer names, and another that loads the selected layers from the project file. In this case it wouldn't be duplicates either.

chchwy:
Oh, I missed that the object is reloaded after importing. Then things is easier. I will send a PR maybe tomorrow to address some issues.

chchwy:
Just a note for myself: importing a vector layer needs to import the palette as well.

chchwy:
@davidlamhauge I created a PR at https://github.com/davidlamhauge/pencil/pull/15

They are generally memory management things for PCLX import feature.

1. Using std::unique_ptr to manage `mImportObjet`, this object was never deleted.
2. Implemented the ownership transfer of a layer between two objects
3. Import layer by LayerId rather than layer name (prevent same name layers)
4. Load all keyframes into memory when importing the layer.
5. Misc improvements and memory leak fixes.

Please see each commits for the details

PR #1436, opened 2020-08-20
Project recovery after Pencil2D crashes

chchwy:
Annotation 2020-08-20 153254

Try to restore the last unsaved project, usually because of Pencil2D crashes.
A recovery popup will show if the app finds any project files left in the temp folder. 

There are two recovery modes:
1. Both `main.xml` and images (.png and .vec) are fine. In this case, just load the project up from the temp folder.
2. `main.xml` is broken. Then try to rebuild the main.xml based on the images left in `data` folder, and then call mode 1. In this mode, the images are no way to be on the correct position anymore.

Things need to do next but not included in this PR:
1. Preset pop-up. Both Recovery & Preset popup are showing when the app starts.
2. Inconsistency between images and main.xml. Maybe there are images in the data folder but images are not added to main.xml yet due to uncompleted save procedure. Not sure if we need this. 
3. Multiple projects recovery. Currently it only recocvers one project at a time.

How to test:
1. Create/Load a project.
2. Force close the app
3. Maybe go to the temp folder and mess up `main.xml` 
4. Restart Pencil2D. See if the restore pop-up shows

### DIFF
--- a/app/src/mainwindow2.cpp
+++ b/app/src/mainwindow2.cpp
@@ -1568,12 +1568,15 @@ void MainWindow2::tryRecoverUnsavedProject()
 
 void MainWindow2::startProjectRecovery(int result)
 {
-    if (result != QMessageBox::Open)
-    {
-        return;
-    }
     QMessageBox* msgBox = dynamic_cast<QMessageBox*>(QObject::sender());
     const QString recoverPath = msgBox->property("RecoverPath").toString();
+
+    if (result != QMessageBox::Open)
+    {
+        // The user presses discard
+        QDir(recoverPath).removeRecursively();
+        return;
+    }
 
     FileManager fm;
     Object* o = fm.recoverUnsavedProject(recoverPath);
--- a/app/src/mainwindow2.cpp
+++ b/app/src/mainwindow2.cpp
@@ -1580,11 +1580,15 @@ void MainWindow2::startProjectRecovery(int result)
     if (!fm.error().ok())
     {
         Q_ASSERT(o == nullptr);
-        QMessageBox::information(this, tr("Recovery Failed."), tr("Sorry! Pencil2D is unable to restore your project"));
+        const QString title = tr("Recovery Failed.");
+        const QString text = tr("Sorry! Pencil2D is unable to restore your project");
+        QMessageBox::information(this, title, QString("<h4>%1</h4>%2").arg(title).arg(text));
     }
 
     mEditor->setObject(o);
     updateSaveState();
 
-    QMessageBox::information(this, tr("Recovery Succeeded!"), tr("Please save your work immediately to prevent loss of data"));
+    const QString title = tr("Recovery Succeeded!");
+    const QString text = tr("Please save your work immediately to prevent loss of data");
+    QMessageBox::information(this, title, QString("<h4>%1</h4>%2").arg(title).arg(text));
 }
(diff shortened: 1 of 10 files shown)


Jose-Moreno:
@chchwy Great news! Thank you for working on this. A recovery feature is certainly needed and one thing we discussed on discord a while ago among the many permutations of possible recovery methods was the following:

In addition to your proposals outlined here which covers the cases related to files actually existing untouched by the OS or Pencil2D automatic removal, we discussed the necessity to add a direct file copy / backup functionality which would occur prior to opening the actual file making a carbon copy of the file in the state previous to starting the work session.

Ideally this would make a copy, number it and allow Pencil2D users to have up to X number of copies handy to recover from unforeseen bugs and to cover the cases where all the files are wiped without trace or the PCLX files are "zeroed" out becoming a garbage file. 

These issues have been found to occur by a combination of the OS privacy & security handling, Pencil2D's unseen memory leaks and an interruption of the saving procedure.

This was my proposal:
https://discordapp.com/channels/342369662710972417/342670204351938562/662025930683842579

This was @scribblemaniac 's proposal at the time (along some thoughts related to the actual implementation):
https://discordapp.com/channels/342369662710972417/342670204351938562/662048769373896716

In summary, and after some afterthought, while we've discussed auto save ideas for months we managed to settle that having backup files on the same folder was agreeable.

If the user dislikes the idea of having extra packed or unpacked (PCL) backup files on the project folder (a la Blender), at most we could move it to APPDATA and keep the list of **backups** there so we can load them without issues. 

The reason I mention using the APPDATA folder as opposed to the TEMP which has the unpacked version of the working file during a work session is because this folder tends to be erased either by the OS or by the users and due to the privacy keepers in each Os getting more annoying, using that folder for recovery seems an inconvenient risk.

Besides even though we now have a direct access to the TEMP folder, since there was also a conversation about moving settings to APPDATA in preparation for MyPaint brush configuration settings, there could also be a direct access to APPDATA via a similar button in case people want to explore the potential backups manually along other settings.

chchwy:
Thanks @Jose-Moreno. I didn't know there is a proposal. It would be great to be written down somewhere (forum etc.) rather than buried in the discord chat history.

I read both your proposal. Many of your ideas have been in my mind for a long time. However, this PR will focus on the recovery part. I will send more PRs sooner or later for the file-save things.

CandyFace:
This is a great step forward, I look forward seeing this merged.
I've made two observations so far:

1. When creating a new project, the main file is never created and images are of course not exported either so the recovery popup never shows. This is expected since we don't have any data yet but maybe we should consider saving changes gradually to the temp folder instead of only on project load, as there is otherwise no gain for those people working on something but never saves.
2. When loading a project, making changes and killing the application, the dialog shows as expected. The layers however has been renamed and the camera layer (in my case) moved to the top.

recovery

chchwy:
Mmm...the layers are renamed only when the `main.xml` is damaged. I will check again.

chchwy:
@CandyFace I fixed the layer renaming issue

chchwy:
Just found that Windows will clear the temp folder even if force closing the app by the task manager. So you will need an actual crash to test this feature on Windows.

scribblemaniac:
I'd have preferred if this was reviewed before merging.

chchwy:
@scribblemaniac you can still review any code anytime if you like, I would appreciate it. I am trying to utilize my limited time and making progress.

CandyFace:
I agree with @scribblemaniac, let's try to keep the PR process and make sure things are reviewed and accepted by at least one before merging. You can still utilize your limited time but making progress should be a team effort. That's just my two cent.

chchwy:
Right. I will follow next time.

PR #1441, opened 2020-08-25
Fix #1437: "Reset Windows"

### LINKED ISSUE #1437: Implement [Reset Windows to Default] action
### Issue Summary

After various reports of people missing their panels and having trouble when they can't get them back after troubleshooting, I think it's time to consider having an action command that resets the window layout to that of the **first time experience** for pencil2d users. While we have one that resets back to the **previous state** having an option to simply start from scratch would be ideal.

chchwy:
The `Reset Windows` in Windows menu should reset all sub-panels to their default positions

### DIFF
--- a/app/src/mainwindow2.cpp
+++ b/app/src/mainwindow2.cpp
@@ -1049,11 +1049,25 @@ void MainWindow2::preferences()
 
 void MainWindow2::resetAndDockAllSubWidgets()
 {
+    QSettings settings(PENCIL2D, PENCIL2D);
+    settings.remove(SETTING_WINDOW_GEOMETRY);
+    settings.remove(SETTING_WINDOW_STATE);
+
     for (BaseDockWidget* dock : mDockWidgets)
     {
         dock->setFloating(false);
+        dock->raise();
         dock->show();
     }
+
+    addDockWidget(Qt::RightDockWidgetArea, mColorBox);
+    addDockWidget(Qt::RightDockWidgetArea, mColorInspector);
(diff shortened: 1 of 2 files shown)


CandyFace (approved):
LGTM!

app/src/mainwindow2.cpp
```diff
@@ -1049,11 +1049,25 @@ void MainWindow2::preferences()
...
     }
+
+    addDockWidget(Qt::RightDockWidgetArea, mColorBox);
+    addDockWidget(Qt::RightDockWidgetArea, mColorInspector);
+    addDockWidget(Qt::RightDockWidgetArea, mColorPalette);
+    addDockWidget(Qt::LeftDockWidgetArea, mToolBox);
+    addDockWidget(Qt::LeftDockWidgetArea, mToolOptions);
+    addDockWidget(Qt::LeftDockWidgetArea, mDisplayOptionWidget);
+    addDockWidget(Qt::LeftDockWidgetArea, mOnionSkinWidget);
+    addDockWidget(Qt::BottomDockWidgetArea, mTimeLine);
```
scribblemaniac, line 1070: It would be better if these were refactored into their own private function so the code could be shared with the same code in `MainWindow2::createDockWidgets`.

Jose-Moreno:
@chchwy I would have preferred to keep the older behavior and make this a new option, but i'll just make a new request so we can improve the scope of the previous behavior (i.e asking for a way to save up to x number of workspace layouts) Thanks again for the work!

Issue #1435, opened 2020-08-20
Canvas missing update when undoing strokes created with polyline tool.


Jose-Moreno:
### Issue Summary

When drawing any given line with the polyline tool, undoing the stroke will result in the buffer not being cleaned / updated from the stroke until the polyline tool is used to draw again. 

The after effects of the buffer not being updated (or the stroke "event" not being added?) hold remarkable resemblance to the problem from #1196 and fixing this could potentially help fixing that one which seems to be more elusive as you have to jerk your stylus in a certain way while the stroke is being computed, in order to reproduce.

### Expected Results

Creating any kind of line with any drawing tool should never become fixated to the canvas as this can and will cause confusion in the end user.

### Video or Image Reference

https://youtu.be/onnZfHpHNWw

### Steps to reproduce

1. Create a new file
2. Pick the polyline tool and draw a single "complete" stroke (that is, click, move the mouse and double click to finalize the stroke drawing mode)
3. Undo
4. Notice that the stroke is still present
5. Using the polyline tool draw a new stroke away from the previous one click on the canvas once without moving the mouse. Nothing happens. Now move the mouse slightly and you'll notice that the previous stroke disappears as soon as move the pointer.

### System Information
+ Pencil2D Version:

Version: 0.6.5

commit: 50a0c7c318f38988af85d34e955a5a78d3953474
date: 2020-08-01_01:32:15

Operating System: Windows 10 (10.0)
CPU Architecture: x86_64
Qt Version: 5.12.9

Issue #1426, opened 2020-08-16
CanvasPainter incorrectly showing elements from previously selected frames & layers


Jose-Moreno:
### Issue Summary

Using the selection tool in specific ways will break the canvas painter into showing selected elements on incorrect layers, frames and even new files. This behavior is also affecting the onion skins once it becomes broken. The issue also happens when clicking with the move tool after using the selection tool.

Basically these are the three main problems that stem form this issue:
1. Problem between keyframes (noticeable with onion skins; can't properly align frames due to missing skins when clicking)
2. Problem between layers (noticeable with/without layer opacity)
3. Problem between files (noticeable when creating new files)

Note: Input hardware is not a factor. This problem occurs regardless of mouse or tablet being used.

### Expected Results

Transforming selections with or without onion skins on different layers should not render previously selected elements to avoid confusion for the user.

### Video or Image Reference

https://youtu.be/-o0M9mDTegU

### Steps to reproduce

Between Keyframes
1. Create a New file (Important as it's possible any minute selection on a previous file has already affected the behavior)
2. Turn on onion skins to see the problem more clearly.
3. Draw something for reference on keyframe 1
4. Create a second keyframe and draw on it as well
5. Choose a keyframe (it can be the first) and place the playback header on the frame
6. Create a selection anywhere you want, doesn't have to be on top of the content.
7. [Optional] You can test leaving the selection enabled or hiding it, it doesn't seem to matter, but could help debugging.
8. Move to the other keyframe and create a selection or click with the move tool. You'll notice the previous keyframe onion skin disappears.
9. Go back to the first keyframe and create a selection or click with the move tool. You'll notice the second keyframe does NOT disappear.

Between Layers
1. Create a New file (Important as it's possible any minute selection on a previous file has already affected the behavior)
2. Turn on layer visibility: relative to see the problem more clearly.
3. Draw something for reference on the first bitmap layer available on keyframe 1
4. Create a second bitmap layer and draw on it as well.
5. Choose a layer keyframe (it can be the first) and place the playback header on the frame
6. Create a selection anywhere you want on the chosen layer, doesn't have to be on top of the content.
7. Move to the other layer and create a selection or click with the move tool. You'll notice the other layer content disappears.
8. Go back to the first keyframe and create a selection or click with the move tool. You'll notice the second layer does NOT disappear.

Between Files
1. After following either of the reproduction steps for keyframes or layers go to File > New 
[...]

PR #1443, opened 2020-08-29
More bugfixes

scribblemaniac:
I found that missing commit I forgot to push before the release :facepalm:. This will fix some of the visual rendering bugs, including probably almost everything in #1426 and #1435 (both untested). The second commit should speed up some operations by stopping the cache from being cleared unnecessarily. I suspect there is still some room for improvement in this area. I will also probably be adding some more bugfixes to this later today. More details to follow.


scribblemaniac:
Alright I'm going to leave it at this so this PR doesn't get too big. The additional fixes were the resulting of the debugging session with @Jose-Moreno for a variety of issues he was having. See individual commits for details.

app/src/mainwindow2.cpp
```diff
@@ -1380,10 +1380,19 @@ void MainWindow2::makeConnections(Editor* editor, ScribbleArea* scribbleArea)
...
     connect(editor->tools(), &ToolManager::toolChanged, scribbleArea, &ScribbleArea::setCurrentTool);
     connect(editor->tools(), &ToolManager::toolPropertyChanged, scribbleArea, &ScribbleArea::updateToolCursor);
-    connect(editor->layers(), &LayerManager::currentLayerChanged, scribbleArea, &ScribbleArea::updateAllFrames);
+    connect(editor->layers(), &LayerManager::currentLayerChanged, [editor, scribbleArea]
+    {
+        // Changing the current layer will only change the frame (as viewed by the user) under the following circumstances
+        if(editor->preference()->isOn(SETTING::PREV_ONION) ||
+           editor->preference()->isOn(SETTING::NEXT_ONION) ||
+           scribbleArea->getLayerVisibility() != LayerVisibility::ALL)
+        {
```
CandyFace, line 1423: Is this a necessary check? isn't it better to just update no matter what reason the layer changed?

CandyFace (approved):
Aside from one comment, I think it looks good.

app/src/mainwindow2.cpp
scribblemaniac: When we're talking about necessary, it's only *necessary* to update the frames when the current layer changes if onion skinning or relative layer visibility is enabled. `updateAllFrames` invalidates the entire cache, so it should be treated as a very costly operation that is only done when it is absolutely needed. A couple checks here could prevent hundreds of frames from being re-rendered.

It might be a bit better to always update all frames to avoid visual artifacts, particularly when some new feature is added which also needs to be updated when the current layer changes. However if we follow that logic to it's natural conclusion, it would be best to get rid of the cache altogether to prevent that issue under any circumstance, which I don't agree with.

CandyFace: But clearing the entire cache may also be overkill, even if you enable onion skin, you'll likely never have more than 5 frames skinned. Maybe it would be better to clear only the amount of frames that has been drawn. Sure I can see that it's an expensive operation if you have hundreds of frames but I don't see why you would invalidate all those frames either, if you know the onion skin only has been drawn on 3-5 frames.

There's no natural conclusion, of course the cache is important and should stay but checking various cases, might lead to other issues in the future, especially if we are going to implement features that might require similar handling.

I'm just saying that depending on the situation, it might be okay to not have a check, if the opposite is that we get weird artifacts in some edge case we don't know.

Alternatively I think we should have this logic in a function, so one is less likely to miss it, since it's quite performance relevant.

scribblemaniac: I don't really understand what you want here. You think clearing the cache is overkill, but then you want to remove the conditional and always clear the cache?

I should clarify here what exactly I mean when I say cache in the context of this particular discussion. What we are referring to here is the QPixmapCache in ScribbleArea. This holds *rendered* frames, which includes the onion skin. The onion skin is only for the current layer, so when you switch layers it changes. If for example you:
1. View frame 7 with the first layer active and onion skinning enabled
2. Go to frame 100
3. Switch to the second layer

Now the cached frame 7 has the onion skin for layer 1, not layer 2, and thus needs to be redrawn. Same with every other frame in the cache.

I will move it to its own function if you really want. I can't think of any other situation where you would use it, but maybe it would make it a bit easier to find.

scribblemaniac: > I assume that you mean it prevents hundreds of frames from being re-rendered, because the entire cache is being cleared, that's why I mention that instead of clearing the whole cache, why not clear only the affected frames, then it won't be as costly.

I completely agree with that sentiment. As I tried to explain in the previous message, all of the frames are affected when changing the active layer with onion skinning enabled (and the same logic applies to relative layer visibility), and none of the frames are affected otherwise, hence the check.

CandyFace: Ah... aaah right, yes okay I get what you're saying now :P Alright then what I'm suggesting doesn't make much sense.

CandyFace (approved):

CandyFace:
Alright now I've been testing the build after these fixes and both  #1426 and #1435 persists in some way... 

#1435 
actually I'd say polyline rendering is completely broken now. I can only see the stroke being rendered now after i've applied the polyline stroke, it does not render while drawing. Undo is still broken too.

#1426 
This particular issue has been fixed but a new bug has appeared that's related. 
1. Draw on frame 1
2. Draw on a frame 2
3. Move scrubber to frame 1

Result: 
The second frame is not shown immediately, you have to interact with the canvas or in some other way cause an update event.

core_lib/src/interface/scribblearea.cpp
```diff
@@ -632,6 +633,8 @@ void ScribbleArea::mousePressEvent(QMouseEvent* e)
 
 void ScribbleArea::mouseMoveEvent(QMouseEvent* e)
 {
+    // Workaround for tablet issue (#677 part 2)
+    if (mStrokeManager->isTabletInUse() || !isMouseInUse()) { e->ignore(); return; }
```
CandyFace, line 637: Actually we can't do this here, otherwise the canvas cursor won't be updated, leaving it stuck until you draw on the canvas again.

app/src/mainwindow2.cpp
```diff
@@ -1414,10 +1414,17 @@ void MainWindow2::makeConnections(Editor* editor, ScribbleArea* scribbleArea)
...
+        // Changing the current layer will only change the frame (as viewed by the user) under the following circumstances
+        if(scribbleArea->isAffectedByActiveLayer())
+        {
+            scribbleArea->updateAllFrames();
+        }
+    });
     connect(editor->layers(), &LayerManager::layerDeleted, scribbleArea, &ScribbleArea::updateAllFrames);
 
-    connect(editor, &Editor::currentFrameChanged, scribbleArea, &ScribbleArea::updateFrame);
+    connect(editor, &Editor::currentFrameChanged, [scribbleArea] { scribbleArea->update(); });
```
CandyFace, line 1427: I'd suggest we do something like above to fix cache for onion skinning

```
    connect(editor, &Editor::currentFrameChanged, [scribbleArea](int frame) {

        if (scribbleArea->isAffectedByActiveLayer()) {
            scribbleArea->updateFrame(frame);
        } else {
            scribbleArea->update();
        }
    });
```

core_lib/src/interface/scribblearea.cpp
```diff
@@ -1655,3 +1658,17 @@ void ScribbleArea::floodFillError(int errorType)
...
     mEditor->deselectAll();
 }
+
+/** Check if the content of the canvas depends on the active layer.
+  *
+  * Currently layers are only affected by Onion skins are displayed only for the active layer, and the opacity of all layers
+  * is affected when relative layer visiblity is active.
+  *
+  * @return True if the active layer could potentially influence the content of the canvas. False otherwise.
+  */
```
CandyFace, line 1668: Can we mode this into the header, so we keep documentation there.. there's no consistency here but I' rather we have it there than in the source file.

scribblemaniac:
I will look into the issues raised here after merging soon.

PR #1447, opened 2020-08-31
[Refactor] Break down FileManager::save() and introduce FileManager::flush()

chchwy:
This is a PR of code refactoring. It's a preparation step for further file saving improvements.

There are 2 parts in this PR:

1. Breaking down `FileManager::save()` into 3 smaller functions 
  - `flushKeyFrameFiles()`: write all bitmap/vector keyframes to the temp working folder.
  - `writeMainXml()`: write the main xml to disk
  - `writePalette`: write the palette xml to disk

This change makes the file save process more clear & descriptive, also make the second part possible.

2. Introducing `FileManager::flush()`

This function flushes all changes to the temp working folder without touching `.pclx`. Calling `flush()` at the right time will improve the recoverability of projects. For example, recovering a newly created project which has nothing in the temp working folder. More PRs will come soon.

### DIFF
--- a/core_lib/src/structure/filemanager.cpp
+++ b/core_lib/src/structure/filemanager.cpp
@@ -320,8 +320,9 @@ Status FileManager::save(const Object* object, QString sFileName)
                           tr("An internal error occurred. Your file may not be saved successfully."));
         }
         dd << "Zip file saved successfully";
+        Q_ASSERT(stMiniz.ok());
 
-        if (stMiniz.ok() && saveOk)
+        if (saveOk)
             deleteBackupFile(sBackupFile);
     }
 
--- a/core_lib/src/structure/filemanager.cpp
+++ b/core_lib/src/structure/filemanager.cpp
@@ -337,6 +337,28 @@ Status FileManager::save(const Object* object, QString sFileName)
     return Status::OK;
 }
 
+Status FileManager::flush(const Object* object)
+{
+    DebugDetails dd;
+
+    QStringList filesWritten;
+
+    const QString dataFolder = object->dataDir();
+    const QString mainXml = object->mainXMLFile();
+
+    Status stKeyFrames = flushKeyFrameFiles(object, dataFolder, filesWritten);
+    dd.collect(stKeyFrames.details());
+
+    Status stMainXml = writeMainXml(object, mainXml, filesWritten);
+    dd.collect(stMainXml.details());
+
+    Status stPalette = writePalette(object, dataFolder, filesWritten);
+    dd.collect(stPalette.details());
+
+    const bool saveOk = stKeyFrames.ok() && stMainXml.ok() && stPalette.ok();
+
+}
+
 ObjectData* FileManager::loadProjectData(const QDomElement& docElem)
 {
     ObjectData* data = new ObjectData;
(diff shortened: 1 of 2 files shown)


chchwy:
Please hit the `Rebase and merge` for this PR if anyone reviews and then wants to merge.

core_lib/src/structure/filemanager.cpp
```diff
@@ -374,23 +311,23 @@ Status FileManager::save(const Object* object, QString sFileName)
...
-            dd.collect(s.details());
+            dd.collect(stMiniz.details());
             return Status(Status::ERROR_MINIZ_FAIL, dd,
                           tr("Miniz Error"),
                           tr("An internal error occurred. Your file may not be saved successfully."));
         }
         dd << "Zip file saved successfully";
 
-        if (s.ok() && saveLayersOK)
+        if (stMiniz.ok() && saveOk)
```
CandyFace, line 324: I know we've had this all along, but is the checking the status again needed, we've already determined that it's ok if it comes down here because of the previous check.

core_lib/src/structure/filemanager.cpp
```diff
@@ -400,6 +337,28 @@ Status FileManager::save(const Object* object, QString sFileName)
     return Status::OK;
 }
 
+Status FileManager::flush(const Object* object)
```
CandyFace, line 340: Any particular reason we use 'flush'? couldn't it simply be write or writeToDisk?

core_lib/src/structure/filemanager.cpp
```diff
@@ -400,6 +337,28 @@ Status FileManager::save(const Object* object, QString sFileName)
...
+{
+    DebugDetails dd;
+
+    QStringList filesWritten;
+
+    const QString dataFolder = object->dataDir();
+    const QString mainXml = object->mainXMLFile();
+
+    Status stKeyFrames = flushKeyFrameFiles(object, dataFolder, filesWritten);
```
CandyFace, line 349: Same with this, couldn't we use writeKeyFrameFiles?

CandyFace (approved):
Aside from a few comments, the refactoring looks good to me 👍

core_lib/src/structure/filemanager.cpp
chchwy: No particular reason. A lot of file system API like to use the word `flush` for flushing content to the disk. For example, [this](https://doc.qt.io/qt-5/qfiledevice.html#flush), [this](https://doc.qt.io/qt-5/qtextstream.html#flush) and [this](https://en.cppreference.com/w/cpp/io/manip/flush).

core_lib/src/structure/filemanager.cpp
chchwy: You are right. I will change it to an `assert()`

core_lib/src/structure/filemanager.cpp
CandyFace: That's true but we don't use that particular word ourselves anywhere, we use 'write' instead.

I know flush is used a lot in c++ but not so much in other programming languages anymore. Whether we use flush or write, is not my issue, I just wish for consistency.

chchwy: I am trying to choose a word which means writing to the working folder but not the final `.pclx`. The meaning of `write()` or `writeToDisk()` is too similar to `save()`

CandyFace: Can't we call it writeToWorkingFolder then?

core_lib/src/structure/filemanager.cpp
chchwy: Probably because it's dumping a lot of files rather than one...a vivid description just like flushing a toilet with a lot of water :p

chchwy:
@CandyFace i have made changes accordingly to your comment

chchwy:
The word 'flush' was flushed out.

CandyFace (approved):

PR #1450, opened 2020-09-01
Manage ActiveFramePool using image size rather than number of frames

chchwy:
`ActiveFramePool` can now manage the pool with a memory budget (for example 1GB) and calculate the actual total used memory by adding up the memory usage of each frame. This is a more accurate way to prevent the app crashing from out of memory.

Previously `ActiveFramePool` keeps a certain number of keyframes, for instance, 400 frames in the memory, which is not good due to the variation of the image size as well as memory usage.

image

The default pool size is 1GB.

### DIFF
--- a/app/ui/generalpage.ui
+++ b/app/ui/generalpage.ui
@@ -38,9 +38,9 @@
       <property name="geometry">
        <rect>
         <x>0</x>
-        <y>-335</y>
-        <width>303</width>
-        <height>866</height>
+        <y>-302</y>
+        <width>292</width>
+        <height>845</height>
        </rect>
       </property>
       <layout class="QVBoxLayout" name="lay">
@@ -428,7 +428,7 @@
           <item>
            <widget class="QLabel" name="cacheLabel">
             <property name="text">
-             <string>Cached Frame Number:</string>
+             <string>Memory Cache Budget</string>
             </property>
             <property name="alignment">
              <set>Qt::AlignLeading|Qt::AlignLeft|Qt::AlignVCenter</set>
@@ -458,14 +458,17 @@
               <height>16777215</height>
              </size>
             </property>
+            <property name="suffix">
+             <string>MB</string>
+            </property>
             <property name="minimum">
-             <number>11</number>
+             <number>100</number>
             </property>
             <property name="maximum">
-             <number>1500</number>
+             <number>16000</number>
             </property>
             <property name="value">
-             <number>200</number>
+             <number>1024</number>
             </property>
            </widget>
           </item>
--- a/core_lib/src/activeframepool.cpp
+++ b/core_lib/src/activeframepool.cpp
@@ -21,12 +21,9 @@ GNU General Public License for more details.
 #include <QDebug>
 
 
-ActiveFramePool::ActiveFramePool(unsigned long n)
+ActiveFramePool::ActiveFramePool()
 {
-    size_t maxSize = n;
-
-    Q_ASSERT(maxSize > 10);
-    mMaxSize = maxSize;
+    Q_ASSERT(mMemoryBudgetInBytes >= (1024 * 1024 * 100)); // at least 100MB
 }
 
 ActiveFramePool::~ActiveFramePool() {}
@@ -38,15 +35,25 @@ void ActiveFramePool::put(KeyFrame* key)
 
     Q_ASSERT(key->pos() > 0);
 
+    key->loadFile();
+
     auto it = mCacheFramesMap.find(key);
-    mCacheFramesList.push_front(key);
-    if (it != mCacheFramesMap.end())
+    const bool keyExistsInPool = (it != mCacheFramesMap.end());
+    if (keyExistsInPool)
     {
+        // move the keyframe to the front of the list, if the key already exists in frame pool
         mCacheFramesList.erase(it->second);
     }
+    mCacheFramesList.push_front(key);
     mCacheFramesMap[key] = mCacheFramesList.begin();
+
     key->addEventListener(this);
-    key->loadFile();
+
+    if (!keyExistsInPool)
+    {
+        mTotalUsedMemory += key->memoryUsage();
+        //qDebug() << "Total Memory:" << mTotalUsedMemory;
+    }
 
     discardLeastUsedFrames();
 }
@@ -66,11 +73,11 @@ void ActiveFramePool::clear()
     mCacheFramesMap.clear();
 }
 
-void ActiveFramePool::resize(int n)
+void ActiveFramePool::resize(quint64 memoryBudget)
 {
-    n = std::min(n, 1500);
-    n = std::max(n, 10);
-    mMaxSize = static_cast<unsigned>(n);
+    memoryBudget = qMin(memoryBudget, quint64(1024) * 1024 * 1024 * 16); // 16GB
+    memoryBudget = qMax(memoryBudget, quint64(1024) * 1024 * 100); // 100MB
+    mMemoryBudgetInBytes = memoryBudget;
     discardLeastUsedFrames();
 }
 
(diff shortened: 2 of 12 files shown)


CandyFace:
Yes a very nice improvement, the next step of this if we would want to be certain how much can be cached, is to get the amount of memory that is free. Qt doesn't provide this out of the box though, so we'd have to handle each platform independently. 

Nice work. 👍

core_lib/src/activeframepool.cpp
```diff
@@ -38,22 +35,27 @@ void ActiveFramePool::put(KeyFrame* key)
...
+
     key->addEventListener(this);
-    key->loadFile();
 
-    discardLeastUsedFrames();
-}
+    if (!keyExistsInPool)
+    {
+        mTotalUsedMemory += key->memoryUsage();
+        //qDebug() << "Total Memory:" << mTotalUsedMemory;
```
CandyFace, line 55: Let's cleanup some of these comments.

core_lib/src/activeframepool.cpp
```diff
@@ -110,5 +116,17 @@ void ActiveFramePool::discardLeastUsedFrames()
 void ActiveFramePool::unloadFrame(KeyFrame* key)
 {
     //qDebug() << "Unload frame:" << key->pos();
```
CandyFace, line 118: This one too

core_lib/src/activeframepool.cpp
```diff
@@ -110,5 +116,17 @@ void ActiveFramePool::discardLeastUsedFrames()
...
+    mTotalUsedMemory -= key->memoryUsage();
     key->unloadFile();
+    //qDebug() << "Total Memory:" << mTotalUsedMemory;
```
CandyFace, line 121: This one too

core_lib/src/activeframepool.cpp
```diff
@@ -110,5 +116,17 @@ void ActiveFramePool::discardLeastUsedFrames()
...
+}
+
+void ActiveFramePool::recalcuateTotalUsedMemory()
+{
+    mTotalUsedMemory = 0;
+    for (KeyFrame* key : mCacheFramesList)
+    {
+        mTotalUsedMemory += key->memoryUsage();
+    }
+    //qDebug() << "Total Memory:" << mTotalUsedMemory;
```
CandyFace, line 131: This one too

core_lib/src/util/util.h
```diff
@@ -51,4 +51,6 @@ Container filter(const Container& container, Pred predicate) {
 QString ffprobeLocation();
 QString ffmpegLocation();
 
+quint64 imageSize(const QImage&);
```
CandyFace, line 54: This method should be const, it's not something we apply often but it does make it easier to understand that the content of this shouldn't be modified.

CandyFace (changes requested):
Minor stuff but I do think it should be done before merging.

core_lib/src/activeframepool.cpp
chchwy: I thought it will be easy for someone who wants to debug this in the future. But anyway I can remove them.

core_lib/src/util/util.h
chchwy: This is a simple function, not a class member. I have put a const for the QImage

CandyFace: Fair enough, I like to do it whenever possible for readability sake but it's not a problem to me if we don't do it here.

core_lib/src/activeframepool.cpp
CandyFace: I mean it's simple debug statements, I don't think we should have them lying around, because it's convenient. At least I think they should be removed before merging.

CandyFace (approved):
LGTM!

core_lib/src/activeframepool.cpp
```diff
@@ -66,11 +66,11 @@ void ActiveFramePool::clear()
...
 }
 
-void ActiveFramePool::resize(int n)
+void ActiveFramePool::resize(quint64 memoryBudget)
 {
-    n = std::min(n, 1500);
-    n = std::max(n, 10);
-    mMaxSize = static_cast<unsigned>(n);
+    memoryBudget = qMin(memoryBudget, quint64(1024) * 1024 * 1024 * 16); // 16GB
+    memoryBudget = qMax(memoryBudget, quint64(1024) * 1024 * 100); // 100MB
```
scribblemaniac, line 72: These two lines can be combined with `qBound`

app/ui/generalpage.ui
```diff
@@ -458,14 +458,17 @@
...
+            <property name="suffix">
+             <string>MB</string>
+            </property>
             <property name="minimum">
-             <number>11</number>
+             <number>100</number>
             </property>
             <property name="maximum">
-             <number>1500</number>
+             <number>16000</number>
```
scribblemaniac, line 468: The max in activeframepool does not match this. It is 16 * 1024, not 16000.

app/ui/generalpage.ui
```diff
@@ -458,14 +458,17 @@
               <height>16777215</height>
              </size>
             </property>
+            <property name="suffix">
+             <string>MB</string>
```
scribblemaniac, line 462: Technically this, as well as in the other places where MB is used, should be MiB.

CandyFace: While that's technically correct, Not many are familiar with this term, neither windows or mac give you mib, I believe Ubuntu does but frankly it's not something I've ever noticed.

core_lib/src/activeframepool.cpp
```diff
@@ -87,12 +87,16 @@ void ActiveFramePool::onKeyFrameDestroy(KeyFrame* key)
...
+        // Just recalculate the total usage
+        // Not safe to call key->memoryUsage() here cuz it's in the KeyFrame's destructor
+        recalcuateTotalUsedMemory();
     }
 }
 
 void ActiveFramePool::discardLeastUsedFrames()
 {
-    while (mCacheFramesMap.size() > mMaxSize)
+    while (mTotalUsedMemory > mMemoryBudgetInBytes)
```
scribblemaniac, line 99: I think there should also be a check here to stop if a minimum number of frames has been reached (probably matching the previous minimum for mMaxSize).

Consider the scenario where a single frame exceeds the memory budget. The frame would be immediately unloaded upon adding. I'll admit this scenario is the most unlikely given the 100 MiB minimum, but I've had images larger than that before, and it becomes much more likely if you consider that when scrubbing, `Object::updateActiveFrames` is called which loads all frames on all layers in a range that spans a total of 7 frames. On a project with three layers, each filled with frames, each scrub will load 21 frames, and some of those frames will be immediately unloaded if the average frame size exceeds ~4.7 MB, which is not all that outlandish for pngs.

At worst this causes serious issues for the code calling the loading which likely assumes that the image it has just called has been loaded. At best it causes unnecessary reloading of the files. Just skimming through the code, it looks as though paintImage will load the file on its own so there will be no visual issues, however since the frame is not the in the cache, it's resources will not be freed until it is added again, potentially keeping every frame in memory if you are scrubbing forward or playing under the right circumstances.

core_lib/src/activeframepool.cpp
chchwy: Will do thanks

app/ui/generalpage.ui
chchwy: will update

app/ui/generalpage.ui
chchwy: Didn't see MiB very often to be honest. In many places, MB refers to the same meaning of MiB.

core_lib/src/activeframepool.cpp
chchwy: Good point. Will improve it.

PR #1451, opened 2020-09-02
Miniz improvements

scribblemaniac:
The main changes here is that `mz_zip_reader_end` and `mz_zip_reader_end` are always called now, even if an error is encountered. When compressing always put `mz_zip_get_last_error` in the debug details if the compression fails. The remainder of the changes are cosmetic.


chchwy:
Nice find! thank you @scribblemaniac

PR #1454, opened 2020-09-06
Add Retiming feature

scribblemaniac:
I took a short break from my seemly endless efforts to reduce the bugginess of Pencil2D to introduce my own bug-inducing feature!

Screenshot from 2020-09-05 23-24-00

This is the retiming feature. It is accessible by right clicking on the fps timeline control or accessing it in Animation menu. In this dialog where you can change the fps and speed of the project independently. I.e. you can change the fps of a project without affecting its playback speed, and you can change the playback speed without affecting the fps, or you can change both simultaneously. For example, if you have an animation on 1s at 12 fps, you can retime it to 24 fps and the frames will be on 2s (and the same duration). If you then retimed it again at 2x speed and it would be 24 fps on 1s (and half the duration). Frames may be dropped when decreasing the fps or increasing the speed during retiming for obvious reasons.

I believe this feature, or variants of it, have been requested on multiple occasions. I don't really have a strong interest in it, but it was a simple enough to do, I can see its use, and I'm fine with it as long as it doesn't get in the way. I am open to feedback.

### DIFF
--- a/app/app.pro
+++ b/app/app.pro
@@ -73,7 +73,8 @@ HEADERS += \
     src/doubleprogressdialog.h \
     src/colorslider.h \
     src/checkupdatesdialog.h \
-    src/presetdialog.h    
+    src/presetdialog.h \
+    src/retimedialog.h
 
 SOURCES += \
     src/importlayersdialog.cpp \
@@ -109,7 +110,8 @@ SOURCES += \
     src/colorslider.cpp \
     src/checkupdatesdialog.cpp \
     src/presetdialog.cpp \
-    src/app_util.cpp
+    src/app_util.cpp \
+    src/retimedialog.cpp
 
 FORMS += \
     ui/importimageseqpreview.ui \
@@ -137,7 +139,8 @@ FORMS += \
     ui/filespage.ui \
     ui/toolspage.ui \
     ui/toolboxwidget.ui \
-    ui/presetdialog.ui
+    ui/presetdialog.ui \
+    ui/retimedialog.ui
 
 
 
--- a/app/src/actioncommands.cpp
+++ b/app/src/actioncommands.cpp
@@ -56,6 +56,7 @@ GNU General Public License for more details.
 #include "doubleprogressdialog.h"
 #include "checkupdatesdialog.h"
 #include "errordialog.h"
+#include "retimedialog.h"
 
 
 ActionCommands::ActionCommands(QWidget* parent) : QObject(parent)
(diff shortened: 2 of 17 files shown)


CandyFace:
I've reviewed the code and it looks fine style wise, no comments really... I would approve it but then I tested the feature and I can't make it work. When I use the retime feature, the result ends up either removing my frames or crashing the application. 

What I tried:
* Added three frames with content
* Opened the retime dialog
* changed the speed from original 1.0x to 1.5x
Result:
Lost my third frame

Then I tried:
* Add three frames.. same procedure as last time
* opened the retime dialog
* changed fps from 10 to 15
result:
Application crashes.

Am I missing something?

Regardless of the above, the crashes should be fixed before we merge.
I also think that it re-timing shouldn't be able to remove frames, eg. you shouldn't be able to decrease timing if it means that frames will be removed, unless someone can see a usecase for it?

I don't have other comments currently since I haven't seen it work yet but I think the feature could be pretty useful.

A future enhancement could also involve re-timing within a range, eg. selected frames instead of modifying globally.

J5lx:
What’s the status of this?

scribblemaniac:
I need to find and fix the crashing that @CandyFace encountered and we need to have a discussion about the desired behavior for this. The frame removal behavior was intentional and I think is the way it should behave. It's also why I stuck that bright red warning on the dialog box. If you want to speed up your animation, at some point you will have to either increase the fps or remove frames. You can already change the fps simple enough, so this feature is to handle the other approach. If you need specific use case, considering importing an image sequence that is animated on twos at double your project frame rate. You would import this in and it would play at half the desired speed. Currently the only way to work around this is to delete every other frame and move them all, or import every other frame, both of which are not feasible for large image sequences. With this feature it would be a couple of clicks. Another use case is for reducing the frame rate of a project after you have imported a video. There's going to be a keyframe on every frame of the video import layer. If you want the video to still play back at full speed, you would need to do the same thing as with the image sequence import issue, or you would need to delete the whole layer and reimport it after changing the fps. Yet another use cases is where you want an import to be played back faster than real time, perhaps as a timelapse reference. The best way to do this currently would probably be to speed up the video in another program before importing which is not as inconvenient as some of the other issues, but if Pencil2D can handle it just by not limiting a generic feature, then great let's do that imo. 

An additional note on the status of this PR. It isn't a top priority for me because I believe it may also be best to wait until after the undo/redo rewrite before introducing this feature. This would enable us to avoid the possibility of a user using this feature and accidentally wiping out a bunch of their frames permanently.

MrStevns:
Looking at this again from a usability pow, I personally think the problem with removing frames is that you don't know what frames will be affected. I also find it very easy to make an unwanted behavior, because you need to know exactly what you're doing and what it might potentially affect. Seeing the result would definitely improve the UX of this.

Is it necessary to be able to change fps and speed at the same time, maybe we could allow only one of them eg. via a combobox?

Also, I did some more testing and I've encountered another problem, though I can't reproduce it consistency yet... sometimes the frames will be placed in a wrong order... eg. 1 | 2 | 3 | 4 -> 1 | 2 | 4 | 3

Last but not least, I've made a small PR that should fix the crash I was experiencing

Issue #1455, opened 2020-09-06
Implement a Customizable toolbar for screen tablet users


Jose-Moreno:
### Issue Summary

This issue is meant to track the feature request raised in https://github.com/pencil2d/pencil/issues/487 point #2 which deals with the necessity of having a customizable toolbar / widget to add different action buttons on demand. 

This should allow the user to add any Pencil2D action that is exposed to the menus and / or to the shortcuts list to improve efficiency and flexibility in unique workflows and when using screen tablets.

This kind of feature is best used in conjunction with a full screen / canvas-only mode such as the one proposed in https://github.com/pencil2d/pencil/issues/1151

As reference please take a look at TVPaint Custom panels / toolbars. Here's a video on what they are and how to use them https://www.youtube.com/watch?v=qv1fpK9IItQ

If we reference an open source software think of Blender capability to create panels, though their process is quite more involved as it depends on an API to create the panels from scratch. 

In Pencil2D however we could provide a simpler configuration interface to add the action buttons between columns such as the **main menu editor** found in the GNOME desktop ![image](https://user-images.githubusercontent.com/4912907/92330596-aa574380-f035-11ea-8e07-3bc863463084.png)

Issue #1417, opened 2020-08-08
Incorrectly overlapped panel layout contents on interface resizing


Jose-Moreno:
### Issue Summary

On the past 2 days we've had two different user reports that evidence an issue with how the panels minimum size react unexpectedly to sudden changes in their environment. Coincidentially I managed to see this issue myself the first time I opened Pencil2D 0.6.5 barely after restoring the _"factory settings"_ of Pencil2D in the Windows registry.

### Expected Results

While it's true that most panels should have a minimum size to preserve most of their characteristics, this sometimes constraint the users to properly rearrange the interface of Pencil2D so it matches their screen resolution. We must not forget that not everyone work using Desktop computers, but usually they work with laptops and even then sometimes these come with limited non-standard resolutions.

The expected result is to allow people to resize, move, stack and close the panels without other panels overflowing their information.

It would be worth looking at this PR which implemented FlowLayout https://github.com/pencil2d/pencil/pull/1317
As well as this open PR which was meant to solve the minimum size issues on the onion skin panel (however it could be possible to apply this to other panels as well) https://github.com/pencil2d/pencil/pull/1369

### Video or Image Reference

![image](https://user-images.githubusercontent.com/4912907/89702415-bd77d600-d906-11ea-862d-769743c823d5.png)

_Courtesy of user Krijart_

https://www.youtube.com/watch?v=5c7_QLa2Yfs
_Courtesy of user aymanugod_
https://discuss.pencil2d.org/t/timeline-keeps-disappearing-whenever-i-press-alt/4743

### Steps to reproduce
It seems 100% of the current reports come from Windows 10 users (why doesn't that surprise me anymore...) However I'd like to ask other OS users to test this as well. Also the users seem to have approximate resolution of 1366 x 768 which is known as Wide XGA or WXGA.

In my case I have a 1920 x 1080 resolution, so it is more difficult to reproduce. For testers you can consider changing your resolution to the aforementioned resolution or something with similar aspect ratio for testing, but it is not required.

Since it's difficult to reproduce this with the timeline itself, we're going to use a different panel.
1. Open Pencil2D
2. Detach the tools panel
3. Attach it horizontally and reduce the vertical size until it fits and can't go beyond.
![image](https://user-images.githubusercontent.com/4912907/89702617-a508bb00-d908-11ea-89ee-07b31a1952e3.png)
4. Make sure there is at least one panel below the tools panel to see this effect.
[...]

PR #1468, opened 2020-09-16
Toolbars

chchwy:
An initial PR of adding Toolbars. The toolbars are in a basic usable level that having only essential/frequently used items.
I want to bring it up early for your attention and discussion. More features/PRs will come later on.

Currently, there are 3 toolbar sections:

1. Main toolbar: New, Open, Save, Cut, Copy, Paste, Undo, Redo
2. View toolbar: Zoom in/out, Reset view, Flip horizontal/Vertical
3. Overlay toolbar: Show/hide Grid

You can show/hide each of the toolbars under `Windows` menu.

Implementing toolbars have two goals:

1. Addressing #1047 & #1455 for touch screen users to easily access menu features (e.g., Undo/Redo) without a keyboard.
2. Addressing #1417, to efficiently utilize UI space for low res display users. It's meant to replace DisplayOptions with View/Overlay Toolbars. (not done yet, will come later on)

toolbars

BTW, toolbars can be snap to any side of the main window, not only the top area.
snap-left


chchwy:
And note: some of the menu items don't have icons. We probably need to make them.

Jose-Moreno:
@chchwy Nice work. A quick note on  #1455 that issue meant to be requesting an empty, fully customizable widget (I called it a toolbar but it could be a panel) that people can modify according to their preference and allow them to place any action available on Pencil2D menus or other panels as a quick access. In that sense this PR should not be linked to close the aforementioned issue.

From the presented view toolbar are the zoom icons meant to be clicked repeatedly to increase / decrease the zoom % factor? Or are they meant to be held down to drag over the surface? I'm asking because having these buttons would probably be duplicating the _zoom slider_ that will be implemented by #1442 If anything perhaps that zoom slider widget could be moved to your toolbar instead after both PR's are merged.

Regarding the view toolbar itself, I understand you separated this one to contain the actions that affect the view itself such as zoom, mirror, etc. I just made a request in #1469 to separate the reset transforms. These individual buttons could also be added to this toolbar in that case.

I think it's great that the overlays will also have the grid toggle and adding the layout composition guides will make sense there. For the vector related buttons _show outlines_ and _show invisible_ I'd probably keep them in the **overlays toolbar** since just by the current division we could consider the overlays to affect how the user sees or filters the content without necessarily changing the view (unlike the usual transforms).

If I can think of anything else i'll comment again.

Jose-Moreno:
> And note: some of the menu items don't have icons. We probably need to make them.

For this along the overall interface icons there was some progress discussed in #1356 and the PR is #1361 however it's true that there are plenty of missing icons for the menu actions

chchwy:
1. Currently it's not customizable. It's always possible to implement the ability of toolbar customization in the future.
2. The items on the toolbar are not decided yet. I will be happy to remove the zoom things if the status bar zoom slider code is merged.
3. All the items on the toolbar are buttons at the moment, no drag, just click.
4. I am going to add the rest of "DisplayOptions" to toolbars. 
5. Probably `show outlines` and `show invisible` could be in a separate "Vector" toolbar?

MrStevns (approved):
Changes looks good to me, I say we merge this one, it's been sitting there for far too long.

I found conflict markers in the latest commit, so I fixed that and will merge the PR now.
There are users who have been waiting for this for a long time, good job Matt 👍

PR #1470, opened 2020-09-17
Hide dock widget title bars when "Lock Windows" is selected

chchwy:
As title. Please see screenshots:

Go to `Windows` menu -> tick `Lock Windows`: Hide dock widget titlebar.
hide-toolbar

Untick `Lock Windows`: the title bar shows again
show-toolbar

would love to know how it looks like on other platforms.

### DIFF
--- a/app/src/mainwindow2.cpp
+++ b/app/src/mainwindow2.cpp
@@ -984,9 +984,16 @@ void MainWindow2::importGIF()
 void MainWindow2::lockWidgets(bool shouldLock)
 {
     QDockWidget::DockWidgetFeatures feat = shouldLock ? QDockWidget::NoDockWidgetFeatures : QDockWidget::AllDockWidgetFeatures;
+
     for (QDockWidget* d : mDockWidgets)
     {
         d->setFeatures(feat);
+
+        // https://doc.qt.io/qt-5/qdockwidget.html#setTitleBarWidget
+        // A empty QWidget looks like the tittle bar is hidden.
+        // nullptr means removing the custom title bar and restoring the default one
+        QWidget* customTitleBarWidget = shouldLock ? (new QWidget) : nullptr;
+        d->setTitleBarWidget(customTitleBarWidget);
     }
 }
 


app/src/mainwindow2.cpp
```diff
@@ -984,9 +984,16 @@ void MainWindow2::importGIF()
...
+
     for (QDockWidget* d : mDockWidgets)
     {
         d->setFeatures(feat);
+
+        // https://doc.qt.io/qt-5/qdockwidget.html#setTitleBarWidget
+        // A empty QWidget looks like the tittle bar is hidden.
+        // nullptr means removing the custom title bar and restoring the default one
+        QWidget* customTitleBarWidget = shouldLock ? (new QWidget) : nullptr;
+        d->setTitleBarWidget(customTitleBarWidget);
```
CandyFace, line 996: If we intend to add this to all our dock widgets, wouldn't it be better to apply this to the BaseDockWidget class instead?

chchwy: What do you mean exactly? you mean adding a function called `BaseDockWidget::hideTitlebar()`? 

I don't do so because it's one line code and won't be called anywhere except here.

chchwy: I thought again. Yea, probably we can do that to make code more expressive.

chchwy:
Two good things about hiding title bars:

1. Timeline has space to show 2 more tracks. Consider with the low res display (1366x768), the number of visible timeline tracks increases 3 -> 5.
2. The title bar is kinda useless when all panels are locked. You can't drag/move/undock. So it may not be very helpful from my aspect just showing names when all dock widgets are self-descriptive.

chchwy:
BTW, @CandyFace mentioned in our discord channel that he will be implementing a hover over title bar, which is hidden by default and shows only when the mouse cursor goes on it.

CandyFace:
This has been merged now, feel free to make the refactoring whenever you like. It's such a small detail that I don't think we need to stall the review for that.

PR #1471, opened 2020-09-17
Fix issues caused by #1443

scribblemaniac:
The following issues have been identified as caused by #1443, and they are all hopefully resolved with these changes:
- Onion skins not updating when cached and a previous/next frame is changed
- Crash when scrubbing to large positions (exact cutoff where it starts to crash depends on the frame pool size)
- The workaround for the mouse move events produced by some tablets (i.e. the reed issue) broke the polyline tool and the cursor. I have changed the approach used to fix this issue, and hopefully everything is working alright now. This still needs to be tested to make sure it fixes the issue that @Jose-Moreno was having.

### DIFF
--- a/core_lib/src/interface/scribblearea.cpp
+++ b/core_lib/src/interface/scribblearea.cpp
@@ -1703,13 +1703,6 @@ void ScribbleArea::floodFillError(int errorType)
     mEditor->deselectAll();
 }
 
-/** Check if the content of the canvas depends on the active layer.
-  *
-  * Currently layers are only affected by Onion skins are displayed only for the active layer, and the opacity of all layers
-  * is affected when relative layer visiblity is active.
-  *
-  * @return True if the active layer could potentially influence the content of the canvas. False otherwise.
-  */
 bool ScribbleArea::isAffectedByActiveLayer() const
 {
     return mPrefs->isOn(SETTING::PREV_ONION) ||
--- a/core_lib/src/interface/scribblearea.cpp
+++ b/core_lib/src/interface/scribblearea.cpp
@@ -99,9 +99,7 @@ bool ScribbleArea::init()
     setSizePolicy(QSizePolicy(QSizePolicy::MinimumExpanding, QSizePolicy::MinimumExpanding));
 
     QPixmapCache::setCacheLimit(100 * 1024); // unit is kb, so it's 100MB cache
-
-    int nLength = mEditor->layers()->animationLength();
-    mPixmapCacheKeys.resize(static_cast<unsigned>(std::max(nLength, 240)));
+    mPixmapCacheKeys.clear();
 
     return true;
 }
@@ -187,17 +185,17 @@ void ScribbleArea::updateCurrentFrame()
 
 void ScribbleArea::updateFrame(int frame)
 {
+    Q_ASSERT(frame >= 0);
+
     int frameNumber = mEditor->layers()->lastFrameAtFrame(frame);
     if (frameNumber < 0) { return; }
 
-    Q_ASSERT(frame >= 0);
-    if (mPixmapCacheKeys.size() <= static_cast<unsigned>(frame))
+    auto cacheKeyIter = mPixmapCacheKeys.find(static_cast<unsigned int>(frameNumber));
+    if (cacheKeyIter != mPixmapCacheKeys.end())
     {
-        mPixmapCacheKeys.resize(static_cast<unsigned>(frame + 10)); // a buffer
+        QPixmapCache::remove(cacheKeyIter.value());
+        mPixmapCacheKeys.remove(cacheKeyIter.key());
     }
-
-    QPixmapCache::remove(mPixmapCacheKeys[static_cast<unsigned>(frameNumber)]);
-    mPixmapCacheKeys[static_cast<unsigned>(frameNumber)] = QPixmapCache::Key();
 
     update();
 }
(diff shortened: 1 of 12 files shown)


chchwy:
Access an std::vector is a lot faster than access a QMap. Reszing an std::vector isn't expensive due to the size of `QPixmapCache::Key`. Any reason we cannot just resize the vector to avoid out of bound issue?

scribblemaniac:
> Accessing an std::vector is a lot faster than accessing a QMap.

I'm skeptical that there is that significant of a difference. Resizing the vector frequently would be worse in terms of time complexity, and definitely worse in terms of memory. Quite frankly I feel like using a vector in this situation is inappropriate, whether or not it is the best for speed. It's sparse data, it crashes if we forget to resize it anywhere before accessing, and we end up creating all these fake keys for when there is no key at a frame. We are already using QMap for keyframe access which is a very similar use case, and is arguably accessed a lot more.

Finally, if that is all not enough, consider that it's not *just* the vector access. When we want to retrieve a frame from the cache the old way, we look the position in the vector (fast), but then we have to look up the position of that key in the QPixmapCache to see if there is actually a cached pixmap for that key, which *at best* performs this search as well as QHash, potentially worse since the complexity of QPixmapCache::find does not appear to be documented. In fact because of the way it removes the oldest cached items when running out of space and it's not fixed length, it's probably a linked list with an O(n) search time. With QMap, we can tell if there is a cached frame in the QPixmapCache before running find at all.

I could switch it to QHash easily and that would have an amortized O(1) complexity for access if speed here is really a concern for you, the only reason I went with QMap instead is for the sake of consistency with the Layer's keyframe storage.

chchwy:
Actually `std::vector::resize()` doesn't allocate/deallocate memory most of the time. It only happens when the internal capacity is not enough. Most of the cases, calling `std::vector::resize()` only changes an internal count number, doesn't touch the memory, and not moving any elements.

Talking about `QPixmapCache::find()`, if you look at the Qt source code 
https://code.woboq.org/qt5/qtbase/src/gui/image/qpixmapcache.cpp.html#_ZN12QPixmapCache4findERKNS_3KeyEP7QPixmap

Checking a fake key is a pointer check. 
```
if (!key.d || !key.d->isValid)
    return;
```
In terms of time complexity and actual speed, `std::vector` is the best. I am fairly sure to say that it will be faster than QHash and QMap overall.

I am not saying the QMap implementation is inappropriate, just point out some of your knowledge about `std::vector` are not correct. And using a `std::vector` is a simpler way to achieve the same goal with the bonus of performance.

About why we used a vector to store QPixmapCache, that is because the String key implementation was buggy. It happens a couple of times in the old Qt versions that it got fixed and broken and fixed and then broken again. Can't remember exactly but is about the Qt 5.2~Qt5.7 period.

I made the Layer's keyframe storage with an `std::map`, and I regret a little bit. It should have done by an array structure with a binary search. It would perform better in both speed and memory.

chchwy:
Going into the PixmapCache, Qt itself is using a QHash to manage these keys
https://code.woboq.org/qt5/qtbase/src/gui/image/qpixmapcache.cpp.html#_ZNK8QPMCache6objectERKN12QPixmapCache3KeyE

CandyFace (approved):
Changes looks good to me, tested the tablet handling and while there is still produced one occational mouse move in between tablet events, it indeed gets filtered at tool level, so I'd say it works as intended.

I don't feel too strong about the cache methodology, so whatever goes is fine with me.
I also tested the opacity cache issue and it seemed gone to me.

Well done 👍

chchwy:
Looked through the PR I think it generally is good. Having some style suggestions but nothing critical. 

1. move the `event->inputType() == mCurrentInputType` comparison to a StrokeTool's member function. And make `mCurrentInputType` private if it's not intended to be modified by subclasses.
2. Using QMap to manage pixmap cache keys seems unnecessary, we need to maintain one more object's state when QPixmapCache can managing itself quite well. All we need to do is making sure std::vector has enough capacity. This is a simple and straightforward solution with performance bonus.

chchwy:
Merged. thank you @scribblemaniac

Jose-Moreno:
Commenting here mainly because I already tested this branch PR on sunday and told scribblemaniac that the change that was made in https://github.com/pencil2d/pencil/pull/1471/commits/05ef0aa9e3bfc24aa14c9da866f5ab39a2b911af it didn't fully fix what https://github.com/pencil2d/pencil/pull/1443/commits/5c8f2da429f795992643160143905e931d79f05e had improved previously.

Yes it happens _less often_, but for me that's like 1-2 less dabs. I'm not kidding, these are the images I sent him last sunday:

**Current PR Branch compiled with 5.14.1**
image

**Nightly Build from September 15th**
image

And this one i just captured

**Version compiled with Qt from 55433a7f69eb9c8fd6834c60ccf250a029a250b1**

image

The problem is that i'm not the only one using Windows 10 that has these serrated broken lines when using Windows Ink, and I've shown this before in the discord server as well. 

The only way to use the tablet drivers for real, according to our debugging session and further research is to enable Windows 7 compatibility which forces Windows to use the Wintab32.dll since Microsoft disabled this to favor Windows Ink API. 

This is honestly an issue since by default everyone ends up using Windows Ink when they first use Pencil2D and a "new" tablet.

After a lot of head scratching, I can use Windows Ink rather decently with Pencil2D, but it only works with that patch where the move events are non existent when using a stylus. Please let's try to reach a solution where there's no leftover moveEvents like in the previous PR

CandyFace:
I've been a reading up on how Krita has fixed the problem on their end and they too provide a lot of custom event handling, to make sure mouse events are eaten when a tablet event is expected.
We could take inspiration from some of their code.

https://github.com/KDE/krita/blob/44dd6e1e22c2b43a63224658ccf69aa40de8dbaa/libs/ui/input/kis_input_manager_p.cpp
https://github.com/KDE/krita/blob/44dd6e1e22c2b43a63224658ccf69aa40de8dbaa/libs/ui/KisMouseClickEater.cpp

Qt simply doesn't handle tablet events very well it seems...

PR #1475, opened 2020-09-18
Implement show titlebar only while hovering widget.

CandyFace:
Experimental feature, and not sure how it works across windows and linux.
On mac it works alright though whether you can adjust to the bouncing is debatable... we can always make it possible through a setting.

titlebar

### DIFF
--- a/core_lib/src/interface/basedockwidget.cpp
+++ b/core_lib/src/interface/basedockwidget.cpp
@@ -34,7 +34,7 @@ BaseDockWidget::BaseDockWidget(QWidget* pParent)
                       "border-width: 1px; }");
     }
 #endif
-
+    mNoTitlebar = new QWidget();
 }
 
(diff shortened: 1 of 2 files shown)


scribblemaniac (approved):
Technically everything here looks alright so far, but I'm having a hard time imagining who would want this. If you think there's a use for it, I'm fine with it being added, but as an opt-in setting.

chchwy:
One thing I noticed immediately is the sudden movement of UI positions, it could be irritating. After seeing the results, probably it's better to leave it to users whether they want to hide the title bar or not.

CandyFace:
I don't know exactly who would want to, I felt like experimenting a bit to see whether it would work or not but having used it until now, I think the effect is too extreme and causes annoyance because of the displacement, so I'll be closing the PR again. 

The other PR #1470 seems like a good compromise, so I will close this PR for now.

PR #1476, opened 2020-09-19
#1435 - Fix polyline buffer not being cleared.

### LINKED ISSUE #1435: Canvas missing update when undoing strokes created with polyline tool.

CandyFace:
Closes #1435

### DIFF
--- a/core_lib/src/tool/polylinetool.cpp
+++ b/core_lib/src/tool/polylinetool.cpp
@@ -269,5 +269,8 @@ void PolylineTool::endPolyline(QList<QPointF> points)
         if (bitmapImage == nullptr) { return; } // Can happen if the first frame is deleted while drawing
         bitmapImage->paste(mScribbleArea->mBufferImg);
     }
+
+    mScribbleArea->setAllDirty();
+    mScribbleArea->clearBitmapBuffer();
     mScribbleArea->setModified(mEditor->layers()->currentLayerIndex(), mEditor->currentFrame());
 }


chchwy:
Confirmed issue fixed. thanks @CandyFace

PR #1478, opened 2020-09-24
UI Widget layout

CandyFace:
This PR tries to further address the onion skin widget as well as some other UI changes.
Screenshot 2020-09-24 at 18 58 36
- The Onion Skin widget has been made a tad smaller
- Margins has been adjusted
- Min/max has been added next to the spinbox and the percentage sign has become a suffix of the spinboxes.

Both vertical and horizontal spacing should be cared for, even if the screen is much wider than it is tall, otherwise we'll just make the canvas area smaller, which I don't think is a good idea. I think we should keep the onion skin widget as is.

All widgets except the color inspector and colorwheel has been made scrollable, this makes it possible now to make the application quite small in height and should also fix most if not all problems with widget overflow, which we've seen in the wild.

The reason why I haven't made the colorwheel scrollable is because I think it will hurt the usability of it. The widget can become quite small already. The same can be said about the color inspector

Screenshot 2020-09-24 at 18 59 06
Min height with the setup I'm using.

Additional changes:
Timeline:
- Extra separator removed
- FPS box has been moved next to playback

Onion skin widget:
- Can only enable red/blue onion skin when the respective previous/next button has been checked. 

DisplayOptionWidget:
- Can only enable thinLines/outline when the current layer is of vector type.

### DIFF
--- a/app/src/displayoptionwidget.cpp
+++ b/app/src/displayoptionwidget.cpp
@@ -45,8 +45,6 @@ void DisplayOptionWidget::initUI()
     updateUI();
     makeConnections();
 
-    delete ui->innerWidget->layout();
-
     FlowLayout *layout = new FlowLayout;
     layout->setAlignment(Qt::AlignHCenter);
     layout->addWidget(ui->mirrorButton);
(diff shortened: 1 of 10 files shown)


chchwy:
Quickly went through the PR and checked with a 1366x768 display. It's a lot better than before. Good improvements.

PR #1480, opened 2020-10-04
Separate docs update from CI

J5lx:
This is a change to our CI setup to run Doxygen separately from the regular CI job. This is useful to reduce the build times a little and ensures the docs are kept up-to-date even when there are compiler errors, nightly build uploads fail, etc. While I was at it, I rewrote the deployment script is because the current one ([source](https://gist.github.com/vidavidorra/548ffbcdae99d752da02)) does not have a license, so technically [we never had permission](https://choosealicense.com/no-permission/) to include it in the repository.


CandyFace:
LGTM 👍

CandyFace (approved):

Issue #1429, opened 2020-08-17
Undoing strokes with specific timeline drawing behavior creates unwanted keyframes


Jose-Moreno:
### Issue Summary

In the preference settings under the "drawing" section the "Keep drawing on previous keyframe" is configured by default, so consider this as it will and has already affected new users.

The problem is that whenever you're drawing in an empty frame, even if you move back over to the "previous" keyframe, undoing any stroke will create a new keyframe **with a copy of the drawing** on the empty frame where the playhead was positioned at the time of the drawing. And not only that but if you keep pressing undo, it will undo the strokes on the copy, until the original state of the stroke previous to drawing on the empty frame, will be reached.

This is extremely frustrating even when knowing about it as sometimes you need to draw on a layer that only has a single keyframe, to trace the motion path or create a layout guide of an existing animation. So if you undo a single stroke, it will create the new keyframe for you and mess up all your drawings because undoing anymore will undo the drawing itself and leave you with a different copy.

### Expected Results

Creating or deleting keyframes if you drew when the playhead was positioned over an empty frame should only work with the modes: Create a blank keyframe or Duplicate the previous keyframe.

### Video or Image Reference

https://youtu.be/Cb8XRC-PzK8

### Steps to reproduce

Make sure when you go to Edit > Preferences > Timeline > Drawing section > That the option mode that's enable is "Keep drawing on previous keyframe" otherwise the steps won't work.

1. Create a new file
2. Draw on the first keyframe on the bitmap layer a single stroke
3. Move the playhead to any empty frame after it
4. Draw enough strokes to notices changes once you start undoing
5. Go back to the first frame (you can stay there but it's so you see it does not affect the behavior)
6. Start undoing slowly. Notice how the playhead moved to the previous FRAME, creates a duplicate KEYFRAME and each stroke starts to become undone on the new keyframe.
7. Once you reach the initial stroke that was done in the first keyframe, this will be undone on that keyframe alone, and the same stroke will remain on the new keyframe.

### System Information
+ Pencil2D Version:
Version: 0.6.5

commit: 50a0c7c318f38988af85d34e955a5a78d3953474
date: 2020-08-01_01:32:15

Operating System: Windows 10 (10.0)
CPU Architecture: x86_64
Qt Version: 5.12.9

Issue #1412, opened 2020-08-05
Crash when undoing after layer deletion


Kaikogf:
### Issue Summary

 

I imported a new **Movie Video** and deleted its layer after that, and when I tried to UNDO it to have the layer back pencil2D closed without warning

### Steps to reproduce

Remember not to click any where besides the steps mentioned here, otherwise the issue will not show
File > Import > Movie Video > (I selected a 850kb MP4 movie video with sound). After imported successfully, go to **Layers** and delete the bitmap layer (which contains the movie video), and then click **UNDO**

One Note:
When importing a **Movie Video**, it will not import its audio. Is this intentional?

### System Information
+ Pencil2D Version:
Last stable version v.0.6.5

+ Operating System:
Windows 10

+ RAM Size:
32GB

+ Graphics Tablet:
HUION HD 185

scribblemaniac:
Undoing a layer deletion is not supported currently, regardless of what its contents are. What you are probably doing is attempting to undo the image import which crashes because the layer no longer exists. I thought there was an issue for this already but I can't seem to find it at the moment. I've been aware of it for a long time and it will be fixed with #953. This has nothing to do with the movie video import.

> When importing a Movie Video, it will not import its audio. Is this intentional?

Yes. You must use the Movie Audio import separately to import the audio.

Kaikogf:
Hi @scribblemaniac 

> What you are probably doing is attempting to undo the image import which crashes because the layer no longer exists. 

No. I imported a movie video to the default layer (it played OK on the timeline) and after that I deleted that specific layer and tried to undo it, which crashed pencil2d 👍

scribblemaniac:
Think of it this way. There is a stack with all the undoable actions. When you start a new file it is empty. When you import a movie, it adds many actions to the stack. If for instance the video is a total of 4 frames in length, the stack would look like this:
1. Insert image on Frame 1 of Layer 1
2. Insert image on Frame 2 of Layer 1
3. Insert image on Frame 3 of Layer 1
4. Insert image on Frame 4 of Layer 1

Now you delete a layer. Since deleting a layer is not an undoable action, it is not added to the stack. Now you hit the undo action. It finds the last undoable action, which in this case is "Insert image of Frame 4 of Layer 1", and attempts to undo it. It does this by reverting Frame 4 on Layer 1 to what it was before the image was added. However, Layer 1 no longer exists, so it doesn't know how to modify that frame and it crashes.

[...]

Issue #1120, opened 2018-11-01
Undoing Selection Flip X / Y will delete the drawing


Jose-Moreno:
### Issue Summary

Testing the program I forgot to report this. Basically if you select an area and use the Selection Flip X / Y function and then undo that operation, the drawing will be deleted. You can redo the drawing being deleted to the flipped state though.

Current workaround is flipping the selection again in the same axis

### Expected Results
Flipping Selection should be un-doable /re-doable

### Steps to reproduce

1. Draw something
2. Press CTRL + A or Create a Box Selection over the strokes
3. Go to Edit > Selection > Flip x/y
4. Remove selection. You can keep it too, it doesn't matter.
5. Press CTRL + Z. Drawing disappears.

### System Information
+ Pencil2D Version:
Version: 0.6.2
commit: 6471882892c0d34ac20b0a70b9591685c9f68d04
date: 2018-10-12_11:17:33
Development build
Operating System: Windows 7 SP 1 (6.1)
CPU Architecture: x86_64

CandyFace:
I'll check if I can reproduce this in the undo-redo rewrite branch.

Issue #1098, opened 2018-10-06
Selection Flipping is not undoable / redoable


Jose-Moreno:
### Issue Summary

A very simple user experience issue / bug. when flipping a selection either on the X or Y axis, this command is not undo-able / redo-able.

### Expected Results
Due to the nature of art, most if not all drastic changes in a drawing should be undoable. Flipping a drawing changes it's direction and sometimes artists try to find the best way to present them.

Note: This is not a critical bug by any means, you can flip the drawing again instead of undoing, but from a functionality standpoint all of these commands should be undo-able.

### Steps to reproduce

1. Draw something
2. Enclose it with the selection tool
3. Go to Edit > Selection > Flip X / Y
4. Press Ctrl + Z or go to Edit > Undo

### System Information
+ Pencil2D Version:

Version: 0.6.2
commit: ecbe8537aff9e70e99e16797383d21aedf5e39fe
date: 2018-10-05_13:12:56
Development build
Operating System: Windows 7 SP 1 (6.1)
CPU Architecture: x86_64

chchwy:
will take a look after merging the Undo Refactoring PR

Jose-Moreno:
[2020] **Testing on 0.6.5** Currently it seems you can undo the selection flipping, but it's still a bit buggy. After flipping the selection, if you immediately undo while the selection marquee still enabled this will delete from view the drawing inside the selection.

I assume that since it also removes the selection, anything that it's inside is being treated as part of the selection by the Canvas Painter. 

So the drawing disappears, but if you redo again you'll see the drawing reappear again and the flip is undone, also a marquee selection appears along the drawing even though it should be treated as a separate event (it should required an extra undo/redo step).

PR #1486, opened 2020-10-23
Undo/redo bug fixes

J5lx:
CandyFace mentioned that #953 probably won’t get merged very soon, so I went ahead and killed a few bugs in the current undo implementation. The fixes are simple enough that it’s hard to justify letting our users wait for the rewrite.

- Fixed broken smudge undo (#1004)
- Fixed weird undo/redo behaviour for selection flipping (#1098, #1120) by adding explicit support for it
- Crash when undoing after deleting a layer (#1412)
  - The code for this is kinda long, but mostly just because its inflated with boilerplate code
  - The actual layer deletion undo support (#748) still has to wait for the rewrite, though
- Unwanted keyframes added when undoing strokes drawn on an unselected keyframe through an empty frame (#1429)


CandyFace (approved):
LGTM!

I tested the mentioned bugs and they all seem to be fixed.
Well done 👍

Issue #1469, opened 2020-09-16
Implement separate reset pan, rotation and zoom options


Jose-Moreno:
### Issue Summary
The keyword here is _separate_. There's been plenty of times when I'm working at a desired zoom factor either closing in or out towards the drawing, however once I tilt the canvas since I can't determine the exact angle to go back once I have to press **reset view** and it will rollback the canvas location, rotation and zoom, so then I have to relocate and zoom again.

This behavior also affects the camera layer and it's a bit more frustrating in the case when you are animating the transforms individually but you depend on the view to be reset for all transforms.

What I'm proposing however is not to get rid of the ability to reset all the transforms, but to allow the user to choose which ones should be reset so i see something like this:

- [x] Reset All (this is the current behavior, and we should keep it too)
- [ ] Reset Pan / Location
- [ ] Reset Tilt / Rotation
- [ ] Reset Zoom / Scale

It could either be that the reset view has some visual toggles perhaps on a widget or toolbar, or that we invoke the extra reset options through modifiers. 

In the menu actions we could have the Reset View as a menu category > and then have all the 4 options implemented.

PR #1487, opened 2020-10-24
Implement separate reset pan and rotation options

J5lx:
This PR implements separate actions for resetting pan (i. e. centring the view) and resetting rotation, as requested in #1469. I did not implement an option to reset zoom since we already have one (View→Zoom→100%). As suggested in the original issue, the default shortcuts are based on the “Reset All” shortcut (Ctrl+H) with different modifiers. I didn’t add any visual toggles because I think those would work best on a toolbar, which is about to be added by #1468. Since the reset zoom option already had a menu entry in the zoom submenu, I did not create a new submenu, but rather grouped the “reset rotation” action with the existing rotation actions and the “center view” action with the “reset all” action. But of course we can still choose to go with a little redundancy and add the extra menu instead if that turns out to be better.


CandyFace (approved):
LGTM!

J5lx:
Builds for testing: [Windows x86_64](https://drive.google.com/file/d/1ppCb75dn4OMJ0sfLikt0mWzFoQgk9xd9/view?usp=sharing), [Windows x86](https://drive.google.com/file/d/1CJVRmY9SUANuL2uP2mpaAbB4hBe4o49Z/view?usp=sharing), [macOS](https://drive.google.com/file/d/1uoLRWVCTrsZ492FmtTG5ZG75vQjs4rRl/view?usp=sharing), [Linux](https://drive.google.com/file/d/1Sv3vv67tE1q-K3tZ5WjOv7_4b9Necckp/view?usp=sharing)

Jose-Moreno:
@J5lx Alright I tested the build you provided. Thank you for that. I think both Center View and Reset Rotation settings work well. My only question is why a separate Reset Zoom was not included, considering it was part of the issue requirements? 

While the general **Reset** does take care of the zoom reset, it also defaults all the other properties. While drawing there are certain times where you want to reset only the zoom but not the tilt or the position of the canvas, particularly if you're drawing details and you want to zoom out quickly to gauge the whole picture and then zoom in again to continue drawing.

That'd be my only observation to an otherwise good job 👍

J5lx:
Maybe if it isn’t that intuitive we could also rename it to “100% (Reset)” or something like that. Does that sound like a good idea?

Jose-Moreno:
@J5lx Apologies didn't see your last comment. I think it's alright to leave it as it is for now and probe the community to see if there are better ways to organize these options with workflow efficiency and UX in mind.

PR #1489, opened 2020-10-24
1488 fix camera interpolation

### LINKED ISSUE #1488: Camera interpolation doesn't update
### Issue Summary
When you try to animate the camera movement, nothing happens. This only effects the widget afaik, interpolation in export works fine.

### Actual Results
Once you start doing a playback or moving to any of the interpolated frames, only the first change can be seen, the rest will be ignored.

### Expected Results
The interpolation should work on all affected frames.

### Steps to reproduce
1. Add a new camera frame on frame 30 or something
2. Zoom or pan the camera
3. scrub any of the in-between frames
Result: the interpolation is not shown.

### System Information
+ Pencil2D Version: master

+ Operating System:
+ Mac OS 10.15.6 (19G2021)

CandyFace:
fixes #1488 

To minimize painting bugs, we should keep all update within scribblearea, so that we can rely on updates always coming from inside, at least when it affects cache and the like.


chchwy:
Confirm it's fixed. Thanks!

PR #1492, opened 2020-10-29
Use thinner cross cursor

### LINKED ISSUE #489: Use thinner cross cursor for drawing
I feel like the cross cursor when drawing is rather thick, I'd propose to have a thinner cursor so that you can see better/more accurately where you are drawing, like:

[CursorCrossInv.bmp](https://www.listech.com/liscad/help/see/topics/images/CursorCrossInv.bmp)

J5lx:
Fixes #489


CandyFace:
The cross could become slightly thinner still, also I like center dots.
cross

While it may not appear rounded, Qt (on mac at least) Qt seems to render the cursor with linear interpolation, so the cross appears slightly rounded.

J5lx:
> I like center dots

Fair point, I’ll add it.

> slightly thinner still

The lines are all 1px wide plus a 1px white outline for dark backgrounds, it doesn’t get much thinner than that. Since you mention macOS and linear interpolation, could that be another HiDPI-related issue? Can you try calling setDevicePixelRatio(2) on one of the cursor QPixmaps and see if that helps (even if it makes the cursor smaller)?

CandyFace:
> The lines are all 1px wide plus a 1px white outline for dark backgrounds, it doesn’t get much thinner than that. Since you mention macOS and linear interpolation, could that be another HiDPI-related issue? Can you try calling setDevicePixelRatio(2) on one of the cursor QPixmaps and see if that helps (even if it makes the cursor smaller)?

It seems to be a mac thing maybe, the standard cursor is also interpolated, eg. not pixelated but slightly blurry when you zoom in with accessibility zoom, not Qt related. The black line is 1 pixel but with the outline it makes it possible to narrow the thickness of the white outline.

It's not important though, let's just add that center dot then I'll merge this.

J5lx:
I added the centre dot.

CandyFace:
Looks good to me, merged.

PR #1545, opened 2020-12-06
Restructure startup around application class

J5lx:
I recently found myself looking through the source code of Qt Designer, and I really liked the way their application startup is organised, so I wanted to try and bring some of that over to Pencil2D. I also split out the code for the command line interface into dedicated classes, since it’s currently a rather unwieldy blob of code in main.cpp. There’s a bunch of changes in this PR, but most of it is code that moved from one place to another.

The idea is that main.cpp is now only responsible for some basic initialisation and for bootstrapping the application. The application runs the command line parser and, depending on the result, starts either the CLI exporter or sets up the GUI.

I know this isn’t exactly the most important stuff, but I kinda wanted to give it a try. I’ll try to review some of the existing PRs or kill some bugs in return ^^

Edit: Looks like CI doesn’t agree with my code… I'll look into it…

Edit 2: Fixed!


CandyFace (commented):
Exposing MainWindow2 which is supposed to be the main entry for the GUI application is wrong, if possible the CommandLine application, should create its own Editor and use that as its core.

I know that's difficult though given that Scribblearea is also widget bound and used as part of our core module... Ideally this would require a bigger refactoring of dismantling our Core module from any GUI related elements, such that it could be used by a CLI tool or as a module for someone implementing their own interface.

Aside for one small comment I think it looks good.

core_lib/src/util/fileformat.cpp
```diff
@@ -63,3 +64,47 @@ QString retrieveProjectNameFromTempPath(const QString& path)
...
+            { "BMP", false },
+            { "MP4",  true },
+            { "AVI",  true },
+            { "GIF",  true },
+            { "WEBM", true },
+            { "APNG", true },
+        };
+
+    Q_ASSERT(formatMapping.contains(format));
+    return formatMapping[format];
```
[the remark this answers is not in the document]
J5lx: The idea is that the keys in this formatMapping correspond exactly to the values of the extensionMapping above. Poor man’s constants, basically. That’s essentially what I wanted to express with the assertion: If the exact key does not exist in the mapping, it doesn’t just indicate an unknown format, it indicates an error in the program logic. But you are absolutely correct that, as it is now, it is a very ugly solution; one inherited [from the original command line exporter](https://github.com/pencil2d/pencil/pull/1545/files/984da01a00d768e3c1331e4b6d91cd3248c9f532#diff-a68bbd02ff27204e90b2bcf21c852b2a42bd7921db32fb0169dca0661e388820L257-L295) prior to this commit. The reason I didn’t do anything about it is that our handling of non-native file formats in general is currently not very centralised, with custom code in several different places (for example here, in Object, MovieExporter, ExportMovieDialog…). Cleaning that up is on my todo list already, so instead of fixing this particular piece of code right now, I’d rather address the file format situation as a whole separately from this PR. Would that be okay with you or would you still prefer to have this particular part cleaned up sooner rather than later?

J5lx (commented):
> Ideally this would require a bigger refactoring

Exactly, 100% this. Not just ideally, even. As it is, implementing the command line exporter without at least *instantiating* a main window is straight up impossible, and trying to fix it is like going down a super-sized rabbit hole. (Trust me, I’ve been there!) But perhaps, one day…

Thanks for reviewing!

core_lib/src/util/fileformat.cpp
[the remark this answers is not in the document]
CandyFace: That sounds fine to me, in that case then I have no further issues, LGTM.

app/src/pencil2d.cpp
```diff
@@ -0,0 +1,130 @@
...
+}
+
+void Pencil2D::installTranslators()
+{
+    QSettings setting(PENCIL2D, PENCIL2D);
+    QString userLocale = setting.value(SETTING_LANGUAGE).toString();
+    QLocale locale = userLocale.isEmpty() ? QLocale::system() : QLocale(userLocale);
+    QLocale::setDefault(locale);
+
+    std::unique_ptr<QTranslator> qtTranslator(new QTranslator(this));
```
chchwy, line 105: Should a QTranslator be deleted after levaing the function? 
Also two memory management methods are mixed here while both `std::unique_ptr` and parent pointer are set at the same time. Usually only one is needed.

chchwy: Had a look at the QTranslator source code. A plain `QTranslator* translator = new QTranslator(this);` is good. `std::unique_ptr` is not necessary.

QTranslator calls `QCoreApplication::removeTranslator(this);` in the destructor, so we probably want to keep QTranslator alive rather than deleting in right away. Just leave the object deletion to its parent object.
https://code.woboq.org/qt5/qtbase/src/corelib/kernel/qtranslator.cpp.html#_ZN11QTranslatorD1Ev

chchwy: oh sorry, didn't notice that you use `qtTranslator.release()` to give away the ownership. Then it's ok now, My bad.

app/src/pencil2d.cpp
```diff
@@ -0,0 +1,130 @@
...
+    QString inputPath = parser.inputPath();
+    QStringList outputPaths = parser.outputPaths();
+
+    if (outputPaths.isEmpty())
+    {
+        prepareGuiStartup(inputPath);
+        return Status::OK;
+    }
+
+    auto mainWindow = new MainWindow2;
```
chchwy, line 70: Having `std::unique_ptr` here? probably making no difference in the context, just a good style though.

chchwy (approved):
Looks neat and tidy, thanks @J5lx.

PR #1547, opened 2020-12-07
Fix extra movement when moving vector objects

J5lx:
A pretty simple fix (once you figure out what’s going on), but I’d still prefer to have this reviewed because I’m not that familiar with the selection / move tool code.


CandyFace (approved):
Fix confirmed.

PR #1558, opened 2020-12-20
Replace AppVeyor and Travis CI with GitHub Actions

J5lx:
This is the GitHub Actions “port” that I’ve been working on. I just took care of some remaining issues and at this point it should be pretty much complete. That said, we have a relatively complicated CI setup, so let me know if I overlooked something. The following secrets need to be configured in the repository settings, analogous to AppVeyor / Travis CI:

- LINUX_X86_64_PARENT, MACOS_X86_64_PARENT, WINDOWS_X86_64_PARENT, WINDOWS_X86_PARENT: The Google drive parent folder ids for the nightly builds
- GOOGLE_PRIVATE_KEY: The JSON key used for Google Drive authentication, no escaping required
- DOCS_REPO_TOKEN: The GitHub token used to update the generated docs in the pencil-docs repo

In terms of build time:

- Linux builds take roughly the same time
- macOS builds are roughly 2-3 times faster
- 32-bit Windows builds are slightly faster
- 64-bit Windows builds are roughly twice as fast
- Plus, both Windows builds can now run concurrently rather than sequentially!


J5lx:
Don’t mind the failed AppVeyor build :P

chchwy:
Great! will review as soon as I can.

chchwy:
Tested on my own repo.

1. Mac & Linux build weren't uploaded. anything else needed other than GDrive parent folder IDs?
2. Window filenames: `pencil2d-winwin64-1-2020-12-22.zip`,  `pencil2d-winwin32-1-2020-12-22.zip`
3. How are these enviornment variabls used in the action/workflow? I searched the whole `.github` folder and can't not find a keyword `LINUX_X86_64_PARENT`. I can only see `upload-parent: linux_x86_64_parent` at `ci.yml` line 33. Aren't they case sensitive?

Generally I felt that the Github Action is rather unclear/complicated in configuration and result. Perhaps just lack of understanding.

J5lx:
> Mac & Linux build weren't uploaded. anything else needs to be set other than GDrive parent folder IDs?

I think you just overlooked them:

Screenshot_20201222_144256
Screenshot_20201222_144315

> Window filenames: `pencil2d-winwin64-1-2020-12-22.zip`, `pencil2d-winwin32-1-2020-12-22.zip`

You’re right, I must have broken those somewhere along the way. I’ll look into it.

> How are these enviornment variabls used in the action/workflow? I searched the whole .github folder and can't not find a keyword LINUX_X86_64_PARENT. I can only see upload-parent: linux_x86_64_parent at ci.yml line 33. Aren't they case sensitive?

Now that you mention it, no, they’re case-insensitive. But I suppose it would make sense to keep it consistent nonetheless. That line 33 which you found specifies the name of the secret to use, the actual secret is then retrieved in line 108.

> Generally I felt that the Github Action is rather unclear/complicated in configuration and seeing result. Perhaps just lack of understanding.

Lack of understanding is one possibility. The basic concept of GitHub Actions is quite different from “traditional” CI services like Travis or AppVeyor, and it certainly took me some time to “stop thinking the Travis way”. In fact, even now this Actions configuration is still quite Travis-y, so to say.

However, it’s possible that this mix of Travis and Actions approaches is what makes the configuration so unclear. Take the package creation step, for example. Right now it’s an action in the same repo which calls a shell script which then calls {linux,mac,win}deployqt, copies libraries, applies fixes and wraps it all up in a package. It’s roughly what the after-build.{sh,ps1} scripts currently do. Ideally, you’d have an external, standalone action which knows how to do all of this, is neat and self-contained and reusable, so you can just reference it in the main ci.yml and tell it “gimme package”. So as an application vendor you wouldn’t even have to concern yourself with all the nitty-gritty details. That’s how Actions is *meant* to work. Unfortunately, it’s not so easy in this case. linuxdeployqt doesn’t support GStreamer out-of-the-box, the windeployqt tool doesn’t care about OpenSSL, macdeployqt needs additional fixes after it is run, and then last but not least there’s still that FFmpeg part which is *entirely* specific to Pencil2D…

I believe that’s where a lot of the complexity comes from – it’s that, as of now, our current complex, scripting-heavy CI setup is just difficult to adapt to the kind of neat-and-tidy setup that GitHub Actions was designed to facilitate. It’s most likely *possible*, but it will take a bit more work than I want to invest into an *initial* working configuration. Until then, this setup works, is well-integrated with GitHub (duh!), and even faster than what we have now.

J5lx:
Alright, the Windows filenames are fixed again.

chchwy:
Thanks. I think this PR is basically good to go. 

BTW, can you please keep the `after-build.sh/ps1`? I can be handy if I want to make a mac legacy build or windows legacy build. It's good to have the ability to deploy locally without CI services.

J5lx:
Oh, I didn’t think those scripts were actually used for that considering how they rely on the specific CI environment. But sure, if it’s needed I can adjust the actions a little to make them more suitable for standalone use. I know git comes with bash on Windows, so would it be fine if I keep using bash on Windows? I find powershell somewhat cumbersome to work with.

chchwy:
At least the `after-build.ps1` is made intentionally to run both locally and on CI services. I appreciate the existing of deploy scripts when I want to do it manually, e.g., Mac/Windows legacy builds.

Powershell is a built-in part of Windows OS without needing to install anything else. That's why I used it in the first place. It's fine to use Bash for Windows if you feel comfortable. Probably I just need to update the `win-deploy.ps1` for the local deployment.

chchwy:
This PR is ready, just hold off until the next version. I will merge after v0.6.6 is released.

J5lx:
Should I still add the stuff with the scripts to this PR then (if I get it done in time) or should I make a separate PR for that?

chchwy:
Please add to this PR.

J5lx:
I adjusted the packaging script so that it can also be used locally.

J5lx:
I just resolved the merge conflicts. Now that 0.6.6 is out this should be ready to merge, right? @chchwy

chchwy:
Yes, let's do this.

J5lx:
@chchwy I just noticed GHA builds are failing during the package upload / docs repo checkout stage. Could it be that the secrets haven’t been set on this repo yet?

Also, AppVeyor reports failing builds as well, so it’d be great if you could activate its “Skip branches without appveyor.yml” setting (or deactivate AppVeyor altogether).

Issue #1557, opened 2020-12-19
Undo frame deletion double undo/crash


scribblemaniac:
### Issue Summary

Undoing a frame deletion can cause two actions to be undone, or a crash in some situations. This is likely the same issue as #843, although the steps to reproduce are more involved.

### Steps to reproduce

1. Create a new project.
2. Create a layer with at least two keyframes.
2. Delete the first keyframe. If you are using a bitmap layer, the first keyframe must be at a position > 1.
3. Undo the frame deletion. This will undo the frame deletion and the action before that, or it will crash if no other undoable actions were performed.

### System Information
+ Pencil2D Version: 6f0b75b, v0.6.5

+ Operating System: Ubuntu 20.04

+ RAM Size: 16 GB

+ Graphics Tablet: N/A

### Developer Notes

This crash is happening when `Editor::backup` is called before reverting the deletion. In that function it uses `getLast*ImageAtFrame`, which returns null if there isn't a previous frame. If it's null, the backup is skipped, however `Editor::undo` always decrements `mBackupIndex` whether or not the backup was successful. Thus it goes back two actions instead of one, or causes an out of bounds exception if it goes below 0.

Jose-Moreno:
@scribblemaniac Hi. Since this issue has been closed, just to make sure, when you're free can you check if the merged fix solves your original issue with this ticket? Thanks!

scribblemaniac:
Can confirm this has been fixed.

Jose-Moreno:
@scribblemaniac Excellent thanks a lot for testing! 🙇

PR #1559, opened 2020-12-21
Sound import/export bugfixes

### LINKED ISSUE #1448: Movie audio import on existing empty keyframe fail under specific conditions
### Issue Summary

If you try to import a movie audio on an existing keyframe created by either the sound import feature or by using the `Add Keyframe` actions, will lead the movie audio import to fail.

### Expected Results

Regardless of origin or feature set, it is expected that any kind of sound will be allowed to be imported in a sound layer at any given empty frame.

There seems to be a mismatch between how the Movie Audio and Sound files are treated inside Pencil2D, which should be happening considering those are converted after import AFAIK.

### Video or Image Reference

image
_Movie Audio not loading on empty keyframe created by the sound import / add frame actions_

### Steps to reproduce

1. Open Pencil2D
2. File > Import > Sound > Create a new layer when prompted > Once the file browser opens press **`Cancel`**. Note there's a keyframe created on frame 1
3. File > Import > Movie Audio >Browse for an mp4 or support movie file > Press **`Open`**
4. A loading bar will appear but note that the movie audio will not be loaded into the timeline and the keyframe will remain empty.

**Checking that the movie audio > sound file workflow is possible**

1. Open Pencil2D
2. Create a sound layer (movie audio does not automatically create one)
3. File > Import > Movie Audio >Browse for an mp4 or support movie file > Press **`Open`**
4. Movie audio is loaded. Now press the **`Remove frame`** button
5. File > Import > Sound > Browse for a sound sample > Press **`Open`** . Note that the sound will be correctly loaded into the keyframe container. However.
6. Now press the **`Remove frame`** button again.
7. Try to load the previous Movie Audio, and you will not be able to.

### System Information
+ Pencil2D Version:
Version: 0.6.5

date: 2020-08-01_01:32:15

CPU Architecture: x86_64
Qt Version: 5.12.9

### LINKED ISSUE #1449: Sound import creates unwanted keyframe before importing audio sample
### Issue Summary

Whenever you use the **File > Import > Sound** feature a new keyframe will be created before the actual sample is imported (just after the file explorer window appears), there's also inconsistent behavior when using the `Add frame` action on a sound layer, since canceling the file browser without importing a sample, will actually erase the previously created keyframe.

If you press the action by mistake, or you decide to go back without importing a sound, this keyframe will later become a stepping stone for a nastier bug found in issue #1448

This does not happen when using the **File > Import > Movie Audio** feature.

### Expected Results

Do not create a sound keyframe until you have actually decided to import a sound file (after pressing `Open`)

### Video or Image Reference
|**Before Import**|**During Import**|
|---|---|
|image|image|

### Steps to reproduce

**Empty keyframe creation when importing sound**

1. Open Pencil2D
2. File > Import > Sound
3. Create the sound layer you are prompted to
4. When the file explorer opens, note that a single keyframe will be created on the sound layer
5. If you press **`Cancel`** the empty keyframe will remain in the timeline.

---
**Empty keyframe creation when using the `Add Frame` action**
1. Open Pencil2D
2. Create the sound layer you are prompted to
3. Press the `Add frame` button on the timeline widget
4. When the file explorer opens, note that a single keyframe will be created on the sound layer
5. If you press **`Cancel`** the empty keyframe will be erased form the timeline

### System Information
+ Pencil2D Version:
Version: 0.6.5

date: 2020-08-01_01:32:15

CPU Architecture: x86_64
Qt Version: 5.12.9

### LINKED ISSUE #1452: New implemented sound import formats are not properly loaded.
### Issue Summary

In 0.6.5 new sound formats were added to allow for their import. However so far from my testing the following types don't work:
+ OGG
+ FLAC
+ WMA
+ AIFF (I have a sample that imports fine and another one that doesn't, probably a similar setting as the issue with )
+ AAC
+ OPUS (Opus files by default run disgustingly slow on Windows 10, but this problem aside, Pencil2D just won't load the samples)
+ CAF (Have not tested CAF but i'm pretty sure it doesn't work either)

The weird thing is that the files are actually stored in the PCLX / PCL data folders, and if I play back those sound, they are properly played by VLC. The only one that wasn't was the OGG file that I used.

### Expected Results

All provided sound formats should work, not only load into the file, but load in the timeline and be able to be played back and scrubbed normally like WAV or MP3 files.

### Video or Image Reference

image
_The single keyframes is a sign of limited success_

### Steps to reproduce

1. Open Pencil2D
2. File > Import > Sound >Create New Sound Layer
3. In the file explorer > Choose a sample for each of the mentioned filetypes. Please do note if certain filetypes work on macOSX or Linux distributions as I haven't tested that (e.g Opus)

### System Information
+ Pencil2D Version:
Nightly Build Sep  4 2020

commit: 09190a77649b69a7e5497768a1544762bfaf6e23
date: 2020-09-05_00:07:49

CPU Architecture: x86_64
Qt Version: 5.12.9

scribblemaniac:
This PR fixes multiple bugs related to importing and exporting audio. In particular it fixes:
- Different behavior between movie audio import and sound import (closes #1448, closes #1452)
- Exporting with empty sound frames causes the movie export to fail
- Fix empty sound keyframe being left when cancelling a sound import (closes #1449)
- Implement sound keyframe deletion undoing and fix empty keyframe being produced (note you may encounter #1557 when testing this, that has not been fixed and is unrelated)
- Fixed removing the wrong sound keyframe when importing a sound on an existing frame and cancelling
- Prevent empty keyframe from being created if a sound keyframe is removed and there are no other keyframes left on the layer

This should hopefully address all the issues some of our users were having adding sounds a certain way.

### DIFF
--- a/app/src/actioncommands.cpp
+++ b/app/src/actioncommands.cpp
@@ -162,9 +162,6 @@ Status ActionCommands::importSound(FileType type)
     layer = mEditor->layers()->currentLayer();
     Q_ASSERT(layer->type() == Layer::SOUND);
 
-
-    int currentFrame = mEditor->currentFrame();
-
     // Adding key before getting file name just to make sure the keyframe can be insterted
     SoundClip* key = static_cast<SoundClip*>(mEditor->addNewKey());
 
@@ -194,7 +191,6 @@ Status ActionCommands::importSound(FileType type)
 
     if (!st.ok())
     {
-        layer->removeKeyFrame(currentFrame);
         mEditor->removeKey();
         emit mEditor->layers()->currentLayerChanged(mEditor->layers()->currentLayerIndex()); // trigger timeline repaint.
     }
--- a/app/src/actioncommands.cpp
+++ b/app/src/actioncommands.cpp
@@ -203,6 +203,7 @@ Status ActionCommands::importSound()
 
     int currentFrame = mEditor->currentFrame();
 
+    // Adding key before getting file name just to make sure the keyframe can be insterted
     SoundClip* key = static_cast<SoundClip*>(mEditor->addNewKey());
 
     if (key == nullptr)
@@ -214,14 +215,13 @@ Status ActionCommands::importSound()
 
     QString strSoundFile = FileDialog::getOpenFileName(mParent, FileType::SOUND);
 
-    if (strSoundFile.isEmpty())
-    {
-        return Status::SAFE;
-    }
-
     Status st = Status::FAIL;
 
-    if (strSoundFile.endsWith(".wav"))
+    if (strSoundFile.isEmpty())
+    {
+        st = Status::CANCELED;
+    }
+    else if (strSoundFile.endsWith(".wav"))
     {
         st = mEditor->sound()->loadSound(key, strSoundFile);
     }
@@ -233,6 +233,8 @@ Status ActionCommands::importSound()
     if (!st.ok())
     {
         layer->removeKeyFrame(currentFrame);
+        mEditor->removeKey();
+        emit mEditor->layers()->currentLayerChanged(mEditor->layers()->currentLayerIndex()); // trigger timeline repaint.
     }
 
     return st;
--- a/app/src/actioncommands.cpp
+++ b/app/src/actioncommands.cpp
@@ -118,48 +118,10 @@ Status ActionCommands::importMovieVideo()
 
 Status ActionCommands::importMovieAudio()
 {
-    QString filePath = FileDialog::getOpenFileName(mParent, FileType::MOVIE);
-    if (filePath.isEmpty())
-    {
-        return Status::FAIL;
-    }
-
-    // Show a progress dialog, as this can take a while if you have lots of images.
-    QProgressDialog progressDialog(tr("Importing movie audio..."), tr("Abort"), 0, 100, mParent);
-    hideQuestionMark(progressDialog);
-    progressDialog.setWindowModality(Qt::WindowModal);
-    progressDialog.show();
-
-    MovieImporter importer(this);
-    importer.setCore(mEditor);
-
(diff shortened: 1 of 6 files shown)


CandyFace (approved):
LGTM 👍

CandyFace:
Although you mention #1557 as a reason for weird undo behaviours, I haven't experienced that but other inconsistent behaviors.

Currently you can make the application crash if you try to undo a removed soundclip, although it only seems to happen when you have added one soundclip. it does not seem to be caused by the double undo behavior.
This specific problem seems to be related to the Editor::undo() method subtracting indexes before calling restoreKey().

I've also experienced another weird behavior, though I can't make it happen consistently, in a certain situation, you can make the undo operation restore an empty keyframe on a wrong layer or simply undo a wrong index.
If I add a normal bitmap keyframe first, then draw on it, then add a soundclip and remove it.. when you undo, the soundclip will still be gone and the bitmap image will be empty. This was seemingly also not caused by a double undo operation.

I'm not sure we should be adding undo for restoring soundclips yet.

scribblemaniac:
> edit: ah... it might be related to the double decrementing you mention in #1557 afterall, even so though.. we should at least fix so it doesn't crash, if we intend to undo the soundclip deletion.

Yup it's the same issue. I absolutely agree that we should fix that issue and I've just added it to the 0.6.6 milestone, but I didn't want to include it with these changes because it isn't specifically a sound issue.

As for undoing sound deletion, I technically did not add this feature, it was already there but the file-path property was not being set so it was just leaving an empty keyframe. It crashes almost the same as before, with the only difference being that it was harder to trigger the crash before because an empty keyframe was created at frame 0 if there were no keyframes left after deleting a keyframe. As far as I can tell there's nothing particuarly special that needs to be done for sound keyframe deletion undoing because the sound file is kept in the temporary folder while the project is open.

PR #1562, opened 2021-01-02
Rework frame updating

### LINKED ISSUE #1548: [ISSUE] Pencil2D not able to update images on real time when importing them (regardless their extension)
### Issue Summary
When importing an image into a frame, pencil2D will not show it till you click on the canvas or hit another layer

### Steps to reproduce
Open Pencil2D
Go to bitmap layer and draw some lines on the first two frames.
Go to import image (select any kind of images, it also happens with GIF images) and import an image to frame #1
Now if you go to frame #1, the image is not there yet, only the lines you previously draw
Hitting canvas will refresh the content of that frame and will bring the imported image.

### System Information
+ Pencil2D Version: 9th Dec2020

+ Operating System:
WIN10pro

+ RAM Size:
32GB

+ Graphics Tablet:
HUION hd185

### LINKED ISSUE #1560: Copy / Pasting frame content into other frames does not update the canvas properly.
### Issue Summary

This issue is related to #1496 (fixed) and #1548 (open) 

As the title says, pasting content from external software or from other Pencil2D frames, will not update until you force refresh the canvas painter by selecting another layer or using the selection tool.

### Expected Results

Pasting content on new frames should refresh the canvas otherwise this can confuse the user into thinking that the software didn't import or copy their content properly.

### Video Reference
pencil2d_copy_pate_canvas_update_bug
_The reference uses onion skins to showcase there are other drawings, but this doesn't work whether onion skins are active or not_

### Steps to reproduce
1. Draw 1 frame
2. Select the drawing
3. Copy It with shortcuts or with the menu to make sure you've copied to clipboard
4. Create a new blank frame
5. Paste the content
6. Notice nothing appears on the canvas
7. Select any other layer aside from the drawing layer OR create a selection on the canvas
8. Your drawing will be updated.

### System Information
+ Pencil2D Version:

Development Build Dec 27 2020

Operating System: Windows 10 Version 1909
CPU Architecture: x86_64
Qt Version: 5.14.1

Commit: 414c0a23f11b0dfd7378267d21683292926b3527

CandyFace:
Although we were about to have the frame update situation under somewhat control, I found many cases where the entire cache doesn't have to be invalidated, thus I've added more signals and methods to the appropriate situations where we require more control of the cache. Furthermore I found a bug where if our current frame cache is relevant, we still don't paint the rendering stack properly, eg. missing onion skin.

All cache control is now placed privately inside ScribbleArea and only through the appropriate slots, will the cache be updated. UpdateFrame and updateCurrentFrame will call update() and only invalidate frame cache when a frame is modified.

Frame cache invalidation will happen mostly automatically through all tools via ScribbleArea::setModified() which invalidates the current frame cache and updates the frame.

When modifying a frame outside the context of Scribblearea, the emitter "Editor::frameModified" should be used instead. 

I've documented signals and slots to the best of my ability, so that it hopefully makes sense when to use what signal.

closes #1560
closes #1548

### DIFF
--- a/core_lib/src/interface/scribblearea.cpp
+++ b/core_lib/src/interface/scribblearea.cpp
@@ -1668,7 +1668,7 @@ void ScribbleArea::deleteSelection()
             Q_CHECK_PTR(bitmapImage);
             bitmapImage->clear(selectMan->mySelectionRect());
         }
-        invalidateAllCache();
+        setModified(mEditor->currentLayerIndex(), mEditor->currentFrame());
     }
 }
 
--- a/core_lib/src/interface/scribblearea.cpp
+++ b/core_lib/src/interface/scribblearea.cpp
@@ -335,7 +335,7 @@ void ScribbleArea::onSelectionChanged()
 
 void ScribbleArea::onOnionSkinTypeChanged()
 {
-    invalidateAllCache();
+    invalidateLayerPixmapCache();
     updateCurrentFrame();
 }
 
@@ -1569,21 +1569,24 @@ void ScribbleArea::setLayerVisibility(LayerVisibility visibility)
 {
     mLayerVisibility = visibility;
     mPrefs->set(SETTING::LAYER_VISIBILITY, static_cast<int>(mLayerVisibility));
-    invalidateAllCache();
+
+    invalidateLayerPixmapCache();
 }
 
 void ScribbleArea::increaseLayerVisibilityIndex()
 {
     ++mLayerVisibility;
     mPrefs->set(SETTING::LAYER_VISIBILITY, static_cast<int>(mLayerVisibility));
-    invalidateAllCache();
+
+    invalidateLayerPixmapCache();
 }
 
 void ScribbleArea::decreaseLayerVisibilityIndex()
 {
     --mLayerVisibility;
     mPrefs->set(SETTING::LAYER_VISIBILITY, static_cast<int>(mLayerVisibility));
-    invalidateAllCache();
+
+    invalidateLayerPixmapCache();
 }
 
 /************************************************************************************/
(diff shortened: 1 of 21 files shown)


core_lib/src/interface/scribblearea.cpp
```diff
@@ -274,60 +243,114 @@ void ScribbleArea::removeOnionSkinsCacheAround(int frameNumber)
...
+void ScribbleArea::onCurrentFrameModified()
 {
-    Layer* layer = mEditor->object()->getLayer(layerNumber);
-    if (layer)
+    onFrameModified(mEditor->currentFrame());
+}
+
+void ScribbleArea::onFrameModified(int frameNumber)
+{
+    mCurrentCacheInvalid = true;
```
chchwy, line 311: Question. Why raising a flag here rather than clearing the cache directly?

chchwy:
Thanks @CandyFace, this PR lays a good foundation of cache invalidation. Good work. 

While playing with this PR, I found some issues:

1. The PixmapCache is not used. The result of CanvasPainter is inserted in PixmapCache but never really used.
2. Some of the cases do need clear all cache rather than layers cache only. e.g., `viewChanged`, `layerChanged`, `OnionSkinTypeChanged` and `resize()`, the incorrect results are just hidden by point 1.
3. Canvas doesn't re-draw correctly after resizing the window.
4. Canvas doesn't re-draw after changing the layer visibility in Timeline.
 
Not all issues above are made by your code, I just list what I have found. I am going to merge this PR, and then will push fixes based on your changes.

core_lib/src/interface/scribblearea.cpp
CandyFace: Hmm good question, I guess it's a remnant from before where I had multiple places where mCurrentCacheInvalid was set to true but yeah it doesn't make sense anymore.

PR #1571, opened 2021-02-13
Fix cache rendering issue with onion skin

CandyFace:
When calling frameModified, the onion skin cache not would be updated, meaning the cache for the skinned frames would be invalid and cause the familiar residue effect...

Way to reproduce:
1. Add 2 frames or move while onion skin is enabled
2. move scrubber to the first frame you added
3. Draw on the frame
4. Now scrub to the last added frame and paint on it

result: notice the residue effect due to the onion skin cache not having been updated.

This PR fixes that problem by making sure the onion skin cache is updated when the current frame is modified.

### DIFF
--- a/app/src/actioncommands.cpp
+++ b/app/src/actioncommands.cpp
@@ -659,7 +659,7 @@ void ActionCommands::moveFrameForward()
     }
 
     mEditor->layers()->notifyAnimationLengthChanged();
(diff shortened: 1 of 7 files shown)


chchwy:
Confirmed it's fixed. thanks @CandyFace

PR #1575, opened 2021-02-20
fix ui rendering issue when using direct2d platform on windows. (-platform direct2d)

iynaur:
error message: EndDraw failed: 0x88990014
Some widgets show just blank with black color.


CandyFace (approved):
Looks good to me.

Thanks for the contribution 👍

PR #1577, opened 2021-02-21
Fix cache bug when drawing on previous keyframe

### LINKED ISSUE #1576: Undoing a stroke with "keep drawing on previous key-frame" leads to weird behaviour
### Issue Summary
This happens if "Preferences->Timeline->Drawing" is set to "Keep drawing on the previous keyframe".
On a bitmap layer, if you draw a stroke while the timeline is not on a keyed frame, and then undo it, several odd things happen:

### Actual Results

1) A keyframe is generated at the current position in the timeline.
2) The new keyframe is a duplicate of the previous keyframe, but with the stroke undone
3) The previous keyframe retains the stroke that was meant to be undone

It might have other consequences with other tools and layer types, I haven't tried.

### Expected Results

1) A new keyframe should not be created.
2) N/A
3) The stroke should be undone on the previous keyframe.

### Steps to reproduce

The above should be self-explanatory. If not, let me know and I'll provide a step-by-step.

### System Information
+ Pencil2D Version: v0.6.5

CandyFace:
Fix cache issue caused by onFrameModified not being called when drawing on previous keyframe, making the latest stroke invisible upon release event.

closes #1576

### DIFF
--- a/core_lib/src/interface/scribblearea.cpp
+++ b/core_lib/src/interface/scribblearea.cpp
@@ -960,44 +960,42 @@ void ScribbleArea::handleDrawingOnEmptyFrame()
     }
 
     int frameNumber = mEditor->currentFrame();
+    if (layer->getKeyFrameAt(frameNumber)) { return; }
+
+    // Drawing on an empty frame; take action based on preference.
+    int action = mPrefs->getInt(SETTING::DRAW_ON_EMPTY_FRAME_ACTION);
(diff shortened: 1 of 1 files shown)


CandyFace:
This has been sitting for quite some time and it's a very small change. There should be no side effects from calling onFrameModified(previousKeyFrame->pos()) in this case, except clearing the cache which will fix the bug. 

Therefore I will now merge the PR.

PR #1578, opened 2021-02-22
Generate binary translation catalogues automatically during build

J5lx:
As I proposed quite a while ago, with this change the binary qm files will be generated automatically during the build. That way, it no longer has to be done manually: just merge the PRs made by the Transifex, all done – the new translations will immediately be included in the next build. Plus, the qrc file for the translations will be generated automatically as well, so that no longer has to be updated manually either.

Back when I originally proposed this, there was the concern that this would waste development time by slowing down all builds. However, at least in Makefile mode, the files won’t even be regenerated at all unless the source files have been updated, which usually isn’t more than once a month. And even when they are regenerated, it only takes about one second. Therefore the build times shouldn’t be a problem IMO.

The EXTRA_TRANSLATIONS line in app.pro is needed for the qrc generation because we use a single translation catalogue even though our source code is split into two sub-projects.


CandyFace (approved):
Looks good to me and I don't see a reason why we wouldn't make this automatic either. Even if it should increase build time slightly, adding a few seconds extra doesn't matter enough to justify that the step shouldn't be automatic.

PR #1581, opened 2021-02-28
Fix more cache issues.

CandyFace:
Another PR, this one focuses on the ongoing cache regressions that returned after i reworked the logic...

What's added:
- frameModified on import image and movie
- frameModified on onCurrentFrameChanged, fixes camera interpolation not being visible.
- frameModified on undo/redo


app/src/actioncommands.cpp
```diff
@@ -109,6 +109,7 @@ Status ActionCommands::importMovieVideo()
     }
 
     mEditor->layers()->notifyAnimationLengthChanged();
+    mEditor->framesModified();
```
davidlamhauge, line 112: Here, and when you call framesModified() in editor.cpp, my compiler gives a warning: Missing emit keyword. 
I'm not certain how important it is, but I generally like to avoid warnings.

davidlamhauge (changes requested):
Almost everything works perfect.
Camera interpolation works only when you are on the camera layer. If you choose the Bitmap or Vector layer, the camera interpolation is not updated.
Besides that: Good work!

app/src/actioncommands.cpp
CandyFace: I don't get any warnings, which is odd.. however the emit keyword is simply syntactic sugar, so it shouldn't matter. I will however add it anyway.

CandyFace:
Interpolation issue should be fixed now and the review requests should have been addressed.

davidlamhauge (approved):
LGTM!

PR #1589, opened 2021-03-27
Onion skin panel touchups

scribblemaniac:
These are some small modifications to the onion skin panel with the primary goal of decreasing the minimum width of the panel, and cleaning up the underlying structure a bit. Summary of changes:
- Use QGroupBox to group together items in the panel. This is what this widget was intended for, but it does come with some subtle aesthetic changes that we will have to agree on.
- Remove unnecessary elements. There were some useless horizontal spacers, and a layout item which could be set as the layout of a QWidget object instead.
- Use a combination of the FlowLayout and HBoxLayouts for the distributed opacity section. Currently this is the widest section, and this allows the items to be arranged vertically only if there is not enough horizontal space.
- Remove the fixed minimum size constraint on the widget.
- Change the size hint of the vertical spacer so that a vertical scrollbar does not appear before an items start to go out of view.

Here is a comparison of the minimum configurations before and after.
Before:
Screenshot of onion skin panel before changes
After (min width):
Screenshot of new onion skin panel with a smaller width
After (min height):
Screenshot screenshot of new onion skin panel with same height

As you can see, it can now have a considerably smaller width. The group boxes actually add a little bit to the height and width though, but I think it's worth it for the improved readability of the panel overall.

### DIFF
--- a/app/src/onionskinwidget.cpp
+++ b/app/src/onionskinwidget.cpp
@@ -48,6 +48,7 @@ void OnionSkinWidget::initUI()
     // flow layout to reduce the minimum width
     FlowLayout *opacityLayout = new FlowLayout;
     opacityLayout->setAlignment(Qt::AlignHCenter);
+    opacityLayout->setContentsMargins(0, 6, 0, 6);
     ui->opacityGroup->layout()->removeWidget(ui->minOpacityGroup);
     ui->opacityGroup->layout()->removeWidget(ui->maxOpacityGroup);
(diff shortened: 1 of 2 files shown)


CandyFace (approved):
LGTM. there's a small mis-alignment with the two spinboxes when they're below each-other... Normally I'd fix it with a QGridLayout but since Grid layout doesn't play well with a flow layout, I say we ignore that small detail for now.

PR #1629, opened 2021-04-16
Fix keyframe undo that would cause crash or double undo in some cases (issue #1557)

### LINKED ISSUE #1557: Undo frame deletion double undo/crash

aeroc7:
Add boolean return value on backup function so that we can check if the backup was *actually* successful.

Fixes #1557

### DIFF
--- a/core_lib/src/interface/editor.cpp
+++ b/core_lib/src/interface/editor.cpp
@@ -199,7 +199,7 @@ void Editor::backup(const QString& undoText)
     }
 }
 
-void Editor::backup(int backupLayer, int backupFrame, const QString& undoText)
+bool Editor::backup(int backupLayer, int backupFrame, const QString& undoText)
 {
     while (mBackupList.size() - 1 > mBackupIndex && !mBackupList.empty())
     {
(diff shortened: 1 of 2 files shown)


CandyFace:
Hi @aeroc7

Thanks for your contribution! The code looks good to me, so let's just get it in.
There's a larger rework of undo/redo in the works which was started by me, but until that gets done, I think it's a good idea to fix the bugs we can with the current one we have.

Merging 👍

CandyFace (approved):
👍

scribblemaniac:
Thanks for the contribution and welcome!

PR #1630, opened 2021-04-18
Issue #1207 - Bucket features

### LINKED ISSUE #1207: Bucket Fill Tool Behavior Regression / Re-Implementation
### Issue Summary

Hi I've made a video to encompass two of the key features that should be brought back. Even [back then (2016)](https://github.com/pencil2d/pencil/issues/524#issuecomment-233756957) I've been consistently asking for the fill tool to have this option of a layer selection or "target"

Currently the two main regressions or re-implementations, considering the tool has been refactored quite a bit, would be:

- [ ] Color Bleed: Bring back the function to automatically color a different layer based on the active layer that is being colored (e.g lineart)
    - [ ] Add the option to select up to which layer would the color transpose itself. By default it would always **source** and **target** the active layer.
    - [ ] Add the option to color multiple layers with the same layer "filter" e.g Source layer can be one or many and Target layers can be one or many.
- [ ] Color Spill: Bring back the pixel color expansion when filling the same area
    - [ ] Add an option toggle so this is not default behavior
    - [ ] Add a pixel expansion slider, or use the color tolerance to determine how much it can "bleed"

I think ideally this should be implemented before the tracing & coloring PR #1178 mainly so David's implementation can support itself on this existing behavior and become further simplified in the coloring part and since we always try to strive for simplicity yet for the tools to work functionally and thoroughly, one of the main issues I've been having with the pipeline process is that the **thinning and the blending** portion might only be needed for the color separation part, if we had the **Color Bleed** and **Color Spill** functionalities back as I explain in the video, of course this is subject to discussion as well, but I can only see it as a powerful asset and one that's been consistently requested by actual artists using Pencil2D consistently for their work.

Let me know what you think.

### Video or Image Reference
https://youtu.be/VULg2ONoo9o

### System Information
+ Pencil2D Version: 0.5.4b

Some of the changes made to modify the original behavior were made in #687 by @CandyFace but I'll keep looking for other PR's that might have modified this further to help out have a better picture of the required work.

CandyFace:
Inspired by davids work I set myself to see if I could improve on his expansion algorithm, as I found it problematic that it would be slower the more it needed to expand. I had recently bumped into an interesting image processing algorithm that could tell me the nearest pixels to a given color. This meant that by only searching the image once I could find pixels closest to the color and afterwards simply threshold based on the expand value. The result of this means the algorithm takes the same amount of time to expand a given region as we still only search the image once, compared to looping through the images x times which is not so problematic if you work with a relatively small image or only expand using 2-3 pixels but any more will take significantly longer.

### PR content
I believe the PR fulfills most of the checkboxes in #1207 thus I think the issue can be closed after this PR has been merged.
What do you think?

Here's the UI when on bitmap layer
image

**New feature** - Fill to can be either:
- Current layer
- Layer below (old pencil behavior)

An example of how "Layer below" works:
fill-layer-below

**New feature**  - Reference can be either:
- Current layer
- All layers

An example of how "all layers" reference mode works:
fill-all-layers

The all layer functionality basically flattens all bitmap layers of the current keyframe before it is run through by our floodfill algorithm, thus we can use all layers as reference to fill and in the end decide where we want to fill to.

**Updated feature** - Color tolerance:
- Can be toggled on/off

**New feature** - Expand fill:
- Can be toggled on/off
- Easily changed using the slider

**New feature** - Drag to fill
fill-drag2

From a technical perspective I have moved all bucket related tool property functionality to its own widget.
The bitmap bucket logic has been moved to a separate BitmapBucket class so it doesn't clutter the tool.

Pre-compiled builds are available to test here:
https://drive.google.com/drive/folders/0B7zQuPZEO_64NlFsM0JTTmwzdlk

closes #1207


davidlamhauge:
I've tested this feature thoroughly and finds it working as expected. There were some minor issues, that CF has taken care of. 
This feature will certainly improve the bucket-fill quality of Pencil2D, and I look forward to see someone review and merge it.

Jose-Moreno:
FYI, i'm testing this. So far these are the issues I've found that I'll need to share a video to showcase the issues accurately so please hold off merging this before that. I'll provide the video before the weekend.

Jose-Moreno:
@CandyFace Alright here's the video I promised. I was perhaps too enthusiastic so apologies if I fumble some words.

https://youtu.be/qdeTicqTN6E

I'll summarize the main issues:
- [Bug] Clicking on a color area is not replacing the pixels when using new colors, probably due to a compositing mode mismatch. This is evidenced after noticing how the expansion keeps working when filling a layer without lineart.
- [UX] **Fill to: layer below** option only solves a particular problem where each lineart layer will have a color layer pair. To solve other problem patterns we need at least a "custom layer" target option.
- [UX] **Reference: All layers** option doesn't take into account layer visibility so it undermines the feature usefulness when handling many lineart levels.
- [Feature Request]: Implement as a future PR the capability to fill pixels on top (front) or below (back) current layer existing pixels. Right now as per the mentioned fill bug, it seems the back fill is working, so having the ability to select between back/front fill would be ideal.
- [Bug] Windows Ink is messing up with the application of the color while using a stylus. It seems to duplicate the tablet event. Mouse events work properly, and shutting down windows ink via the tablet drivers options also work properly. There seems to be a problem with Qt 5.15 at least, since other PR's i've tested (i.e David's Camera Interpolation) had a problem with right clicking through the timeline cursor.

CandyFace:
PR is ready to be reviewed

app/src/tooloptionwidget.cpp
```diff
@@ -151,6 +147,18 @@ void ToolOptionWidget::onToolPropertyChanged(ToolType, ToolPropertyType ePropert
 
 void ToolOptionWidget::setVisibility(BaseTool* tool)
 {
+    if (tool->type() == BUCKET) {
+        disableAllOptions();
+        if (mBucketOptionsWidget == nullptr) {
+            mBucketOptionsWidget = new BucketOptionsWidget(editor(), this);
```
chchwy, line 153: Personally, I prefer to create the widget at the initial stage. Especially when the widget will not be destroyed until the application ends.

core_lib/src/graphics/bitmap/bitmapbucket.cpp
```diff
@@ -0,0 +1,245 @@
...
+
+    if (properties.bucketFillToLayerMode == 1) {
+
+        auto result = findBitmapLayerBelow(initialLayer, initialLayerIndex);
+        mTargetFillToLayer = result.first;
+        mTargetFillToLayerIndex = result.second;
+    }
+
+    if (mTargetFillToLayer == nullptr) {
+        Q_ASSERT(true);
```
chchwy, line 59: Do you mean `Q_ASSERT(false)`? it's the case that should never happen, right?

core_lib/src/graphics/bitmap/bitmapbucket.cpp
```diff
@@ -0,0 +1,245 @@
...
+
+    if (mProperties.fillMode == 0 && qAlpha(mBucketColor) == 0)
+    {
+        // Filling in overlay mode with a fully transparent color has no
+        // effect, so we can skip it in this case
+        return false;
+    }
+
+    if (mTargetFillToLayer == nullptr) {
+        Q_ASSERT(true);
```
chchwy, line 87: same as the above comment.

core_lib/src/graphics/bitmap/bitmapbucket.cpp
chchwy: And looks like the `assert` shouldn't be put here as the `mTargetFillToLayer` could be null.

core_lib/src/tool/basetool.h
```diff
@@ -50,6 +50,11 @@ class Properties
     int   fillMode = 0;
     int   stabilizerLevel = 0;
     qreal tolerance = 0;
+    bool toleranceEnabled = false;
+    int bucketFillExpand = 0;
+    bool bucketFillExpandEnabled = 0;
+    int bucketFillToLayerMode = 0;
+    int bucketFillReferenceMode = 0;
```
chchwy, line 57: Looks like all these newly added modes are exclusively for the bucket tool. I wonder if these variables could be moved to somewhere else better? And use `enum` instead?

chchwy:
I think this PR is working as expected and generally good to go. I encountered once a bug that `bucketFillToLayerMode` value doesn't match the dropdown list item but I can't replicate it consistently. probably it only happens in a special scenario.

Before I merge this PR, I will make some minor code changes mentioned above. won't touch the main part.

Great effort, well done.

core_lib/src/tool/basetool.h
CandyFace: Yup I considered that too, maybe we should do that. I simply added it there since we have most of the tool properties in basetool.

CandyFace:
Sounds good, thanks for reviewing :) - I generally agree with your comments. I am away from the computer for the day, so apply the changes you see fit 👍

core_lib/src/graphics/bitmap/bitmapbucket.cpp
scribblemaniac: I'm assuming based off of the changes you made that you realize that it can't be null. editor->layers()->currentLayer() asserts the return is not null, so initialLayer can't be null. findBitmapLayerBelow will return initialLayer or another layer which is allso asserted not to be null.

chchwy: Yes, I finally got it :p

Issue #1032, opened 2018-08-01
Icons too big in FullHD on Manjaro


gpsqueeek:
### --Issue Summary--
Running Manjaro XFCE I cannot see the whole interface on my fullHD 13" screen because the icons are so big that the timeline is out of the screen

### --Actual Results--
The icons are fuzzy because they are represented bigger than they should, and the timeline is out of range of my mouse (and sight)

### --Expected Results--
I should be able to see the whole window correctly ; it actually works fine if I change my resolution : 1400x900 is perfect, 1600x900 has the big icons issue

### --Video or Image Reference--
Correct image with 1400x900 :
![manjaro_1400_900](https://user-images.githubusercontent.com/23551855/43510282-739f8a18-9575-11e8-89c0-c4f7deb5de5c.png)

Incorrect image with 1600x900 :
![manjaro_1600_900](https://user-images.githubusercontent.com/23551855/43510294-81503180-9575-11e8-8eb6-f48876013ef9.png)

### --Steps to reproduce--
Switch to fullHD (or 1600x900), then launch Pencil2D (if already launched in a working resolution and then we change the resolution, there is no problem)

### --System Information--
+ Pencil2D Version: 6.1.1 or pencil2d-linux-2018-07-26.AppImage

+ Operating System:
Manjaro Linux XFCE 64 bits (up-to-date rolling release today August 1st 2018 except x265, ffmpeg and thunar-archive-plugin)

+ RAM Size:
8Gb

orwek:
I'm seeing similar issues on my Huion GT-133 drawing monitor. Here's my specs:

Linux Mint 19.1 Cinnamon 
Huion Kamvas Pro 13 GT-133 (Screen Resolution 1920 x 1080)
Digimend Kernel Drivers v8
Pencil 2D ver 0.6.3

The problem seems to occur with the 16x9 ratio. I switched to 16x10 and it shows up correctly, but my screen flickers so it is unusable. If I switch to a 3x4 ratio, the icons go back to normal size, but then it is stretched on the monitor, and also unusable.

J5lx:
[...]

PR #1635, opened 2021-05-10
Upgrade Linux builds to Xenial and Qt 5.15

### LINKED ISSUE #1017: wrong cursor icon when using a tablet
### --Issue Summary--

When using a tablet, wrong cursor icons are displayed, for instance, I select the drawing pen, then I
move the cursor into the drawing area. While doing so I pass a window border, which changes the cursor to a "move border arrows" icon, the icon stays when I move the cursor into the drawing area. This makes it hard to aim at the exact position where you want to start painting.

As soon as I draw a pixel the icon corrects itself.

The screenshot tool would always capture a normal cursor and not the icon displayed, so I had to photograph my screen :)

image

### --Steps to reproduce--

Use your tablet pen to select a drawing tool, then move into the drawing area, observe the pen cursor/pointer icon.

### --System Information--
+ Pencil2D Version:

git revision 33a13d2aef2c31aa87c3a1428f412773b08397f1

+ Operating System:

Linux, Fedora 27

+ RAM Size:

4GB

+ Graphics Tablet:

Huion H610
---

J5lx:
[Time to ditch Trusty, it seems!](https://github.com/probonopd/linuxdeployqt/commit/58cbda4aeab0d71b859be59a3eef59df2dd415c6) The code changes in this PR aren’t that interesting, the main focus for review should be testing [the build](https://drive.google.com/file/d/1VtA0n4GEFSVVSodQIFhCT5chveHervtm/view) and making sure all those different dependency versions don‘t cause any regressions (especially for multimedia, since I switched from an unofficial PPA to Ubuntu’s official repositories there). Personally, I haven’t found any issues during my brief test, in fact it confirmed my suspicion that a newer Qt version would finally fix #1017 for good. With any luck, it might also fix some of the other Linux issues such as #1032 which we’ve never been able to reproduce. Still, I think it would be good to have it tested on more systems than just my own.

BTW, I noticed that the AppImage gets noticeably larger with this change. Probably due to additional dependencies being pulled in or something, but I haven’t been able to look into it yet. I don’t want to block this PR over that since this upgrade is necessary to get our Linux builds working again.

I also noticed that macOS runners on GHA actually have 3 cores rather than 2 like the Windows and Linux runners, so I adjusted the MAKEFLAGS accordingly.

### DIFF
--- a/.github/actions/install-dependencies/install-dependencies.sh
+++ b/.github/actions/install-dependencies/install-dependencies.sh
@@ -3,31 +3,31 @@
 setup_linux() {
   # Because of how bare-bones our docker image is
   echo "::group::Install prerequisites"
-  sudo apt-get -yq install software-properties-common
+  apt-get -yq install software-properties-common
   echo "::endgroup::"
 
   echo "::group::Add APT sources"
   for ppa in ppa:ubuntu-toolchain-r/test ppa:ubuntu-sdk-team/ppa \
-             ppa:git-core/ppa ppa:mc3man/trusty-media \
-             ppa:beineri/opt-qt597-trusty; do
-    sudo apt-add-repository -y "${ppa}"
+             ppa:git-core/ppa ppa:beineri/opt-qt-5.15.2-xenial; do
+    apt-add-repository -y "${ppa}"
   done
   echo "::endgroup::"
 
   echo "::group::Fetch APT updates"
-  sudo apt-get update -yq
+  apt-get update -yq
   echo "::endgroup::"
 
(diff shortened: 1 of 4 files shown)


chchwy:
I have no problem with the Qt version upgrade. If I recall correctly, we didn't move away from Ubuntu 14 is because AppImage refuses to package the binaries on newer versions of Ubuntu. If AppImage is fine now, that's a good move to upgrade.

chchwy:
Let's merge this and see how it goes.

PR #1638, opened 2021-05-20
Updated build instructions for linux

vrockz747:
Added instructions for ```broken packages``` error while installing using ```apt-get```

### DIFF
--- a/docs/build_linux.md
+++ b/docs/build_linux.md
@@ -28,7 +28,15 @@ Pencil2D relies on the %Qt application framework so you must install it before y
 
 Pencil2D must be built with the %Qt 5 framework. To install %Qt 5, run this command:
 
-    sudo apt install qt5-default qt5-qmake libqt5xmlpatterns5-dev libqt5svg5-dev qtmultimedia5-dev
+    sudo apt install qt5-default qt5-qmake libqt5xmlpatterns5-dev libqt5svg5-dev qtmultimedia5-dev qttools5-dev-tools
+    
+###NOTE 
+
+For some cases installing via  `apt-get` may give `broken packages error`, in such case try installing using `aptitude`:
+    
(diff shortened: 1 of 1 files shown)


vrockz747:
Please review this PR @J5lx

Jose-Moreno:
@vrockz747 I've updated the automation and labels for this PR. We'll now have to wait for a reviewer to merge this. Thanks again for your contribution.

scribblemaniac (commented):
I don't think this additional note should be included in our documentation. Quite frankly, these broken packages are likely an issue specific to your system and are not something that I would expected to happen commonly when using apt normally. Likely an installation of one of these packages or their dependencies failed, either during running this command for the first time, or potentially from a previous package installation attempt. For issues specific to one system, or that are not directly related to the setup/build process, it is best if we defer support to external resources such as StackOverflow.

vrockz747:
Makes sense @scribblemaniac, I will then close this PR

vrockz747:
Makes sense @scribblemaniac, should i just add this [reference] (https://askubuntu.com/questions/223237/unable-to-correct-problems-you-have-held-broken-packages) instead?

chchwy:
Please add it as a reference, please. :)

scribblemaniac:
We have discussed these changes further and have decided that while providing additional guidance to help users compile our program is welcome, we believe in this particular case that the issue you've encountered is too specific to warrant any inclusion in our documentation at this time. We can always revisit this later if we find other people experiencing similar issues.

Speaking personally now as a person who has used a Debian-based distro as his primary OS for many years now, you will encounter many issues like this where you just have to look them the error on StackOverflow or elsewhere online. A broken package could be caused by many things, none of which we have any control over. It could have been an issue with the package of one of the dependencies of this program, or it could have been any error during a previous package installation/update. This is not something I would expect to happen often when following the instructions outlined in our Linux compilation guide.

Sorry for taking so long to deal with this relatively minor PR. Things sometimes move very slowly around here. Thank you for your contribution.

PR #1647, opened 2021-07-13
Run tests without window system integration

J5lx:
As long as we don’t do anything in the tests that actually requires window system integration, this should be enough to make the tests run on CI. Otherwise we’ll need to run an actual display server for the Linux tests, which seems like something of a hassle.


chchwy:
Thanks!

PR #1649, opened 2021-07-20
Fixed duplicate word (line 41)

kant:


chchwy:
Thanks. It also reminds me that the build guide is outdated. I will update it.

PR #1655, opened 2021-08-07
Implement device pixel ratio for canvas

MrStevns:
This will make anything painted on the canvas pixmap much crisper, on high resolutions screens.

The change will be noticeable as soon as we begin to draw more elements eg. (the work done in #1587), it might also improve the overall image quality of the drawings on said screens.

### DIFF
--- a/core_lib/src/interface/scribblearea.cpp
+++ b/core_lib/src/interface/scribblearea.cpp
@@ -761,10 +761,9 @@ void ScribbleArea::mouseDoubleClickEvent(QMouseEvent* e)
 void ScribbleArea::resizeEvent(QResizeEvent* event)
 {
     QWidget::resizeEvent(event);
-    const qreal dpr = devicePixelRatioF();
-    mCanvas = QPixmap(QSizeF(size() * dpr).toSize());
+    mDevicePixelRatio = devicePixelRatioF();
+    mCanvas = QPixmap(QSizeF(size() * mDevicePixelRatio).toSize());
     mCanvas.fill(Qt::transparent);
-    mDevicePixelRatio = dpr;
 
     mEditor->view()->setCanvasSize(size());
 
--- a/core_lib/src/interface/scribblearea.cpp
+++ b/core_lib/src/interface/scribblearea.cpp
@@ -761,8 +761,10 @@ void ScribbleArea::mouseDoubleClickEvent(QMouseEvent* e)
 void ScribbleArea::resizeEvent(QResizeEvent* event)
 {
     QWidget::resizeEvent(event);
-    mCanvas = QPixmap(size());
+    const qreal dpr = devicePixelRatioF();
+    mCanvas = QPixmap(QSizeF(size() * dpr).toSize());
     mCanvas.fill(Qt::transparent);
+    mDevicePixelRatio = dpr;
 
     mEditor->view()->setCanvasSize(size());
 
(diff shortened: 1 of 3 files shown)


core_lib/src/interface/scribblearea.cpp
```diff
@@ -761,8 +761,10 @@ void ScribbleArea::mouseDoubleClickEvent(QMouseEvent* e)
 void ScribbleArea::resizeEvent(QResizeEvent* event)
 {
     QWidget::resizeEvent(event);
-    mCanvas = QPixmap(size());
+    const qreal dpr = devicePixelRatioF();
+    mCanvas = QPixmap(QSizeF(size() * dpr).toSize());
     mCanvas.fill(Qt::transparent);
+    mDevicePixelRatio = dpr;
```
J5lx, line 767:
```suggestion
    mDevicePixelRatio = devicePixelRatioF();
    mCanvas = QPixmap(QSizeF(size() * mDevicePixelRatio).toSize());
    mCanvas.fill(Qt::transparent);
```

J5lx (changes requested):
One small suggestion, otherwise it looks good to me. Looks good on my “normal dpi” monitor as well. I don’t have any HiDPI hardware to test this on, but if you’re confident this works as intended I’d say it’s fine to merge after applying the suggestion. Otherwise maybe scribble can help, since he has a HiDPI monitor as well IIRC.

CandyFace:
I am only confident in what I've tested on, which is Mac with high dpi screen. For that I can say that it works as intended. I discovered the bug when working on another feature and applying it, fixed my issue.

J5lx (approved):
With that, LGTM. If you’d like for the changes to be tested on other kinds of HiDPI systems, let’s simply wait for someone who has that kind of hardware.

scribblemaniac:
I will test this with a high-dpi device on Linux and Windows when I get a chance. The code looks alright as long as scaling doesn't get applied a second time by some qt thing, which would definitely be noticeable in testing. My only concern would be what impact this would have on performance when using a high-dpi display? @CandyFace have you noticed any degradation in playback or drawing performance on your high dpi device?

CandyFace:
Not really no, although the change does make the pixmap size bigger, so it would not be wrong to assume that it will affect performance on high dpi monitors, but that should be expected imo, since you've chosen to use a high dpi monitor, you should get the visual benefits from that.

When that's said, I haven't noticed anything that's gotten slower after applying the change, but I haven't done any benchmarks either.

chchwy:
Just ran this PR on my Windows 10 laptop with device pixel ratio 2.0. Qt 5.15.2.

I noticed that there is a small stroke shift when I release the mouse button. Like a one pixel shift, some times goes left, some times goes up. I suspect something went wrong in the position calculation when pasting the canvas buffer to the actual key frame.

CandyFace:
Are you sure that is introduced by this PR? I recall that having existed for a long time. Can't say I've noticed that. I will see if i can reproduce

chchwy:
@CandyFace you are right. the behaviour exists in the master branch as well. it's not related to this PR.

MrStevns:
Actually, the issue with the selection tool being offset is not caused by this PR. It's found in master too.

It's been fixed now though.

scribblemaniac (approved):
Ah I don't normally use overlays, so I did not notice this issue on master, gj. Everything I said still stands, but I'm ready to merge this now.

PR #1664, opened 2021-09-24
Upload nightly builds only if key is available

J5lx:
Should hopefully fix the failing PR builds. Let’s see if this works.


J5lx:
Looks good! Since this is such a trivial change and it works as intended, I’m just gonna merge this.

PR #1666, opened 2021-09-29
Create no vector layer by default

J5lx:
Something I discussed with Jose on Discord a few days ago. We keep directing people away from vector since it’s in a bad state, so let’s remove the vector layer from the default setup until such time as we get it into shape again. Users who want to keep the default vector layer can enable it in the preferences.


J5lx:
Hmm I think I’m still gonna move some code around here…

J5lx:
Good to go now. I got rid of `Object::createDefaultLayers`, IMO the concept of default layers doesn’t make much sense in core_lib.

MrStevns:
Hmm what about a default camera layer though? We have all sorts of code that expects the camera layer to be there, otherwise the application will crash. So does it make sense to make the GUI always do that when it's a requirement for the core to work properly?

J5lx:
That’s a fair point, although it’s not directly related to the changes in this PR. The only thing that’s changed is that now the client needs to call addNewCameraLayer() instead of createDefaultLayers(), which is essentially a zero-sum change in that regard. It’d probably still be a good idea to better document and enforce requirements like that, but I’m not sure if it’s in scope for this PR. Thoughts?

Jose-Moreno:
@J5lx When you mention keeping the vector layer and enabling it in preferences you're refering to the project presets? or are you going to remove the ability to add a vector layer? If it's the former I have no issue with it. If it's the former I'd say it's more hassle to remove everything and activate it via preferences. 

We shouldn't force users who are actually using the vector system to go out of their way to re-enable it. I think a simple change in the default preset, and giving them an additional preset with the original layout is enough.

J5lx:
Don’t worry, users can always add new vector layers. This PR only changes whether a vector layer is included in the *default* setup that you get when you create a new blank project.

scribblemaniac:
I fully support removing the vector layer from the default startup project. However I don't think that this requires a new setting for add it back. If someone wants a vector layer in their default file, they can create a preset for that, that's what the whole preset system is for. I don't have a strong opinion for where the code for the default layers should be currently, there are points for both sides.

J5lx:
> I don't think that this requires a new setting for add it back

That’s something we talked about actually since that was my thought as well:

> [20:38] JoseMoreno: I mean the current one is called default. We could ship two called "Bitmap Animation" and "Vector Animation" and leave the default as Bitmap.
> [20:38] JoseMoreno: new users won't know better and veteran users know they can add it back
> [20:39] JoseMoreno: at least until the titans have been defeated
> [20:39] JoseMoreno: and we can tackle the vector system as a whole
> [20:41] Jakob: Do you think we really need a vector animation template with the current state of the vector engine?
> [20:41] Jakob: I feel like that is also something that could trick users into using the vector engine
> [20:41] JoseMoreno: while it's subpar there are people that still use it
> [20:41] Jakob: users who don’t know about the state of the vector engine
> [20:42] Jakob: okay
> [20:42] JoseMoreno: last time I had a ... small quibble with user that is dedicated to preach about Pencil2D's vector system
> [20:42] JoseMoreno: saying it's more useful than what we let out
> [20:42] Jakob: haha
> [20:42] JoseMoreno: I got mad and told them that were didn't want to promote the use of WIP tech
> [20:42] Jakob: maybe we should just call it “Vector Animation (experimental)” or something then (edited)
> [20:42] JoseMoreno: that's fine

chchwy:
> I fully support removing the vector layer from the default startup project. However I don't think that this requires a new setting for add it back. If someone wants a vector layer in their default file, they can create a preset for that, that's what the whole preset system is for.
I second that.

Issue #1684, opened 2021-12-31
Camera layer add/delete frame crashes Pencil2D


Jose-Moreno:
### Issue Summary

As reported by @davidlamhauge in discord, there's an issue where adding and deleting frames will cause the software to crash 100%

According to David this is related to how the interpolation between two frames is handled. I'll quote his comment from the server.

**David Says:**
> I found a bug in the master branch, If you start the program, choose the Camera layer, add a frame, go to frame 1, press Delete Frame and then press Add Frame, the program crashes.

> I've looked into it, with fresh eyes, and found out that the problem is in **layercamera.cpp** in the function `linearInterpolateTransform(Camera* cam)`.
> To make the interpolation we create **camera1** and **camera2**, which both can be `nullptr`. After they are created we try to set their **EasingType**, which of course will make a segmentation fault if they are `nullptr`, and the program crashes. 
This code works:
```cpp
    int frameNumber = cam->pos();
    Camera* camera1 = static_cast<Camera*>(getLastKeyFrameAtPosition(frameNumber - 1));
    if (camera1 != nullptr)
        camera1->setEasingType(camera1->getEasingType());

    int nextFrame = getNextKeyFramePosition(frameNumber);
    Camera* camera2 = static_cast<Camera*>(getLastKeyFrameAtPosition(nextFrame));
    if (camera2 != nullptr)
        camera2->setEasingType(camera2->getEasingType());
```
_Note: The Highlights are mine_

### Expected Results

Pencil2D should not crash by adding or removing frames. Since this seems to be related to a null pointer exception, the source code should also not lead to them (obviously)

### Steps to reproduce

1. Start the program
2. Select the Camera layer
3. Add a frame (playhead will move to frame 2nd)
4. Go to frame 1
5. Press the `Delete Frame` button
6. Press `Add Frame` button
7. The program crashes.

### System Information
Tested on the latest Nightly. David tested it on master as well.

+ Pencil2D Version:
Nightly Build Nov 14 2021

commit: 91b1721baef012d49cffd0d23358524a420af9fd
date: 2021-11-14_16:46:59

Operating System: Windows 10 Version 2009
CPU Architecture: x86_64
Qt Version: 5.15.2

Jose-Moreno:
I can confirm the PR merge fixed the specific crash. Thanks! @davidlamhauge @scribblemaniac

PR #1688, opened 2022-01-02
Removed two lines of code that made the software crash #1684

davidlamhauge:
Realized that since the camera is born with a EasingType, there was no need to add it, so the two lines of code, that made the software crash under certain circumstances, have been deleted.
The function containing the two lines of code is heavily improved in a PR #1587, and this PR is hopefully merged into the master branch soon, so just deleting the two lines of code if sufficient now, in my opinion.
Solves #1684


scribblemaniac (approved):
Looks good. Pointless lines and should not be called if camera1 or camera2 can be nullptr. Thanks!

PR #1698, opened 2022-03-01
Add mimetype to pclx files

scribblemaniac:
This includes a special "mimetype" file at the beginning of the zip archive of pclx files containing the mime type used for pclx files. Many common zip-based file formats (epub, openraster, openoffice, krita, and more) use this specific approach to help distinguish their files from other zip files. This provides a unique file signature for pclx projects and allows them to be identified by other programs such as the "file" utility on linux, and possibly file association on other platforms as well. This should not impact the reading of project files in any way since the file is outside the data directory and in fact won't even appear in the temporary directory.

### DIFF
--- a/core_lib/src/qminiz.cpp
+++ b/core_lib/src/qminiz.cpp
@@ -39,7 +39,7 @@ bool MiniZ::isZip(const QString& sZipFilePath)
 }
 
 // ReSharper disable once CppInconsistentNaming
-Status MiniZ::compressFolder(QString zipFilePath, QString srcFolderPath, const QStringList& fileList)
+Status MiniZ::compressFolder(QString zipFilePath, QString srcFolderPath, const QStringList& fileList, QString mimetype)
 {
     DebugDetails dd;
     dd << QString("Creating Zip %1 from folder %2").arg(zipFilePath, srcFolderPath);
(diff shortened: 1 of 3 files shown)


scribblemaniac:
Apparently fmemopen is not a thing on Windows. I will have to look at some alternate way of doing this on Windows.

scribblemaniac:
Alright, I fixed it with a different solution. Should be cross-platform now.

MrStevns:
Regarding the miniz changes, wouldn't it be better to apply these changes to the fork we have and then bump the version? https://github.com/pencil2d/miniz

scribblemaniac:
There are no changes to miniz here, only to the QMiniz wrapper which is not a part of that fork. I initially did it this way to make updating the library as simple as possible, but with some of the latter changes I did, it ended up being C++ specific, so it could not be done this way in the miniz library which is purely c-based.

Looking at that fork, there doesn't seem to be any reason for us to use that anymore. There's only one different commit, which claims to change fopen_s for _wfopen_s. However, it doesn't actually appear to do that. Moreover, it appears that change has already been done upstream.

MrStevns:
🤦‍♂️ ah you're right, I failed to see that it was the wrapper you modified. I have no objections in that case 👍

chchwy:
I just realized the miniz source in https://github.com/pencil2d/miniz is not exactly the code we use in the mina repo (damn it that was my bad).

I made two changes in our midified version: 

  1. replace `fopen_s` by `_wfopen_s` for Unicode file path support. 
  2. convert the UTF8 file paths to UCS2 because that's the character encoding used by `_wfopen_s`

As for what @scribblemaniac said, yes, the upstream has implemented the same thing in [this PR](https://github.com/richgel999/miniz/pull/203/files) as what I did. We should consider using the upstream implementation instead of ours. However, the PR is not in a released version yet.

chchwy:
I have tested this PR and it works as expected. Going to merge.

scribblemaniac:
@chchwy Thanks for looking into that.

MrStevns:
FYI after this change, I can't unzip the project on mac os anymore with the standard "archive utility"
image

PR #1701, opened 2022-03-22
Sound clip count warning

scribblemaniac:
Currently Pencil2D can't export animations with >= 63 frames due to limitations with the ffmpeg filter inputs. This warns the user when they've added too many sound clips, and produces a more specific error when attempting to export an animation with too many sound clips.

This could be improved, but I think it's good enough for a temporary solution. I have a movie exporter rewrite partially finished which I would like to complete before making changes to the ffmpeg command to allow for any number of sound clips to be exported. This could be a while out though, so I think this is a reasonable solution in the mean time.

### DIFF
--- a/app/src/actioncommands.cpp
+++ b/app/src/actioncommands.cpp
@@ -191,6 +191,13 @@ Status ActionCommands::importSound(FileType type)
         mEditor->removeKey();
         emit mEditor->layers()->currentLayerChanged(mEditor->layers()->currentLayerIndex()); // trigger timeline repaint.
     }
+    else
+    {
+        int clipCount = mEditor->sound()->soundClipCount();
+        if (clipCount >= 63) {
+            showSoundClipWarning(clipCount);
+        }
+    }
 
     return st;
 }
@@ -721,6 +728,10 @@ void ActionCommands::duplicateKey()
     if (layer->type() == Layer::SOUND)
     {
         mEditor->sound()->processSound(dynamic_cast<SoundClip*>(dupKey));
+        int clipCount = mEditor->sound()->soundClipCount();
+        if (clipCount >= 63) {
+            showSoundClipWarning(clipCount);
+        }
     }
     else
     {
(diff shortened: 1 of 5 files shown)


app/src/actioncommands.cpp
```diff
@@ -191,6 +191,13 @@ Status ActionCommands::importSound(FileType type)
         mEditor->removeKey();
         emit mEditor->layers()->currentLayerChanged(mEditor->layers()->currentLayerIndex()); // trigger timeline repaint.
     }
+    else
+    {
+        int clipCount = mEditor->sound()->soundClipCount();
+        if (clipCount >= 63) {
```
MrStevns, line 197: I think we should keep a constant with this value, instead of having it three places, even if it's meant as a temporary solution.

scribblemaniac: Agreed, changed.

scribblemaniac:
This is going to conflict with #1690 so I think merging this PR should wait until that PR is merged and I have had a chance to add this check when duplicating the sound layer as well.

app/src/actioncommands.cpp
```diff
@@ -234,6 +241,15 @@ Status ActionCommands::exportGif()
 Status ActionCommands::exportMovie(bool isGif)
 {
     FileType fileType = (isGif) ? FileType::GIF : FileType::MOVIE;
+
+    int clipCount = mEditor->sound()->soundClipCount();
+    if (fileType == FileType::MOVIE && clipCount >= MovieExporter::MAX_SOUND_FRAMES)
+    {
+        ErrorDialog errorDialog(tr("Something went wrong"), tr("You currently have a total of %1 sound clips. Due to current limitations, you will be unable to export any animation exceeding %2 sound clips. We recommend splitting up larger projects into multiple smaller project to stay within this limit.").arg(clipCount).arg(MovieExporter::MAX_SOUND_FRAMES), QString(), mParent);
```
chchwy, line 248: Any reason that not using `showSoundClipWarning()` here?

MrStevns: The former is a warning dialog, the latter is an error, so I think it's fine

J5lx:
Doesn’t this check still need to be added when duplicating a sound layer as well, as scribble said?

MrStevns:
> Doesn’t this check still need to be added when duplicating a sound layer as well, as scribble said?

Agh... I managed to skip that part, yes it should.

PR #1702, opened 2022-03-22
Check for multiple application instances

scribblemaniac:
On startup this will check if there is another instance of Pencil2D already running. If there is, the user will be warned of the issues that can arise from running multiple instances of Pencil2D simultaneously (as a result of the shared temporary folder directory and shared preferences file).

This check is done by creating a lock file and lock it on startup. If the locking fails, the warning is shown. If Pencil2D crashes, the lock file will remain and will be stale, however Qt has some checks under the hood that will check for running processes and match the process id and name so that in this case, the next instance of Pencil2D will still be able to successfully acquire the lock and no warning will be produced.

Screenshot from 2022-03-21 22-13-28

### DIFF
--- a/app/src/main.cpp
+++ b/app/src/main.cpp
@@ -39,9 +39,11 @@ int main(int argc, char* argv[])
 
     Pencil2D app(argc, argv);
 
+    #ifndef QT_DEBUG
     if (app.isInstanceOpen()) {
         return EXIT_SUCCESS;
     }
+    #endif
 
     switch (app.handleCommandLineOptions().code())
     {
--- a/app/src/main.cpp
+++ b/app/src/main.cpp
@@ -38,6 +38,11 @@ int main(int argc, char* argv[])
     initCategoryLogging();
 
     Pencil2D app(argc, argv);
+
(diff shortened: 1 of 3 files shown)


MrStevns:
Tested on Mac OS and works as expected.
Maybe we should check this for release builds only though? I'm often launching multiple instances of Pencil2D to test various things, I'd still like to be able to do that without having to comment out that piece of code or interacting with that popup on each launch. What do you think?

scribblemaniac:
Perhaps it could be disabled from builds from source, but I think it should still run the check for nightly builds. An average user could conceivably try to run multiple nightly builds or a nightly build and a release build at the same time. Would this work for you?

MrStevns:
> Perhaps it could be disabled from builds from source, but I think it should still run the check for nightly builds. An average user could conceivably try to run multiple nightly builds or a nightly build and a release build at the same time. Would this work for you?

Sounds good to me 👍

scribblemaniac:
Thanks @MrStevns. I forgot about this and indeed have been very very busy. You change looks good to me.

PR #1709, opened 2022-04-28
Fix: app carshes when duplicating a layer that has unloaded bitmap key frame

chchwy:
This is a PR to address the issue mentioned here: 
https://github.com/pencil2d/pencil/pull/1690#pullrequestreview-926936260

Actually ,the app could crash whenever calling `BitmapImage::clone()`  on an unloaded bitmap keyframe.
My solution is to make a copy of the original image file with a temporary filename if the image is unloaded.


MrStevns (approved):
Changes looks good to me, although I wasn't able to reproduce the crash that was the reason for the PR, it did however fix a load problem where the unloaded frames wouldn't be shown if you set the visibility of the layer to hidden and then duplicated it.

PR #1710, opened 2022-05-05
Filemanager improved error handling

MrStevns:
This PR makes a few changes to the filemanager, improving the error handling and fixing a few situations where the project could be corrupted and a backup wouldn't be made.

### DIFF
--- a/core_lib/src/structure/filemanager.cpp
+++ b/core_lib/src/structure/filemanager.cpp
@@ -558,25 +558,38 @@ void FileManager::handleOpenProjectError(Status::ErrorCode error, const DebugDet
     removePFFTmpDirectory(mstrLastTempFolder);
 }
 
+int FileManager::countExistingBackups(const QString& fileName) const
+{
+    QFileInfo fileInfo(fileName);
+    QDir directory(fileInfo.absoluteDir());
+    const QString& baseName = fileInfo.completeBaseName();
+
+    int backupCount = 0;
+    for (QFileInfo dirFileInfo : directory.entryInfoList(QDir::Filter::Files)) {
+        QString searchFileBaseName = dirFileInfo.completeBaseName();
+        if (baseName.compare(searchFileBaseName) == 0 && searchFileBaseName.contains(PFF_BACKUP_IDENTIFIER)) {
+            backupCount++;
+        }
+    }
+
+    return backupCount;
+}
+
 QString FileManager::backupPreviousFile(const QString& fileName)
 {
     if (!QFile::exists(fileName))
         return "";
 
-    QFileInfo info(fileName);
-    int backupCount = 1;
-    QDir directory(info.absoluteDir());
-    QString backupString = "backup";
-    for (QFileInfo info : directory.entryInfoList(QDir::Filter::Files)) {
-        if (info.completeBaseName().contains(backupString)) {
-            backupCount++;
-        }
-    }
+    QFileInfo fileInfo(fileName);
+    QString baseName = fileInfo.completeBaseName();
 
-    QString sBackupFile = info.baseName() + "." + backupString + QString::number(backupCount) + "." + info.suffix();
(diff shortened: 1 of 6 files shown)


core_lib/src/structure/filemanager.cpp
```diff
@@ -547,10 +569,19 @@ QString FileManager::backupPreviousFile(const QString& fileName)
         return "";
 
     QFileInfo info(fileName);
-    QString sBackupFile = info.completeBaseName() + ".backup." + info.suffix();
+    int backupCount = 1;
+    QDir directory(info.absoluteDir());
+    QString backupString = "backup";
+    for (QFileInfo info : directory.entryInfoList(QDir::Filter::Files)) {
+        if (info.completeBaseName().contains(backupString)) {
```
chchwy, line 576: The word "backup" is commonly used in terms of file naming. The chances are that users could put "backup" in their filename and then the `backupCount` number would be wrong.

I would suggest to check with full filenames such as `sBackupFile` below to get a correct backup count.

core_lib/src/structure/filemanager.cpp
```diff
@@ -547,10 +569,19 @@ QString FileManager::backupPreviousFile(const QString& fileName)
...
+            backupCount++;
+        }
+    }
+
+    QString sBackupFile = info.baseName() + "." + backupString + QString::number(backupCount) + "." + info.suffix();
```
chchwy, line 581: `completeBaseName()` instead of `baseName()`? Just in case there are dots in filenames.

chchwy:
Thanks @MrStevns I think generally it's good. It can be merged after some minor improvements as comments.

core_lib/src/structure/filemanager.cpp
```diff
@@ -255,8 +271,8 @@ Status FileManager::save(const Object* object, const QString& sFileName)
     QString sMainXMLFile;
     QString sDataFolder;
 
-    const bool isOldType = sFileName.endsWith(PFF_OLD_EXTENSION);
-    if (isOldType)
+    Status isArchiveSt = isArchiveFormat(sFileName);
+    if (isArchiveSt == Status::NOT_ARCHIVE_FORMAT)
```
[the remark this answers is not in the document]
MrStevns: You're right.. previously I was using the sanity check function, and used the Status details to enrich the error report but since we don't do that anymore, there's no need to use Status, I will change that

core_lib/src/structure/filemanager.cpp
MrStevns: Ah.. good point, I forgot about that! will fix.

core_lib/src/structure/filemanager.cpp
```diff
@@ -64,7 +69,15 @@ Object* FileManager::load(const QString& sFileName)
...
-        unzip(sFileName, obj->workingDir());
+        Status sanityCheck = MiniZ::sanityCheck(sFileName);
+
+        // Let's check if we can read the file before we try to unzip.
+        if (!sanityCheck.ok()) {
+            dd.collect(sanityCheck.details());
+        } else {
+            Status unzipStatus = unzip(sFileName, obj->workingDir());
+            dd.collect(unzipStatus.details());
+        }
```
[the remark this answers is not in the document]
MrStevns: We don't do much beyond that point, that would cause further problems I think, so I left it there because we would get more data for the report. If it fails to unzip, then it should be caught by one of the following checks.

core_lib/src/structure/filemanager.cpp
MrStevns: Fixed

core_lib/src/structure/filemanager.cpp
MrStevns: Fixed

core_lib/src/structure/filemanager.cpp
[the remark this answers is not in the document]
MrStevns: fixed

PR #1712, opened 2022-05-14
Fix inconsistent usage of apply transform dialog

MrStevns:
Implement UI block logic for when an action is required

While the main problem I sought to fix was that while using the move tool, you couldn't drag a keyframe on the timeline, the underlying issue was caused by the layerSwitching logic which wasn't fully thought through.

The new logic depends on the tool or widget to tell the editor that a user action is required before the execution can continue, this should be done using the new "requireUserAction" and "RequestUserAction" function. The "RequestUserAction" method will check if the flag has been enabled and in that case, call the respective tool that requires it. It's then up to the tool to create a dialog for the said action.

To make the dialog appear consistently when interacting with widgets that would apply the transformation, we apply an invisible widget on top of the base dock widget which will appear when the UI should be blocked. This makes it possible for the respective QWidget to get the QEvent instead of its children.

Fixes:
- Fix move tool blocking timeline keyframe dragging action
- Fix many other cases where the transform dialog should have appeared but didn't.
- Fixed that when the action popup disappears, the the timeline press event would be called but not the release event, leaving the widget in a pressed state where the scrubber would be moved upon hover.

### DIFF
--- a/core_lib/src/interface/basedockwidget.cpp
+++ b/core_lib/src/interface/basedockwidget.cpp
@@ -49,9 +49,7 @@ void BaseDockWidget::resizeEvent(QResizeEvent *event)
 {
     QDockWidget::resizeEvent(event);
 
-    if (mBlockUIWidget) {
-        mBlockUIWidget->setGeometry(QRect(QPoint(), event->size()));
-    }
+    mBlockUIWidget->resize(event->size());
 
     // Not sure where the -2 comes from, but the event width is always 2 more than what is passed to FlowLayout::setGeometry
     int minHeight = getMinHeightForWidth(event->size().width() - 2);
--- a/core_lib/src/interface/basedockwidget.cpp
+++ b/core_lib/src/interface/basedockwidget.cpp
@@ -35,6 +35,10 @@ BaseDockWidget::BaseDockWidget(QWidget* pParent)
     }
 #endif
 
+    mBlockUIWidget = new QWidget(this);
+
+    mBlockUIWidget->setHidden(true);
+    layout()->addWidget(mBlockUIWidget);
 }
 
 BaseDockWidget::~BaseDockWidget()
@@ -44,6 +48,10 @@ BaseDockWidget::~BaseDockWidget()
 void BaseDockWidget::resizeEvent(QResizeEvent *event)
 {
     QDockWidget::resizeEvent(event);
+
+    if (mBlockUIWidget) {
+        mBlockUIWidget->setGeometry(QRect(QPoint(), event->size()));
+    }
 
     // Not sure where the -2 comes from, but the event width is always 2 more than what is passed to FlowLayout::setGeometry
     int minHeight = getMinHeightForWidth(event->size().width() - 2);
@@ -62,4 +70,15 @@ int BaseDockWidget::getMinHeightForWidth(int width)
 {
     Q_UNUSED(width)
     return -1;
+}
+
+void BaseDockWidget::blockWidget(bool block)
+{
+    if (block) {
+        mBlockUIWidget->raise();
+        mBlockUIWidget->setHidden(false);
+    } else {
+        mBlockUIWidget->setHidden(true);
+        mBlockUIWidget->lower();
+    }
 }
--- a/core_lib/src/tool/movetool.cpp
+++ b/core_lib/src/tool/movetool.cpp
@@ -168,9 +168,8 @@ void MoveTool::pointerReleaseEvent(PointerEvent*)
     mScribbleArea->updateToolCursor();
     mScribbleArea->updateCurrentFrame();
 
-    if (selectMan->transformHasBeenModified()) {
-        mEditor->requireUserAction(true);
-    }
+    mEditor->requireUserAction(selectMan->transformHasBeenModified());
+
 }
 
 void MoveTool::updateTransformation()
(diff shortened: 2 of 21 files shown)


core_lib/src/tool/movetool.cpp
```diff
@@ -363,15 +367,11 @@ bool MoveTool::leavingThisTool()
...
-bool MoveTool::switchingLayer()
+void MoveTool::requestAction()
 {
-    auto selectMan = mEditor->select();
-    if (!selectMan->transformHasBeenModified())
-    {
-        mEditor->deselectAll();
-        return true;
-    }
+    if (mWarningShown) { return; }
```
J5lx, line 372: Isn’t this always reset to false at the end of this method anyway?

J5lx (changes requested):
I just took a look at the changes, but there are still some rough edges.

Firstly, there’s a bug when you are using the select tool and switching to the move tool temporarily using the Alt key and the disabled widgets do not get re-enabled after moving the selection. But even while the widgets were disabled, it’s still possible to use keyboard shortcuts to switch tools and probably also for other actions – maybe relying purely on the UI to prevent those actions is a bit too brittle?

Secondly, I found it a bit confusing to disable all those widgets. To me, disabled normally means that I can’t interact with those widgets, but in this case I can – causing the confirmation dialog to show up (which I noticed isn’t modal btw, not on master either). I reckon keeping the widgets interactive/enabled and simply showing the dialog on interaction would be sufficient.

Lastly, I think requireUserAction and requestUserAction could use some clarification. It’s easy to forget which one is which and not documented what kinds of interaction they are meant to prevent.

Other than that, the code does look good though. It’s definitely good to have this case handled more comprehensively than it is now.

Edit: I should probably also mention that I can still interact with the main menu even though it should be disabled. But that seems to be an issue with my Global Menu widget, not with your code.

core_lib/src/interface/basedockwidget.cpp
```diff
@@ -45,6 +49,10 @@ void BaseDockWidget::resizeEvent(QResizeEvent *event)
 {
     QDockWidget::resizeEvent(event);
 
+    if (mBlockUIWidget) {
```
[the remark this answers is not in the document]
MrStevns: Hmm good point, redundant code gone now

core_lib/src/tool/movetool.cpp
MrStevns: Yes but UI dialogs block the execution loop so it's not reset until one has interacted with the dialog. The logic is from an older iteration of the PR though, it's redundant now and thus I've removed it.

MrStevns:
Thanks for reviewing Jakob!

I've removed the disable logic now, I previously had it for some event handling but it's redundant now, also I agree that it might not make much sense to disable the widget when you can't interact with them anyway because of the overlay. 
As requested i've clarified the usage methods now by merging them into `requestUserActionIfNeeded` which handles both requestUserAction and requireUserAction. I realized that it did not make much sense to first check and then do, when the latter method didn't do anything anyway.

Ready for second round 😄

J5lx:
I’m simply pressing the . and , keys which I believe are the default shortcuts for that. I’m not really doing anything special, just pressing those keys like normal.

> It was rather messy

Hmm I see… :thinking:

J5lx:
I also just tried disabling my global menu to see if the shortcut thing is related to the bug with the disabled toplevel menus that I mentioned in my initial review (which doesn’t seem to be the case) and I realised that it might technically also make sense to keep some entries enabled, e.g. File→Quit or Edit→Clear frame. Though of course that would also be additional effort… maybe it would make sense to build something more general to let the Editor reliably communicate available actions to the UI… though that might just be overkill for this PR…

MrStevns:
The issues should have been addressed now. Triggering a shortcut will now trigger the dialog. I'm also now only disabling the import menu as opposed to the entire file menu.

I'm still not sure about the let Editor tell UI what is possible, it technically makes sense but the amount of work required to pull it off is not trivial as far as I can see. I also thought again about giving the responsibility to the editor but I always run into the same problem, if you need to be able to cancel an action then you either need to intercept all calls or have checks in various parts of the core which a return value that also has to be acted upon. 

The PR focuses on fixing the inconsistent behavior of the dialog (which causes bugs), which I think it achieves, when that's said, there might be better ways or more resilient methods that surely can be implemented when the need arises.

The clear frame is still disabled but that's a deliberate choice given that you've started a transformation, this I think it makes no sense wanting to clear your entire keyframe.

J5lx (changes requested):
After trying it out, I like the solution with the shortcut filter more than I would have expected. Nice work! I like the menu change too. That said, it’s still not quite enough unfortunately since the shortcut filter is only installed on the tool switching actions, whereas other keyboard shortcuts, like the ones for navigating between frames which I mentioned before, can still be used. So I guess one would have to install the filter on all of the other relevant actions as well…? There really is no silver bullet for this issue, huh. *sighs*

> you need to be able to cancel an action then you either need to intercept all calls or have checks in various parts of the core

Yeah that would be the tricky/cumbersome part I suppose… I just wish there was a simple and straightforward solution, lol. Nevertheless, thanks for looking into it.

> deliberate choice given that you've started a transformation

Fair point.

MrStevns:
I'm looking into removing the apply dialog in another PR where i've reworked the select and move tool to handle transformations properly, so i'll close this pr.

PR #1715, opened 2022-06-04
Fix perspective line length not taking zoom into account

### LINKED ISSUE #1714: Overlays do not extend to the edges of the canvas
### Issue Summary
Currently the overlays uses a fixed length factor which means that they become unusable as guides when zooming too far out.

 

### Actual Results
The lines are drawn with a fixed length factor

 

### Expected Results
The lines always extend to the region of the canvas, no matter the zoom level.

 

### Steps to reproduce
1. Enable any perspective tool
2. Zoom out

 

### System Information
+ Pencil2D Version:
master: 855c42307213fc67f7ea805eb30432f5c42b1276

MrStevns:
This addresses #1714 by making sure the line length factor is taking the zoom level into account.

Closes #1714

### DIFF
--- a/core_lib/src/overlaypainter.cpp
+++ b/core_lib/src/overlaypainter.cpp
@@ -6,7 +6,6 @@
 
 
 Q_CONSTEXPR static int OVERLAY_SAFE_CENTER_CROSS_SIZE = 25;
-Q_CONSTEXPR static qreal LINELENGTHFACTOR = 2.0;
 Q_CONSTEXPR static int LR_DIFF = 10;         // minimum difference for Left and Right point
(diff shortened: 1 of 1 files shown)


J5lx:
If I’m reading the issue correctly, I don’t think this PR resolves it. This change does cause the length of the lines to stay the same regardless of zoom level, but it’s still relatively easy to end up in a state where they do not entirely cover the canvas:



Or did I misunderstand what the report is about?

MrStevns:
Ah silly me.. no good point, this is not as trivial as I thought it would be. Changing to draft...

PR #1716, opened 2022-06-05
Ignore pixel-based scrolling deltas on X11

J5lx:
[Build](https://drive.google.com/file/d/1T6_DKj6VQC919bAz4Mds9Qha8dtJIp6a/view?usp=sharing)


MrStevns (approved):
Fix looks good to me. I'm thinking right now as it's just x11 it's ok to keep it here but if we need to add more logic surrounding system window behavior, we might have to put it somewhere else.

PR #1719, opened 2022-06-28
Store nightly builds as artifacts

J5lx:
This is the thing I proposed on Discord earlier this month, so I’ll repost what I wrote there:

> Hi, I’ve been thinking about our nightly build uploads recently because the Linux upload broke once more and I’m kinda tired of fixing it again and again. It’s essentially because of the old version of Python on the build system, so whenever some dependency stops supporting that old Python version I get to figure out which one it is and what’s the last version that still works. Of course one could try to find a newer Python version, but I’ve been wondering if it might be a good idea to consider GitHub Actions artefacts.
>
> Pros:
> - Fewer build dependencies to worry about
> - No hard storage limit on public repositories
> - All forks automatically get their own builds uploaded, even without setting up Google Drive (which is ridiculously complicated)
> - Pull requests also get their builds uploaded
> - Downloads links right next to the build logs
>
> Cons:
> - Only 90 days retention for public repositories (instead of current 120 days)
> - No anonymous downloads (though it’s quite easy to work around that using the API)
> - GitHub insists on zipping artefacts for download even if they’re only a single file or already zipped

For the API workaround that I mentioned I created a little service at get.pencil2d.org which uses the GitHub API to generate a download URL (which is valid for one minute) and returns a redirect to that URL. A link to that service is automatically generated in the summary for each job, so the builds can be easily downloaded even without having to sign in. I considered automatically adding comments on PRs as well (as suggested in #1152), but unfortunately that’s a bit more tricky due to API permissions and the fact that you need to aggregate the job results first, so I’ve left that for later. I did however adjust the nightly build download page on our website as necessary, for which I’ve created a companion PR at pencil2d/pencil2d.github.io#25.


MrStevns:
How many changes are required in order to make it work for our forks? if possible we'll need to update https://github.com/pencil2d/pencil/wiki/CI-build-and-deploy too

When that's said, I welcome these changes, a simpler CI pipeline is always nice

J5lx:
It should work out of the box, with no setup needed. That was actually one of my motivations for this :)

MrStevns (approved):
Changes look good to me, well done 👍

PR #1721, opened 2022-07-02
Assorted fixes

J5lx:
This PR is a collection of various unrelated fixes and enhancements. I don’t think any of them have been (formally) reported. Here’s a list of the issues:

- An issue where the project recovery dialog was not modal, which caused it to appear behind the main window on my system
- Some runtime warnings about missing object names for the toolbars. This didn’t seem to cause any actual issues, but I figured I should fix it anyway
- An issue that caused the undo/redo shortcuts not to work under certain circumstances, for example right after program startup
- An issue where the translation context was set incorrectly for some strings due to inappropriate use of QObject::tr()
- An issue where Pencil2D ignored even the --help and --version options in release builds when another instance was already running
- A memory leak related to object data

I also took another look at the setlocale() workaround for #940 because I always felt like there must be a better solution. And I was right! Not only that, but the Qt devs already implemented it some time ago in 5.14 – because as it turns out, this was [a bug in Qt](https://bugreports.qt.io/browse/QTBUG-80068) all along. I’ve verified that the bug is indeed gone in current Qt versions and updated the code accordingly. I also updated it to use the C locale instead of US English (because the latter might not always be available) and moved the code so that it is run after QCoreApplication has been initialised, [as recommended in the documentation](https://doc.qt.io/qt-5/qcoreapplication.html#locale-settings).

### DIFF
--- a/app/src/main.cpp
+++ b/app/src/main.cpp
@@ -15,8 +15,6 @@ GNU General Public License for more details.
 
 */
 
-#include <clocale>
-
 #include "log.h"
 #include "pencil2d.h"
 #include "pencilerror.h"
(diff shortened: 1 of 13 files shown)


scribblemaniac (approved):
Looks good to me.

MrStevns (approved):
LGTM too, I will merge the changes now

PR #1729, opened 2022-10-14
Fix recently reported issues

### LINKED ISSUE #1726: Bitmap Cut / Copy & Paste "on top" functionality broken
### Issue Summary

I've been piling up issues for a long time found on the nightlies and it's time to unleash them.

The ability to copy a selection and paste it on the same place is paramount to maintain a quick workflow when creating "moving holds" and retracing parts of an animated character. 

Pencil2D 0.6.6 has the feature working, but after we got into the dev nightlies, after some of the changes made to the canvas painter & clipboard it seems this ability was lost, now anytime you paste a copied a selection in Pencil2D, whenever you're pasting the clipboard content on the same frame or a different frame, it will be **pasted on the center of the current view instead**.

I thought it was pasted only to the absolute center of the canvas, but it seems it's pasting it to the center of the canvas window area instead.

Currently only bitmap layers have been affected by this, **vector layers have retained this capacity** so I'm asking for a regression of sorts.

While most apps have the ability to paste both "on top" and "off place" by default I'd say 90% of the time for animation purposes one only needs to be able to paste drawn content in the same place, and since Pencil2D does not have floating selections it's better to avoid damaging other parts of the drawing by moving the selection over too much.

### Expected Results

Copying a partial selection should retain the absolute position on the canvas while pasting on any layer or frame of the same type.

### Steps to reproduce

1. On a bitmap layer purposely draw anything near a corner of the camera view just to better see the issue.
2. Use the select tool to frame a section of your drawing
3. Either with the **select tool** or the **move tool** press CTRL / CMD + C to send to clipboard
4. You can stay on the same frame or create a new blank frame
5. Paste the selection
6. Notice that your selection was pasted dead center on the canvas.
7. Move the view so the canvas and the current drawing are translated
8. Paste the clipboard selection again notice the drawing is pasted on a different place relative to the canvas, but it's still on the center of the current canvas view area

### System Information
+ Pencil2D Version:

 

+ Operating System:

**Nightly Build Mar  8 2022**

commit: 34940a0c6b4257abfbfe983268f75ded6c1d8e53
date: 2022-03-08_14:18:06

Operating System: Windows 10 Version 2009
CPU Architecture: x86_64
Qt Version: 5.15.2

**Nightly Build Aug 18 2022**

commit: 8a2feec15536175c4f656743d794aa8c609953e3
date: 2022-08-18_17:26:43

Operating System: Windows 10 Version 2009
CPU Architecture: x86_64
Qt Version: 5.15.2

### LINKED ISSUE #1727: Camera layer crash when dragging timeline cursor after moving frame
### Issue Summary

This issue was found by David Lamhauge back in January 14th 2022 and it's still happening with the latest version of the dev builds however I hadn't been able to submit an issue until now to keep track of it, apologies.

https://discord.com/channels/342369662710972417/395346048727515137/963399508224507904

Basically there's a easy to reproduce crash that occurs when selecting the camera, moving the first camera frame and then clicking & drag on the playhead / timeline cursor.

### Expected Results
Clicking on the playhead / timeline cursor after moving _any_ frame container should not crash the application.

### Steps to reproduce

As per David's instructions

```
1) Open Pencil2D with a standard file
2) Select the camera layer
3) Select frame 1 and move it to frame 10
4) press on the scrubber on a frame less than 10
5) P2D crashes
```

### System Information
+ Pencil2D Version:

**Originally tested with** 

```
Nightly Build Jan 14 2022

commit: d9078b1a47f611e5389b1c36dce8825b41f9bd49
date: 2022-01-14_16:11:07

Operating System: Windows 10 Version 2009 (it's actually 21H2)
CPU Architecture: x86_64
Qt Version: 5.15.2
```

**Today tested with**

```
Nightly Build Aug 18 2022

commit: 8a2feec15536175c4f656743d794aa8c609953e3
date: 2022-08-18_17:26:43

Operating System: Windows 10 Version 2009
CPU Architecture: x86_64
Qt Version: 5.15.2
```

### LINKED ISSUE #1728: Importing video & replacing colors w/ alpha using fill tool replace mode fails to be saved
### Issue Summary

Basically the title explains the issue. Replacing a color with another that has transparency set to 0% will ignore transparency and change the color back to a premultiplied state (replacing it with fully opaque black) when the file is saved.

### Expected Results

It is expected that the images modified inside the editor will retain their properties regardless of their origin. So replacing a color with another transparent color should be preserved.

This issue is most definitely related to https://github.com/pencil2d/pencil/issues/1641 which was previously fixed. This means that we should check all the import formats for this bug to be thoroughly fixed. I'll report later with findings on the comments if needed.

### Video or Image Reference

image

After saving the raw image, and without closing the file on the `%TEMP% \ Pencil2D \alpha_test_Y2xD_woaionvu\data\` system folder is already seen like this

image

If I close and reopen the file, sure enough I get a black fill instead of the color with 0% alpha I used to replace the BG (_black is often used to represent alpha when using premultiplied alpha matting_)
image

_images courtesy of Dragonixa user on discord_
### Steps to reproduce

1. File > Import > Movie Video
2. Use fill tool with **replace** blending mode
3. Select any color and lower alpha to 0%
4. Click on any color you wish to replace
5. Save and reopen the file.

### System Information
+ Pencil2D Version:

 

+ Operating System:

```
Nightly Build Aug 18 2022

commit: 8a2feec15536175c4f656743d794aa8c609953e3
date: 2022-08-18_17:26:43

Operating System: Windows 10 Version 2009
CPU Architecture: x86_64
Qt Version: 5.15.2
```

J5lx:
This PR fixes the three issues Jose reported recently:
- Fixes #1726 by reading from the system clipboard only when its contents were written by other applications
- Fixes #1727 by simply removing two noop lines
- Fixes #1728 by ensuring that BitmapImage always uses the premultiplied ARGB32 format. There was already an “on-demand” conversion that scribble added a while back but I figured it’d be a good idea to prevent BitmapImage from ever using unintended formats to begin with. That should also take care of any other potential issues in that area, like Jose suggested in the report.

### DIFF
--- a/core_lib/src/managers/clipboardmanager.cpp
+++ b/core_lib/src/managers/clipboardmanager.cpp
@@ -34,10 +34,14 @@ void ClipboardManager::setFromSystemClipboard(const QPointF& pos, const Layer* l
 {
     // We intentially do not call resetStates here because we can only store image changes to the clipboard
     // otherwise we break pasting for vector.
+    // Only bitmap is supported currently...
+    // Only update clipboard data if it was stored by other applications
+    if (layer->type() != Layer::BITMAP || mClipboard->ownsClipboard()) {
+        return;
+    }
 
     QImage image = mClipboard->image(QClipboard::Clipboard);
-    // Only bitmap is supported currently...
-    if (layer->type() == Layer::BITMAP && !image.isNull()) {
+    if (!image.isNull()) {
         mBitmapImage = BitmapImage(pos.toPoint()-QPoint(image.size().width()/2, image.size().height()/2), image);
     }
 }
--- a/core_lib/src/graphics/bitmap/bitmapimage.cpp
+++ b/core_lib/src/graphics/bitmap/bitmapimage.cpp
@@ -51,7 +51,7 @@ BitmapImage::BitmapImage(const QPoint& topLeft, const QImage& image)
 {
     mBounds = QRect(topLeft, image.size());
     mMinBound = true;
-    mImage = image;
+    mImage = image.convertToFormat(QImage::Format_ARGB32_Premultiplied);
 }
 
 BitmapImage::BitmapImage(const QPoint& topLeft, const QString& path)
@@ -70,7 +70,7 @@ BitmapImage::~BitmapImage()
 
 void BitmapImage::setImage(QImage* img)
 {
-    Q_ASSERT(img);
+    Q_ASSERT(img && img->format() == QImage::Format_ARGB32_Premultiplied);
     mImage = *img;
     mMinBound = false;
 
@@ -125,7 +125,7 @@ void BitmapImage::loadFile()
 {
     if (!fileName().isEmpty() && !isLoaded())
     {
-        mImage = QImage(fileName());
+        mImage = QImage(fileName()).convertToFormat(QImage::Format_ARGB32_Premultiplied);
         mBounds.setSize(mImage.size());
         mMinBound = false;
     }
@@ -181,13 +181,6 @@ BitmapImage BitmapImage::copy(QRect rectangle)
     if (rectangle.isEmpty() || mBounds.isEmpty()) return BitmapImage();
 
     QRect intersection2 = rectangle.translated(-mBounds.topLeft());
-
-    // If the region goes out of bounds, make sure the image is formatted in ARGB
-    // so that the area beyond the image bounds is transparent.
-    if (!mBounds.contains(rectangle) && !image()->hasAlphaChannel())
-    {
-        mImage = mImage.convertToFormat(QImage::Format_ARGB32);
-    }
 
     BitmapImage result(rectangle.topLeft(), image()->copy(intersection2));
     return result;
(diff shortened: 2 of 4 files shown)


core_lib/src/managers/clipboardmanager.cpp
```diff
@@ -34,11 +34,15 @@ void ClipboardManager::setFromSystemClipboard(const QPointF& pos, const Layer* l
...
+    if (layer->type() != Layer::BITMAP || mClipboard->ownsClipboard()) {
+        return;
+    }
 
     QImage image = mClipboard->image(QClipboard::Clipboard);
-    // Only bitmap is supported currently...
-    if (layer->type() == Layer::BITMAP && !image.isNull()) {
-        mBitmapImage = BitmapImage(pos.toPoint()-QPoint(image.size().width()/2, image.size().height()/2), image);
+    if (!image.isNull()) {
+        mBitmapImage = BitmapImage(pos.toPoint()-QPoint(image.size().width()/2, image.size().height()/2), image.convertToFormat(QImage::Format_ARGB32_Premultiplied));
```
MrStevns, line 45: Same here, i'd prefer if we could avoid having convertToFormat around in the codebase, instead of inside BitmapImage

MrStevns (changes requested):
Code looks good to me although as mentioned, i'd prefer if we could keep such format conversions to be inside bitmapimage and not littered around the codebase.

MrStevns (approved):
NIce work, merging!

PR #1730, opened 2022-10-16
Switch away from last layer before deleting it

J5lx:
This fixes a bug where, when deleting the last layer, the layer deletion was performed before actually switching away from the layer so that, for a brief moment, the current layer didn’t actually exist. This issue must have been around for a while, but it was recently exposed through the addition of the onCurrentLayerWillChange signal in #1724. Unfortunately I missed that while reviewing the PR, so on current master this bug means that deleting the last layer while it is active will actually result in a bunch of assertions to fire before the program ultimately crashes.


MrStevns (approved):
Changes looks good to me, you're right.. this must has been a bug for a while but just recently exposed.

PR #1731, opened 2022-10-20
Fix crash when deleting sound layer after scrubbing

J5lx:
I had another look at the sound layer deletion crash and I think I finally did figure how to fix it – or work around it, rather. Looks like it’s a bug in the DirectShow backend of Qt Multimedia that results in a segfault when it is destroyed while paused. Neither the GStreamer backend nor the Windows Media Foundation backend exhibit this behaviour. Fortunately, it’s pretty easy to work around.


MrStevns (approved):
Looks  good to me 👍

PR #1733, opened 2022-10-24
Fix truncated playback for converted audio clips on Windows

J5lx:
I took a look at the truncated audio issue for certain file formats and figured out how to fix it – or work around it, rather. Once again, it’s an issue that is specific to the DirectShow multimedia backend on Windows. This backend seems to have trouble properly playing WAV files containing certain metadata that FFmpeg includes. I actually looked at the files in a hex editor and the format of the metadata looks completely fine, so FFmpeg doesn’t do anything wrong here AFAICT. I also tried both the GStreamer backend and the Windows Media Foundation backend and both of those worked fine. It’s really just the DirectShow backend that doesn’t work properly. As a workaround I simply added some FFmpeg options to strip all metadata. Because users could conceivably import problematic WAV files directly, I also made it so that WAV files too are processed through FFmpeg to strip any and all metadata.

### DIFF
--- a/app/src/actioncommands.cpp
+++ b/app/src/actioncommands.cpp
@@ -177,12 +177,10 @@ Status ActionCommands::importSound(FileType type)
     {
         st = Status::CANCELED;
     }
-    else if (strSoundFile.endsWith(".wav"))
-    {
-        st = mEditor->sound()->loadSound(key, strSoundFile);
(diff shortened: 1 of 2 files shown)


MrStevns (approved):
Looks good to me 👍

PR #1734, opened 2022-10-25
Move display options to toolbars

J5lx:
As suggested on Discord last Sunday. While I was at it I also moved the clear frame button from the tools window to the main toolbar, since it’s not actually a tool.


MrStevns:
We might want to resize some of the icons, so they're aligned with the rest, as currently some of them are huge in comparison.
image

MrStevns (approved):
PR looks good to me 👍 

Maybe we should start looking into putting the toolbar related code into its own class though? For future PR's of course.

PR #1736, opened 2022-10-30
Ensure export file path always contains a valid file name extension

J5lx:
On Windows, it is possible to select an export file path with an invalid file name extension by manually typing the name of a pre-existing file into the text field. That will then confuse the movie exporter because it doesn’t know how to handle the unknown extension, as seen in [this forum thread](https://discuss.pencil2d.org/t/cant-export-animation-as-video/6755?u=j5lx). To avoid this, it is necessary to manually check the extension of the selected file path.


MrStevns (approved):
Code looks good to me, however this will PR clash with https://github.com/pencil2d/pencil/pull/1657

So i'm not sure whether we should wait with merging till the other one is done, what do you think?

PR #1737, opened 2022-10-31
Defer redraw of timeline content

J5lx:
Based on my findings from when I profiled the performance issue in #1724, I changed the timeline to redraw its contents only in response to a paint event rather than every single time in updateContent() (which is triggered by all sorts of operations throughout the program) and also made the existing caching mechanism a bit more aggressive while I was at it.

Additional changes:
- Fixed an issue where the current frame border was not updated while playing
- Fixed an issue where the current frame border did not take the length of sound frames into account
- Fixed an issue that allowed frames to become selected after dragging the timeline via MMB
- Moved the timeline and related classes from core_lib to app because that’s where all the other dock widgets live
- Delayed showing the dock widgets until they have been docked to prevent them from briefly appearing in an undocked state before the main window is shown


MrStevns:
I will start reviewing this now

app/src/timelinecells.cpp
```diff
@@ -784,16 +783,17 @@ void TimeLineCells::paintEvent(QPaintEvent*)
...
-                             QString::number(mEditor->currentFrame()));
+                             QString::number(currentFrame));
         }
     }
 }
 
 void TimeLineCells::resizeEvent(QResizeEvent* event)
 {
-    clearCache();
+    delete mCache;
```
MrStevns, line 795: imo. I think it's better to keep the clearCache method, having to call delete and set it to nullptr manually seems error prone.

MrStevns (changes requested):
Overall the changes looks good, although with additional cache logic now, we're definitely getting to the point where the next time we're touching the timelinecells class, we should refactor it beforehand.. it's getting to the point where it's difficult to understand changes and the control flow.

I noticed a bug which was introduced with the new cache logic
- When right-clicking on the camera keyframe, the menu appears but the frame is not filled with white, which indicates that the frame is highlighted. The fix seems to be to add mRedrawContent to true in the showCameraMenu, so the repaint happens.

The rest of the changes looks good, and nice to see some well deserved clean up too 👍 
I'd like to see the bug fixed but otherwise I only have one comment regarding the clearCache method being removed.

MrStevns:
Changes looks good to me now, merging.

PR #1738, opened 2022-11-05
Camera: calculate control points on the fly

MrStevns:
For this PR the goal has been to create stronger coupling between the camera tool and the painter that handled the interactive parts. This has been done by introducing a BaseTool::paint(...) method which allows us to perform calculations that can both be used by the tool and its painter without addition branching to Scribbearea. One example where we benefit from this when calculating the centered control point without explicitly setting it. Another example calculating the distance between the rotation anchor, its line and the camera frame, and this has to be done two places, locally in the tool and in the world coordinate space.

### Fixes
- Rotation anchor would not keep a fixed distance from the camera frame
How to reproduce: 
1. Load a project, select the camera tool and the camera layer
2. Zoom in and out, notice the distance of the rotation anchor being slightly off
3. Now increase the size of the camera frame and repeat the steps
Result: The rotation anchor has been positioned further away from the frame
- Sometimes dragging the last control handle would drag the previous instead...

- Reworked recovery logic for splitting and merging control points

### Changes
- Camera tool painter logic has been moved to camera tool
- PathCPM has been removed as its not necessary anymore

### DIFF
--- a/core_lib/core_lib.pro
+++ b/core_lib/core_lib.pro
@@ -107,6 +107,7 @@ HEADERS +=  \
     src/util/pencildef.h \
     src/util/pencilerror.h \
     src/util/pencilsettings.h \
+    src/util/transform.h \
     src/util/util.h \
     src/util/log.h \
     src/util/movemode.h \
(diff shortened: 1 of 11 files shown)


MrStevns:
Agh, I forgot to update the tests, will do that tomorrow...

MrStevns:
Tests now updated, PR is ready :)

MrStevns:
Thanks for reviewing, fixing build errors and merging! 💯

Issue #1574, opened 2021-02-18
Consider switching from linuxdeployqt to linuxdeploy


scribblemaniac:
In making the linux legacy release for the latest version of Pencil2D, I was reminded of some changes I made for the build process of the 0.6.5 linux legacy build. Most importantly, I stopped using linuxdeployqt because it was becoming increasingly difficult to get it working with 32-bit architectures. After looking at a couple alternatives I settled on [linuxdeploy](https://github.com/linuxdeploy/linuxdeploy) for that release, and I have used it again for the 0.6.6 release. Based on my experiences with it, I believe that we should consider migrating all of our linux deployments (64-bit releases, and nightlies) to linuxdeploy.

The main reason to migrate is that it it is very easy to call, and handles gstreamer plugins properly without the workarounds we currently have to employ for linuxdeployqt (although I have not tested this with the 64-bit version). The usage for linuxdeploy would look like this (after make install + ffmpeg install + dependency install):
```
linuxdeploy --appdir Pencil2D --plugin qt
appimagetool Pencil2D pencil2d-linux-arch-version.AppImage
```
For context, that would replace all of this in after-build.sh:
```
install -Dm755 "/usr/lib/x86_64-linux-gnu/gstreamer1.0/gstreamer-1.0/gst-plugin-scanner" \
    "Pencil2D/usr/lib/gstreamer1.0/gstreamer-1.0/gst-plugin-scanner"
  local gst_executables="-executable=Pencil2D/usr/lib/gstreamer1.0/gstreamer-1.0/gst-plugin-scanner"
  for plugin in adpcmdec alsa app audioconvert audioparsers audioresample \
      autodetect coreelements gsm id3demux jack mpg123 mulaw playback \
      pulse typefindfunctions wavparse apetag; do
    install -Dm755 "/usr/lib/x86_64-linux-gnu/gstreamer-1.0/libgst${plugin}.so" \
      "Pencil2D/usr/lib/gstreamer-1.0/libgst${plugin}.so"
    gst_executables="${gst_executables} -executable=Pencil2D/usr/lib/gstreamer-1.0/libgst${plugin}.so"
  done
LD_LIBRARY_PATH="${LD_LIBRARY_PATH}:/usr/lib/x86_64-linux-gnu/pulseaudio" \
    ./linuxdeployqt-continuous-x86_64.AppImage \
    Pencil2D/usr/share/applications/org.pencil2d.Pencil2D.desktop \
[...]

PR #1743, opened 2023-02-09
Add support for Qt 6

J5lx:
This PR allows Pencil2D to be allowed to be built against Qt 6 and adds Qt 6 to our CI setup. Overall, not a whole lot of changes are needed. Many of the breaking changes are things that were already deprecated a long time ago (Qt<5.6), so we can simply switch to the replacement API. That said, there were still some places where I needed to rely on conditional compilation. However, a number of these are only needed to support Qt versions < 5.15, so they will no longer be necessary when we start requiring Qt≥5.15.

Some additional notes on the changes:
- `CONFIG+=qt` in qmake seems to be a relic from the Qt 4 days that is simply no longer needed for Qt 5. With Qt 6, it starts causing strange errors.
- The HighQualityAntialiasing render hint has been obsolete since at least Qt 5.5, so we can safely switch to regular Antialiasing
- The CI changes also address some deprecations by [migrating away from the set-output command](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) and updating the jurplel/install-qt-action to its latest version.
- Qt 6 builds for Windows and macOS are uploaded with a -qt6 suffix. There are no 32-bit Windows builds because there are no 32-bit builds of Qt 6 officially available. Qt 6 builds for Linux are not uploaded because (a) I couldn’t be bothered to set up an old-enough build environment for the AppImage to actually be portable and (b) linuxdeployqt also doesn’t support Qt 6 yet (at least not officially). linuxdeploy does support Qt 6, though, so in the future this may be a reason to revisit #1574.

Disclaimer: I did a test run of the main features to make sure they still work, but I didn’t check every little nook and cranny. I also tested only on Linux. I didn’t get around to testing the Windows version, and I don’t have a macOS system to test on.

### DIFF
--- a/docs/build_linux.md
+++ b/docs/build_linux.md
@@ -3,7 +3,7 @@ Building Pencil2D on Linux {#build_linux}
 
 These are instructions for building Pencil2D on Linux. If you are using Windows go [here](@ref build_windows), and macOS go [here](@ref build_macos). This guide is primarily targeted towards developers. If you just want to use the latest version you can just download one of our [nightly builds](https://www.pencil2d.org/download/nightly/).
 
-This tutorial was made with Ubuntu Xenial Xerus (16.04) and Arch Linux in mind, however you should be able to adapt this guide to other versions or distributions if necessary.
+This tutorial was made with Ubuntu and Arch Linux in mind, however you should be able to adapt this guide to other distributions if necessary.
 
 ## Installing Dependencies
 
@@ -26,9 +26,15 @@ Pencil2D relies on the %Qt application framework so you must install it before y
 
 ##### Command-line method
 
-Pencil2D must be built with the %Qt 5 framework. To install %Qt 5, run this command:
+Pencil2D must be built with the %Qt framework, version 5.6 or newer. Therefore, you will need at least Ubuntu 16.10 (Yakkety Yak) to use this method, as Canonical does not provide recent enough versions of Qt for older versions of Ubuntu. To install %Qt 5, run this command:
 
-    sudo apt install qt5-default qt5-qmake libqt5svg5-dev qtmultimedia5-dev
+    sudo apt install qt5-default qtbase5-dev qtmultimedia5-dev qttools5-dev-tools libqt5svg5-dev
+
+@note If you are using Ubuntu 20.04 (Focal Fossa) or newer, you can omit qt5-default from the command above.
+
+If you are using Ubuntu 22.04 (Jammy Jellyfish) or later, you can alternatively install %Qt 6 by running this command:
+
+    sudo apt install qt6-base-dev qt6-l10n-tools qt6-multimedia-dev libqt6svg6-dev
 
 For a more pleasant development experience, you might want to install %Qt Creator as well (recommended). To do so, run the following command:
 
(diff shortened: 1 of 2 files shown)


MrStevns:
I will schedule some time to try it out on my mac in the upcoming weekend, first with my current Qt version 5.15.1 and then I'll upgrade to Qt 6 and see if it still builds.

It would be good if someone else tested the windows environment.

MrStevns:
Tested and works on mac using Qt 5.x and Qt 6.4.2

I did notice a small cosmetic change in the timeline but I assume it's expected behavior from Qt side.
image

On the technical side, when I took the above screenshot and it was put into my clipboard, as soon as I focused on Pencil2D, it crashed.

MrStevns:
I was able to reproduce the bug in a new application, it looks like it's related to qmake. If I recreate the same project using cmake, then it works as expected.

I've created a bug report for now https://bugreports.qt.io/browse/QTBUG-111281
Unless someone else can confirm that the clipboard works on Mac OS, I think that we should either postpone the upgrade to Qt 6 for the Mac platform or put the PR on hold until a fix has been provided.

Thoughts?

MrStevns:
Looking at this again, the issue I created was closed, as further investigation showed that the crash only happened with the homebrew build. There's a new bottle with Qt 6.4.3_1, so i'll try once more today.. if it works fine then i'll merge the PR.

MrStevns:
It's still broken. I've taken the next step and opened a discussion in homebrew's github. Hopefully they can shed some light on the problem... and hopefully reproduce it.

https://github.com/orgs/Homebrew/discussions/4362

MrStevns:
The plot thickens.... a user from the Qt bug report comment section mentioned a similar issue after upgrading to Qt 6. They found a workaround, disabling "Add build library search path to DYLD_LIBRARY_PATH and DYLD_FRAMEWORK_PATH" found in Build & Run -> Run, fixes the crash.
image

Another thread mentioned something interesting... it only crashes through Qt Creator, when launched through Finder, it works as expected.

So in that case, it's probably fine to merge the PR... Mac OS developers will simply have to remember to uncheck that option, then it should work fine.

MrStevns:
In that case, let's merge the PR, then we can always decide whether to distribute a Qt5 or Qt6 build later, if that makes sense.

Jose-Moreno:
> The plot thickens.... a user from the Qt bug report comment section mentioned a similar issue after upgrading to Qt 6. They found a workaround, disabling "Add build library search path to DYLD_LIBRARY_PATH and DYLD_FRAMEWORK_PATH" found in Build & Run -> Run, fixes the crash. image
> 
> Another thread mentioned something interesting... it only crashes through Qt Creator, when launched through Finder, it works as expected.
> 
> So in that case, it's probably fine to merge the PR... Mac OS developers will simply have to remember to uncheck that option, then it should work fine.

@MrStevns @J5lx If you guys don't mind, would it be ok to ask to update the macOS dev guide in the wiki + [developer docs](https://dev.pencil2d.org) with these caveats? Otherwise I feel it will be lost to new contributors, and even ourselves as time goes on. Thank you for the research & reviewing!

MrStevns (commented):
Overall the changes looks good to me, although I'm a bit sad that we have to litter the code with compatibility macros so many places... Aside from that I think that we (the team) need to talk about how far back we want to support because this PR adds a sizable amount of version checking. 

Given the progress of the project and our already limited time, does it make sense to also maintain a larger compatibility layer. I'm asking because I usually don't test on multiple Qt versions... nor am I really interested in doing that but what's the stance on this?

MrStevns:
Fair point, I must admit I had forgotten about that, so thanks for reminding me 😅

> Perhaps it would be easier to compile newer versions of Qt for older operating systems ourselves than it is to keep the codebase compatible with older versions of Qt that others have compiled for those older operating systems.

This resonates with how I feel. I am not against keeping support for older operation systems, if they're compilable using latest Qt LTS. I just don't like littering the code with version macros to make it compilable with older Qt versions.

MrStevns (approved):

MrStevns:
Like previously mentioned, the changes looks good to me. Now it's just also been approved.

PR #1744, opened 2023-02-13
Fix emtpy frame on position 1 when duplicating layer

J5lx:
As reported by Nelphy [on Discord](https://discord.com/channels/342369662710972417/395346048727515137/1074501500103036988). I also thought about renaming loadKey()… fundamentally, it’s more general-purpose than what we’re using it for so far IMO.

In the future, I think we should reconsider how we enforce the “min. one keyframe per layer” requirement. It’s a bit inconsistent right now with different assumptions made in different places, which is fundamentally what led to this issue.

### DIFF
--- a/app/src/actioncommands.cpp
+++ b/app/src/actioncommands.cpp
@@ -704,7 +704,7 @@ void ActionCommands::duplicateLayer()
     Layer* toLayer = layerMgr->createLayer(fromLayer->type(), tr("%1 (copy)", "Default duplicate layer name").arg(fromLayer->name()));
     fromLayer->foreachKeyFrame([&] (KeyFrame* key) {
         key = key->clone();
-        toLayer->loadKey(key);
+        toLayer->addOrReplaceKeyFrame(key->pos(), key);
         if (toLayer->type() == Layer::SOUND)
         {
             mEditor->sound()->processSound(static_cast<SoundClip*>(key));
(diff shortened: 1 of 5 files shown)


app/src/actioncommands.cpp
```diff
@@ -703,10 +702,9 @@ void ActionCommands::duplicateLayer()
     int currFrame = mEditor->currentFrame();
 
     Layer* toLayer = layerMgr->createLayer(fromLayer->type(), tr("%1 (copy)", "Default duplicate layer name").arg(fromLayer->name()));
-    toLayer->removeKeyFrame(1);
     fromLayer->foreachKeyFrame([&] (KeyFrame* key) {
         key = key->clone();
-        toLayer->addKeyFrame(key->pos(), key);
+        toLayer->loadKey(key);
```
MrStevns, line 707: I feel like exposing the loadKey function is wrong, to me it should be an private function. It's only ever been used inside layer, it's a generic function that simply loads a keyframe into the map, however used wrongly, the keyframe could end up not being initialized correctly.

Maybe it would be better to introduce a new function for the purpose of either adding or replacing a keyframe? 

Thoughts?

MrStevns (approved):
LGTM!

MrStevns:
> Are you saying that we should pay attention to this when we eventually add proper undo/redo support or to adjust the current system in some way?

It was just a note that I felt like adding at the time, but yes I suspect that we'll have to make changes to make the undo/redo system work properly when that time come. I didn't mean that we should adjust it to the current system. The less places and more consistent we can make the min 1 keyframe pr layer, the better.

I will merge the PR now, thanks for your work 😄

PR #1745, opened 2023-02-17
Fix crash when switching away from vector layer while not on a keyframe

J5lx:
Stumbled across this bug purely by chance. Seems to be another issue I missed while reviewing #1724. Steps to reproduce:

1. Create a vector layer (or switch to an existing one)
2. Go to any frame that does not have a keyframe on that vector layer
3. Switch to any other layer → Segfault


MrStevns (approved):
Fix confirmed working, good job 👍

PR #1750, opened 2023-03-08
More detailed error reporting for image import failure

### LINKED ISSUE #1674: Better error reporting for image import failure
### Issue Summary

If an image import fails, the same dialog is shown for a variety of unrelated errors.

### Actual Results

Currently if there is any issue with image importing, this dialog is displayed:
Image Import Error Dialog

### Expected Results

When image import fails, an error dialog is displayed similar to the one used for movie export errors detailing what exactly the error was, and providing debug details too if relevant.

### Steps to reproduce

Layer type error:

1. Select a vector layer
2. Attempt to import an image

Image reading error:

1. Download this corrupted image: [pencil2d-corrupted](https://user-images.githubusercontent.com/3461051/140634321-d50764f8-88ba-411a-af61-acdf02362a28.png)
2. Select a bitmap layer
3. Attempt to import the downloaded image

File not found error:

1. Select bitmap layer
2. Attempt to import image but...
3. When the import position dialog comes up, rename the file you selected for importing.
4. Then confirm the position dialog.

Note: There are other errors as well. See https://doc.qt.io/qt-5/qimagereader.html#ImageReaderError-enum

### System Information
+ Pencil2D Version: 0.6.6

+ Operating System: Ubuntu 20.04

+ RAM Size: 16 GB

+ Graphics Tablet: None

<hr>

### Developer Notes

To implement this, `Editor::importImage`, `Editor::importBitmapImage`, and `Editor::importVectorImage` should return `Status` objects with error information, and `MainWindow::importImage` should read those objects and pass the Status data to an `ErrorDialog` if there's an issue.

anpanring:
Fixes #1674

I am new to contributing to Pencil2D and would really appreciate any feedback on these fixes, especially the debugging and error messages. Thank you!

### DIFF
--- a/app/src/mainwindow2.cpp
+++ b/app/src/mainwindow2.cpp
@@ -978,7 +978,7 @@ void MainWindow2::importGIF()
     QString strImgFileLower = gifDialog->getFilePath();
     if (!strImgFileLower.toLower().endsWith(".gif"))
     {
-        ErrorDialog errorDialog(tr("Couldn't import image."), tr("Invalid image type. You can only use import files ending with .gif as Animated GIFs."));
+        ErrorDialog errorDialog(tr("Import failed"), tr("You can only import files ending with .gif."));
         errorDialog.exec();
     }
     else
--- a/app/src/mainwindow2.cpp
+++ b/app/src/mainwindow2.cpp
@@ -868,14 +868,11 @@ void MainWindow2::importImage()
         return;
     }
 
-    bool ok = mEditor->importImage(strFilePath);
-    if (!ok)
+    Status st = mEditor->importImage(strFilePath);
+    if (!st.ok())
     {
-        QMessageBox::warning(this,
-                             tr("Warning"),
-                             tr("Unable to import image.<br><b>TIP:</b> Use Bitmap layer to import bitmaps."),
-                             QMessageBox::Ok,
-                             QMessageBox::Ok);
+        ErrorDialog errorDialog(st.title(), st.description(), st.details().html());
+        errorDialog.exec();
         return;
     }
 
--- a/core_lib/src/interface/editor.cpp
+++ b/core_lib/src/interface/editor.cpp
@@ -939,14 +939,14 @@ Status Editor::importBitmapImage(const QString& filePath, int space)
         switch(reader.error())
         {
         case QImageReader::ImageReaderError::FileNotFoundError:
-            errorDesc = tr("File not found at path \"%1\". Please check if image is present and try again.").arg(filePath);
+            errorDesc = tr("File not found at path \"%1\". Please check the image is present at the specified location and try again.").arg(filePath);
             break;
         case QImageReader::UnsupportedFormatError:
-            errorDesc = tr("This image format is not supported. Please try converting it to one of the following formats and try again:\n%1")
+            errorDesc = tr("Image format is not supported. Please convert the image file to one of the following formats and try again:\n%1")
                         .arg((QString)reader.supportedImageFormats().join(", "));
             break;
         default:
-            errorDesc = tr("We ran into an error while reading this image. Please check if it is valid and try again.");
+            errorDesc = tr("An error has occurred while reading the image. Please check that the file is a valid image and try again.");
         }
 
         status = Status(Status::FAIL, dd, tr("Import failed"), errorDesc);
--- a/core_lib/src/interface/editor.cpp
+++ b/core_lib/src/interface/editor.cpp
@@ -914,17 +914,39 @@ void Editor::updateObject()
     emit updateLayerCount();
 }
 
-bool Editor::importBitmapImage(const QString& filePath, int space)
+Status Editor::importBitmapImage(const QString& filePath, int space)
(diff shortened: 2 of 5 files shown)


scribblemaniac:
Thanks for your submission! We will review it when we get a chance, but I just want to warn you that PRs can move quite slowly in this project. We have very few developers and none of us are all that active right now so it can take some time before someone gets around to reviewing them.

Just skimming the changes you've made, it looks just like what I had in mind for this enhancement. I don't anticipate there being any major issues with this.

scribblemaniac (approved):
Looks good. I've added a couple of further changes you can check out if you would like. Feel free to ask questions about anything I've done.

One of the main changes was adding the use of the `tr` function (https://doc.qt.io/qt-5/qobject.html#tr). We use this function throughout our code to add support for translating strings into other languages. When it comes to error dialogs specifically, we want to translate the title and description as those are meant to be read by the user. We don't translate the debug details as those are meant to be read by the developers and easily searched up in the English code.

The other noteworthy change was adding error dialogs to the image sequence, image predefined set, and animated GIF imports. This wasn't specified in the #1674, but they also should be using the same approach for returning and showing errors. And since they all use the importImage function which now returns a Status object, this was fairly simple to do.

anpanring:
Looks good, and thank you for your help with additional changes! I'll keep the tr function and other import methods in mind for the future.

app/src/mainwindow2.cpp
```diff
@@ -979,29 +976,27 @@ void MainWindow2::importGIF()
     progress.show();
 
     QString strImgFileLower = gifDialog->getFilePath();
-    bool importOK = strImgFileLower.toLower().endsWith(".gif");
-
-    if (importOK)
+    if (!strImgFileLower.toLower().endsWith(".gif"))
+    {
+        ErrorDialog errorDialog(tr("Couldn't import image."), tr("Invalid image type. You can only use import files ending with .gif as Animated GIFs."));
```
MrStevns, line 981:
```suggestion
        ErrorDialog errorDialog(tr("Couldn't import GIF."), tr("Invalid image type. You can only import files ending with .gif as Animated GIFs."));
```

MrStevns (changes requested):
Coding and style wise the changes looks good to me, the messages themselves however could be a bit more aligned.

In general I'd like to see the punctuation removed from the title. We don't use it anywhere else AFAIK and it's generally not common to do so. 
The title should be short and concise, "couldn't" is just extra fluff that doesn't add anything, state the problem, simply put 😉 
I've written a suggestion to one of them, the same title could be used for some of the other errors too.

Thanks for your contribution 😃 
Let me know when you've addressed the mentioned changes, then i'll review again.

core_lib/src/interface/editor.cpp
```diff
@@ -914,17 +914,42 @@ void Editor::updateObject()
...
+        case QImageReader::ImageReaderError::FileNotFoundError:
+            errorDesc = tr("File not found at path \"%1\". Please check if image is present and try again.").arg(filePath);
+            break;
+        case QImageReader::UnsupportedFormatError:
+            errorDesc = tr("This image format is not supported. Please try converting it to one of the following formats and try again:\n%1")
+                        .arg((QString)reader.supportedImageFormats().join(", "));
+            break;
+        default:
+            errorDesc = tr("We ran into an error while reading this image. Please check if it is valid and try again.");
+        }
```
anpanring, line 950: @MrStevns Thank you for your feedback! I've gone through and simplified/removed punctuations from the error titles and descriptions. Do you think I should revise these error descriptions to be a bit more direct?

Jose-Moreno: Hi. I think the descriptions could benefit from having the **error** messages with a bit more concise language, and the **action** the user is supposed to do could be a bit more direct and specific. 

From experience, even if it's contextually obvious the text is referring to the currently imported image, sometimes users can make incorrect choices by not having the description be direct. That is if they even read them instead of pressing thee [OK] button 😅 

That said,  for those who read, I would personally modify these in the following way:

```cpp
case QImageReader::ImageReaderError::FileNotFoundError:
            errorDesc = tr("File not found at \"%1\". Please check the image is present in the specified file location and try again.").arg(filePath);
            break;
        case QImageReader::UnsupportedFormatError:
            errorDesc = tr("Image format is not supported. Please convert the image file to one of the following formats and try again:\n%1")
                        .arg((QString)reader.supportedImageFormats().join(", "));
            break;
        default:
            errorDesc = tr("An error has occurred while reading the image. Please check the file has a valid image format and try again.");
        }
```

What do you guys think?

scribblemaniac: I like Jose's suggested changes, particularly the more active voice in the second error message. I would make a couple slight modifications though:
- "Please check the image is present **at the specified location** and try again."
- "An error has occurred while reading the image. Please check **that** the file **is a valid image** and try again."

anpanring: Thank you both for the help! I've just updated the messages to reflect the suggested changes.

Issue #1085, opened 2018-10-03
Add onion skin global toggle behavior


Jose-Moreno:
### Issue Summary

This is something more of a user experience issue that has been requested here and there. Onion skinning in most programs is normally turned on / off reminiscent of how a light table would. After that you are to handle the inclusion / exclusion of either previous or next frames relative to your current frame.
Consider this I'd like to ask to have a single switch (and shortcut) for toggling the onion skin globally, and leave the current PREV and NEXT functionality to be perceived as filters. This way there would be the following Onion Skin states: All OFF, All ON, PREV ONly, NEXT ONly

So basically we would end up having:

- [ ] A global switch to turn onion skin ON / OFF (w/ shortcut)
- [x] The PREV onion skin switch (w/shortcut)
- [x] The NEXT onion skin switch (w/shortcut)

scribblemaniac:
I think there is potential for confusion if we have a global switch and a prev/next switch. If a user turns on next onion switching, they expect to see onion skins of the next frame. Image a scenario where the global switch is off, and one of the local switches is on but doing nothing. So it can be confusing, especially if they don't realize there is a global switch.

One solution would be to allow both the next and previous onion skin switches to be mapped to the same shortcut. 

An alternative acceptable solution I think would be to have an all on/all off action which simply changes the prev and next switch, but doesn't have any state of its own. The problem with this is, what if there is one on and one off? Should it turn them both on or both off. An action that just toggles both switches is possible too.

CandyFace:
@scribblemaniac To me the natural effect of enabling or disabling the global onion skin would render the previous and next onion skin button gray aka. disabled too.

PR #1753, opened 2023-04-02
Use checkboxes instead of tool buttons in onion skin widget

J5lx:
Just a little interface change which I think will make the onion skin window slightly more intuitive while conserving some space:

image


MrStevns:
Hmm I don't know how I feel about this change. 

I prefer the use of icons over checkboxes when they provide more context, which I think they do in this case. "Previous frames" & "Next frames" doesn't say enough to me about what it does. It does converse a bit more horizontal space though, which is also nice.

Jose-Moreno:
Honestly I'm torn between the two modifications.

The checkbox implementation seems to try to solve the current state of the onion skins where we have the following issues:
1. From user feedback, we've been told it is sometimes unclear how to enable the onion skin feature since the cell layer button to the left does not resemble a button until it is pressed.
2. Additionally it is not obvious at first glance that the feature is enabled / disabled due to the lack of visual cues.
3. When I made the onion skin layout I honestly was wrestling with Qt's UI widgets, but I've always thought that having these sections (Prev. Frame, Next. Frame) stacked like "drawers" might have not been the best choice. If we consider the horizontal nature of the timeline, I sometimes feel it's an additional cognitive obstacle to find out how these options are supposed to work.

### Checkbox Implementation
Thinking positively, the check box proposal made by @J5lx currently seems address the first and second issues pointed above. Elements being subordinated visually & greying out every other elements do help provide a visual hint of the current state of the onion skin at a glance.

### (New) Icon Implementation
image

@MrStevns has been working on improving the icons that were left over in PR https://github.com/pencil2d/pencil/pull/1361 He's showed progress screencaps and at a point we had agreed that the icon that could work best to replace the current ones was a lightbulb since it's been canonically used in other animation software to represents the concept of a light table or light box that lits papers from behind so they become see through.

### Onion Skins UI / UX redesign

Now considering https://github.com/pencil2d/pencil/issues/1085 this request was made by users on other hubs where they asked for a global toggle to enable disable the onion skins instead of having to disable the previous & next options. 

I have frankly needed this too when I'm animating and tried to optimize my workflow for trace backs, although its more of a UX thing, but I thought we could talk about it regarding the UI parts that could help solve the issue.

So with that said I was wondering: What if we had the lightbulb be turned "OFF" with a clear boundary to make it seem like a button (as MrStevns did), as well as having the subordinate items are also greyed out (as J5lx did), this could help showcase that the onion skin status quite well IMO.

If we turn the light bulb "ON" all the other elements would lit up and would help construct the meaning of boolean functionality for the user.

I always thought this general activation button should be the actual lightbulb icon, and in that case we could have the checkboxes for the Prev / Next frames as well. Here's what I'm picturing so we can discuss about it.

image

Left: Onion Skins panel from March 11th | Right: Onion skins panels with both icons and checkboxes to provide visual hierarchy

If the feature was turned off completely, this is how I think it could look

image

Let me know what you think 👋

MrStevns:
As always a very nice and thought-through comment @Jose-Moreno 👏 
I like the draft too! the question is whether we can make it look similar using the standard UI components provided by Qt.

Jose-Moreno:
Sorry for the late reply it's been a hectic week.

> As always a very nice and thought-through comment @Jose-Moreno 👏 I like the draft too! the question is whether we can make it look similar using the standard UI components provided by Qt.

Sure thing!  I had to do a vector mockup since I didn't have much time that day, but whatever Qt allows us to do is fine by me 🙏 

> You also make a good point that the red/blue light bulb icon serve better as an indication of the onion skin exclusively and we should avoid trying to convey multiple things in one icon. You didn't write that explicitly but I could see it in your draft.

Ah, well I  didn't mention it because I wasn't exactly opposed. I was just trying to see how it looked as an alternative and I didn't have time to copy the exact same icon... though I just recalled I could have downloaded them from the PR... 😂 

> These icons however image feels misplaced now, in that they have little to do with min and max opacity but aside from that I like the draft 😄 We could simply remove them, we already have the text explaining what the icons apply to.

I don't have attachment for these icons to be honest, I simply thought It could help to have a visual cue so I put them there because at the time they seemed really fitting to explain the opacity threshold limits, but I won't miss them if we remove them😄

MrStevns:
Thanks for the suggestions, I've tweaked it now and i'm quite satisfied with the result.
image

In the future we should consider making our own custom widgets rather than modifying individual widgets but for now, since we are not using it elsewhere, we'll keep it in the Onion onion widget.

I also had to tweak the spacing of the widget on Mac OS, because for some reason it gets inconsistent between the first and second group box... seems like a another bug, which might have gone unnoticed until now as I can still reproduce it in in Qt dev branch 😩 

Just to prove the point, here's what happens when I set the spacing to 16... 
image

I would have expected the spacing to be seen on each element, however it only happens between the first and second group box.. and then the checkboxes at the bottom. Talk about an infuriating an design bug... 😅 

When that's said, I have pulled the latest changes into the PR and applied my Mac OS tweaks. 
When the redesign PR lands, we could consider making the proposal @Jose-Moreno mentioned.

If you have no further adjustments either, let's merge the PR.

MrStevns (approved):

PR #1754, opened 2023-04-05
Fix move tool axis lock and timeline update issues

J5lx:
This fixes the first two issues reported by JoeyH in [this forum thread](https://discuss.pencil2d.org/t/new-bugs-upgrading-from-nb-nov-8-to-nb-mar-12/7068?u=j5lx). The third one I’m currently not sure about and I’ll have to ask Joey about that first, so I’m leaving that for later.

### DIFF
--- a/core_lib/src/managers/selectionmanager.cpp
+++ b/core_lib/src/managers/selectionmanager.cpp
@@ -150,7 +150,7 @@ void SelectionManager::adjustSelection(const QPointF& currentPoint, const QPoint
     switch (mMoveMode)
     {
     case MoveMode::MIDDLE: {
-        QPointF newOffset = currentPoint - mAlignToAxisStartPosition;
+        QPointF newOffset = currentPoint - mDragOrigin;
 
         if (mLockAxis) {
             mTranslation = offset + alignPositionToAxis(newOffset);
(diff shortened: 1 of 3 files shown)


core_lib/src/managers/selectionmanager.cpp
```diff
@@ -150,12 +150,12 @@ void SelectionManager::adjustSelection(const QPointF& currentPoint, const QPoint
     switch (mMoveMode)
     {
     case MoveMode::MIDDLE: {
-        QPointF newOffset = currentPoint - offset;
+        QPointF newOffset = currentPoint - mAlignToAxisStartPosition;
```
MrStevns, line 153: Although the change itself is fine, I find the variable misleading now. Maybe we should rename to something more fitting?

How about mDragOrigin?

and in header...

/// The point from where the dragging will be based of inside the selection area.
/// Not to be confused with the selection origin
void setDragOrigin(const QPointF point) { mDragOrigin = point; }

MrStevns (changes requested):
Changes looks good to me, i only have a trivial comment about variable naming but otherwise no issues.

Works as expected too for me, good job 👍

MrStevns (approved):
No further comments, changes looks good.
Good job 👍

PR #1756, opened 2023-04-10
Replace -bitexact FFmpeg option with more compatible -(f)flags

J5lx:
The -bitexact option was only added in FFmpeg 4 and is not available in older versions. Luckily, it is no more than a convenient shortcut for the bitexact codec and format flags, which should be old enough.


MrStevns (approved):
Looks good to me.

Maybe it would be a good idea to write a small comment about why we use the specified flag, or refer to the bug it fixes. Just a thought.

PR #1771, opened 2023-07-08
Fix three point perspective overlay anchors

MrStevns:
### How to reproduce:
1. Enable the three point perspective overlay 
2. Select the move tool
3. Hover and click on the left or right anchor

### Expected result
Being able to move the left and right anchor point

### Actual result
Only being able to drag the bottom anchor point


HeCorr:
Current workaround until this gets merged: enable both 2-point and 3-point perspective overlays

image

PR #1773, opened 2023-07-15
Ignore hidden sound layers when exporting video

MrStevns:
Feature or bug... you decide 😄 

https://discuss.pencil2d.org/t/how-to-mute-one-track-when-exporting/7182/4


scribblemaniac:
I have always thought this is the way it should work. Looks good.

PR #1775, opened 2023-07-18
Don't use amerge when exporting a single audio clip

scribblemaniac:
In versions of ffmpeg < 3.2, passing a single audio track to the amerge filter is not allowed. This simply skips the amerge step when there is only one audio clip.

This was causing issues for exporting projects with a single audio clip in the Linux AppImage nightly builds where an older version of ffmpeg is currently being bundled.

### DIFF
--- a/core_lib/src/movieexporter.cpp
+++ b/core_lib/src/movieexporter.cpp
@@ -216,11 +216,22 @@ Status MovieExporter::assembleAudio(const Object* obj,
     panChannelLayout.chop(1);
     // Output arguments
     // Mix audio
-    args << "-filter_complex" << QString("%1%2 amerge=inputs=%3, pan=mono|c0=%4 [out]")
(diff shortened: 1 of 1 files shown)


MrStevns:
LGTM, thanks for fixing this! I assume it is related to this thread? https://discuss.pencil2d.org/t/nb-cant-export-movies-with-sound/7277/7

MrStevns (approved):

scribblemaniac:
> I assume it is related to this thread? https://discuss.pencil2d.org/t/nb-cant-export-movies-with-sound/7277/7

Yes that is correct. Thanks for reviewing. I'm going to have to go through and check what versions of ffmpeg we need to support for legacy builds as there appears to be yet another issue with ffmpeg 2.x and it would be nice to use a newer version if we can.

PR #1778, opened 2023-08-03
Invert canvas rotation direction if view is flipped

### LINKED ISSUE #1777: Counterintuitive canvas rotation when view is flipped
### Issue Summary
When the canvas is flipped either horizontally or vertically (but not both at once), attempting to rotate it (either via the keyboard shortcuts or with Middle Click + Alt), it rotates in the opposite direction from what would be expected which is confusing.
Other software like Paint Tool SAI seem to handle this correctly.

### Actual Results
The view rotates in the wrong direction.

### Expected Results
The view would rotate in the expected/intuitive direction.

### Video or Image Reference
Pencil2D
https://github.com/pencil2d/pencil/assets/75134774/d565bc58-61ee-4c02-b739-3538cb21c0ff

Paint Tool SAI (expected behavior)
https://github.com/pencil2d/pencil/assets/75134774/52234fd0-f6f7-4699-b6ff-2cb6cdbaea69

### Steps to reproduce
1. Open Pencil2D
2. Rotate the canvas Clockwise (R)
3. Flip the view horizontally (Shift + H) or vertically (Shift + V)
4. Try to rotate the canvas Clockwise (R) again

### System Information
+ Pencil2D Version:
```
Nightly Build Jul 18 2023
commit: f5a60600c5b8244ea1836b216fa625bf3c9befa9
date: 2023-07-18_16:43:10
```

+ Operating System:
Windows 10
 

+ RAM Size:
32GB
 

+ Graphics Tablet:
(irrelevant for this issue, but) WACOM CTL-470

HeCorr:
Fixes #1777.

_Please note that I am not an experienced C++ developer, so please let me know if something can be improved!_

This PR adds a check to `ActionCommands::rotateClockwise()`, `ActionCommands::rotateCounterClockwise()` and `HandTool::transformView()` that inverts the rotation direction if the view is flipped either horizontally or vertically (but not both), so it's more intuitive:

Shortcuts (Z then R):
https://github.com/pencil2d/pencil/assets/75134774/ac722475-312d-4ac9-aeee-42c81653dcd6

Hand Tool:
https://github.com/pencil2d/pencil/assets/75134774/5e6d216b-31dd-4e88-844d-06c87b3a86c9

### DIFF
--- a/app/src/actioncommands.cpp
+++ b/app/src/actioncommands.cpp
@@ -547,30 +547,16 @@ void ActionCommands::ZoomOut()
 
 void ActionCommands::rotateClockwise()
 {
-    float currentRotation = mEditor->view()->rotation();
-    // Invert rotation direction if view is flipped either vertically or horizontally
-    if (mEditor->view()->isFlipHorizontal() ^ mEditor->view()->isFlipVertical())
-    {
-        mEditor->view()->rotate(currentRotation - 15.f);
-    }
-    else
-    {
-        mEditor->view()->rotate(currentRotation + 15.f);
-    }
+    // Rotation direction is inverted if view is flipped either vertically or horizontally
+    const float delta = mEditor->view()->isFlipHorizontal() == !mEditor->view()->isFlipVertical() ? -15.f : 15.f;
+    mEditor->view()->rotateRelative(delta);
 }
 
 void ActionCommands::rotateCounterClockwise()
 {
-    float currentRotation = mEditor->view()->rotation();
-    // Invert rotation direction if view is flipped either vertically or horizontally
-    if (mEditor->view()->isFlipHorizontal() ^ mEditor->view()->isFlipVertical())
-    {
-        mEditor->view()->rotate(currentRotation + 15.f);
-    }
-    else
-    {
-        mEditor->view()->rotate(currentRotation - 15.f);
-    }
+    // Rotation direction is inverted if view is flipped either vertically or horizontally
+    const float delta = mEditor->view()->isFlipHorizontal() == !mEditor->view()->isFlipVertical() ? 15.f : -15.f;
+    mEditor->view()->rotateRelative(delta);
 }
 
 void ActionCommands::PlayStop()
--- a/app/src/actioncommands.cpp
+++ b/app/src/actioncommands.cpp
@@ -548,13 +548,29 @@ void ActionCommands::ZoomOut()
 void ActionCommands::rotateClockwise()
 {
     float currentRotation = mEditor->view()->rotation();
-    mEditor->view()->rotate(currentRotation + 15.f);
+    // Invert rotation direction if view is flipped either vertically or horizontally
+    if (mEditor->view()->isFlipHorizontal() ^ mEditor->view()->isFlipVertical())
+    {
+        mEditor->view()->rotate(currentRotation - 15.f);
+    }
+    else
+    {
+        mEditor->view()->rotate(currentRotation + 15.f);
+    }
 }
 
 void ActionCommands::rotateCounterClockwise()
 {
     float currentRotation = mEditor->view()->rotation();
-    mEditor->view()->rotate(currentRotation - 15.f);
+    // Invert rotation direction if view is flipped either vertically or horizontally
+    if (mEditor->view()->isFlipHorizontal() ^ mEditor->view()->isFlipVertical())
+    {
+        mEditor->view()->rotate(currentRotation + 15.f);
+    }
+    else
+    {
+        mEditor->view()->rotate(currentRotation - 15.f);
+    }
 }
 
 void ActionCommands::PlayStop()
--- a/core_lib/src/tool/handtool.cpp
+++ b/core_lib/src/tool/handtool.cpp
@@ -107,7 +107,7 @@ void HandTool::transformView(Qt::KeyboardModifiers keyMod, Qt::MouseButtons butt
         angleOffset = qRadiansToDegrees(angleOffset);
         float newAngle = 0.0;
         // Invert rotation direction if view is flipped either vertically or horizontally
-        if (viewMgr->isFlipHorizontal() ^ viewMgr->isFlipVertical())
+        if (viewMgr->isFlipHorizontal() == !viewMgr->isFlipVertical())
         {
             newAngle = viewMgr->rotation() - static_cast<float>(angleOffset);
         }
--- a/core_lib/src/tool/handtool.cpp
+++ b/core_lib/src/tool/handtool.cpp
@@ -105,7 +105,16 @@ void HandTool::transformView(Qt::KeyboardModifiers keyMod, Qt::MouseButtons butt
 
         qreal angleOffset = static_cast<qreal>(std::atan2(curV.y(), curV.x()) - std::atan2(startV.y(), startV.x()));
         angleOffset = qRadiansToDegrees(angleOffset);
-        float newAngle = viewMgr->rotation() + static_cast<float>(angleOffset);
+        float newAngle = 0.0;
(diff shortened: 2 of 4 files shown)


app/src/actioncommands.cpp
```diff
@@ -548,13 +548,29 @@ void ActionCommands::ZoomOut()
 void ActionCommands::rotateClockwise()
 {
     float currentRotation = mEditor->view()->rotation();
-    mEditor->view()->rotate(currentRotation + 15.f);
+    // Invert rotation direction if view is flipped either vertically or horizontally
+    if (mEditor->view()->isFlipHorizontal() ^ mEditor->view()->isFlipVertical())
```
MrStevns, line 552: Hi Henrique

Thanks again for contributing 😄 

I have a few things this time around, although the code itself is _fine_, I think we can do a better than simply copy/pasting, which could introduce bugs down the line.

Bitwise XOR is not something we use in the project afaik. I don't personally have a stance on that as I think there's something to be said about its simplicity but it could introduce bugs... as such I think it may be better to use conventional logical operators instead, like:
if (A && !B || B && !A)

app/src/actioncommands.cpp
```diff
@@ -548,13 +548,29 @@ void ActionCommands::ZoomOut()
...
+    {
+        mEditor->view()->rotate(currentRotation - 15.f);
```
MrStevns, line 554: I think given that we have the same logic in two places now, we should simplify and refactor this to a new private function.

Something like private ActionCommands::rotateInDirection(int angle) { ... }

core_lib/src/tool/handtool.cpp
```diff
@@ -105,7 +105,16 @@ void HandTool::transformView(Qt::KeyboardModifiers keyMod, Qt::MouseButtons butt
 
         qreal angleOffset = static_cast<qreal>(std::atan2(curV.y(), curV.x()) - std::atan2(startV.y(), startV.x()));
         angleOffset = qRadiansToDegrees(angleOffset);
-        float newAngle = viewMgr->rotation() + static_cast<float>(angleOffset);
+        float newAngle = 0.0;
+        // Invert rotation direction if view is flipped either vertically or horizontally
+        if (viewMgr->isFlipHorizontal() ^ viewMgr->isFlipVertical())
```
MrStevns, line 110: The same comment about bitwise XOR applies here.

MrStevns (changes requested):
Hi Henrique

Thanks again for contributing 😄 

I have a few things this time around, although the code itself is _fine_, I think we can do a better than simply copy/pasting, which could introduce bugs down the line.

There's also the usage of bitwise XOR, which is not something we use in the project afaik. I don't personally have a stance on that as I think there's something to be said about its simplicity but it could introduce bugs... as such I think it may be better to use conventional logical operators instead, like:
if (A && !B || B && !A)

Note: Don't feel disheartened by the lack of approval right away, there's nothing wrong with the code, it just needs a bit of tweaking. We strive for simple and readable while also trying to minimize the possibility of bugs.

app/src/actioncommands.cpp
scribblemaniac: That seems even less clear to me. I don't mind keeping the binary operator, but if we have to go with logical operators, `(A || B) && !(A && B)` seems a bit better to me.

app/src/actioncommands.cpp
scribblemaniac: I had a similar thought, but I think it would make more sense to have a rotate function in view that takes a relative angle, and maybe a bool for the direction (or just use the sign). The existing function could be changed or a new one could be created depending on if it is used anywhere else.

MrStevns: Ah right, yeah that sounds fine to me too

app/src/actioncommands.cpp
HeCorr: Both seem pretty unreadable to me personally  😅

@MrStevns Could you please elaborate on "it could introduce bugs"? I don't understand how a XOR operation could introduce bugs, specially since I thoroughly tested the behavior.

app/src/actioncommands.cpp
HeCorr: `ViewManager::rotate()` is used in:

image

Should I create something like `void rotateRelative(float delta);`?

HeCorr: Here's what it looks like (please note that the shown functions can be improved too):

image

app/src/actioncommands.cpp
HeCorr: also.. using operators like that would mean having to write "A" and "B" twice each, forcing me to declare them as constants which will make me have to choose between good naming and readability:

```C++
const bool h = mEditor->view()->isFlipHorizontal();
const bool v = mEditor->view()->isFlipVertical();

if (h && !v || v && !h)
// or
if ((h || v) && !(h && v))
```
vs
```C++
const bool flipHorizontal = mEditor->view()->isFlipHorizontal();
const bool flipVertical = mEditor->view()->isFlipVertical();

if (flipHorizontal && !flipVertical || flipVertical && !flipHorizontal)
// or
if ((flipHorizontal || flipVertical) && !(flipHorizontal && flipVertical))
```

To me personally it seems cluttered either way.

I found this one on [StackOverflow](https://stackoverflow.com/a/39473219) which seems pretty simple:
```C++
if (mEditor->view()->isFlipHorizontal() == !mEditor->view()->isFlipVertical())
```

app/src/actioncommands.cpp
HeCorr: Improved version

```C++
void ActionCommands::rotateClockwise()
{
    // Rotation direction is inverted if view is flipped either vertically or horizontally
    const float delta = mEditor->view()->isFlipHorizontal() == !mEditor->view()->isFlipVertical() ? -15.f : 15.f;
    mEditor->view()->rotateRelative(delta);
}

void ActionCommands::rotateCounterClockwise()
{
    // Rotation direction is inverted if view is flipped either vertically or horizontally
    const float delta = mEditor->view()->isFlipHorizontal() == !mEditor->view()->isFlipVertical() ? 15.f : -15.f;
    mEditor->view()->rotateRelative(delta);
}
```

app/src/actioncommands.cpp
MrStevns: > @MrStevns Could you please elaborate on "it could introduce bugs"? I don't understand how a XOR operation could introduce bugs, specially since I thoroughly tested the behavior.

Sure, though I didn't mean to imply that the code would break on its own, just that bitwise comparison can become brittle if the implementation changes later in a way that the compiler does not complain about and the person updating the code is not aware. 

In reality that may never happen, I just think it's better to be safe than smart in that regard. 😅 

If the others are fine with bitwise xor comparison, then let's go with that, otherwise what you proposed is also fine:
> if (mEditor->view()->isFlipHorizontal() == !mEditor->view()->isFlipVertical())

app/src/actioncommands.cpp
HeCorr: @MrStevns how does the `rotateRelative()` function above looks to you? can I commit that change?

it's essentially:
```C++
void ViewManager::rotateRelative(float delta)
{
    mRotation = mRotation + delta;
    updateViewTransforms();

    emit viewChanged();
}
```

MrStevns: Sure, that's fine with me

core_lib/src/tool/handtool.cpp
HeCorr: XOR removed as requested!

MrStevns (approved):
Changes looks good to me, no further comments.

Good job 👍

PR #1787, opened 2023-08-31
Add custom Windows tile

J5lx:
Just as it says in the title – a little enhancement for those not yet on Windows 11 which I prepared while working on #848. To try it locally, simply download and extract the Windows artifact from the “Checks” tab of this PR, right-click pencil2d.exe and select “Pin to Start”. This will of course become more useful in conjunction with the installer PR, but it is already functional on its own.

In order to generate the package resource index for the tile images locally, the makepri tool from the Windows SDK must be available – but it is not needed when building only the program itself.


MrStevns:
Can confirm it works
image

Well done!

app/app.pro
```diff
@@ -6,18 +6,11 @@
...
-TARGET = pencil2d
-QMAKE_APPLICATION_BUNDLE_NAME = Pencil2D
-
 CONFIG += precompile_header lrelease embed_translations
+QT += core widgets gui xml multimedia svg network
 
-DESTDIR = ../bin
-MOC_DIR = .moc
-OBJECTS_DIR = .obj
-UI_DIR = .ui
```
MrStevns, line 20: Was this removed on purpose? if yes, is it done implicitly now?

app/app.pro
```diff
@@ -6,18 +6,11 @@
 
 ! include( ../util/common.pri ) { error( Could not find the common.pri file! ) }
 
-QT += core widgets gui xml multimedia svg network
-
 TEMPLATE = app
-TARGET = pencil2d
-QMAKE_APPLICATION_BUNDLE_NAME = Pencil2D
```
MrStevns, line 13: I assume this is no longer necessary, because we have the info.plist defined?
Qt doesn't say much about that specific environment variable...

MrStevns (commented):
The changes looks good to me and as mentioned, it also works as expected. I just have one thing I'd like clarified before approving.

app/app.pro
MrStevns: > As you can imagine it is not done implicitly now, instead the files simply end up in qmake’s default locations. Are there any issues with this that you are worried about?

Not that I am aware of 😅 It's probably fine

MrStevns:
> Also since I just noticed your screenshot, this doesn’t actually do anything on Windows 11 since Microsoft removed start menu tiles in that version. On Windows 10, it looks something like this:
> 
> image

Ah... so it's only for Windows 10... well thanks for clarifying that 🤦

MrStevns:
> I wrote “for those not yet on Windows 11” in the OP, but perhaps I should have been more explicit, sorry for the confusion.

Ah that's on me, I managed to _skillfully_ the "not yet" part and did exactly what I shouldn't have then 😅

MrStevns (approved):
LGTM, merging

PR #1788, opened 2023-09-20
Fix OpenSSL build failure

J5lx:
This switches from the (now unavailable) OpenSSL builds provided by Qt to those provided by FireDaemon. For more info on those builds see [here](https://kb.firedaemon.com/support/solutions/articles/4000121705). These builds seem to be the best option right now, others are either the wrong version (1.0 or 3.x), are not up-to-date or do not provide 32-bit binaries.


MrStevns (approved):
Looks good to me! thanks for fixing this 👍

PR #1790, opened 2023-10-01
Fix perspective overlays not always covering the entire view

J5lx:
This approach uses its own line clipping implementation since Qt doesn’t seem to provide any functionality for this sort of use case.


MrStevns (changes requested):
Very nice work! I am happy to see this one fixed.

I was about to approve the PR but then noticed that the transform has gotten messed up during the implementation.

It does not rotate with canvas rotations
image

MrStevns:
Yup it works now, well done!

MrStevns (approved):

PR #1791, opened 2023-10-01
Fix heap use after free in MoveTool

J5lx:
Steps to reproduce:

- Use something like AddressSanitizer which detects heap use after free
- Open Pencil2D
- Switch to move tool
- Move cursor across ScribbleArea
- Ctrl-N to create new project
- heap use after free occurs in MoveTool::leavingThisTool because the move tool keeps an attribute with a pointer to the last used layer which it tries to access after the old project is already unloaded.

Solution: get rid of the pointer.

I also noticed the same kind of layer pointer attribute in SelectTool. Although I don’t think it’s causing any issues right now I removed it anyway because the layer can just as easily passed as a parameter there and storing it as an attribute basically just serves as a booby trap for the next person making changes to that tool.

### DIFF
--- a/core_lib/src/tool/movetool.cpp
+++ b/core_lib/src/tool/movetool.cpp
@@ -101,12 +101,12 @@ void MoveTool::updateSettings(const SETTING setting)
 
 void MoveTool::pointerPressEvent(PointerEvent* event)
 {
-    mCurrentLayer = currentPaintableLayer();
-    if (mCurrentLayer == nullptr) return;
+    Layer* currentLayer = currentPaintableLayer();
+    if (currentLayer == nullptr) return;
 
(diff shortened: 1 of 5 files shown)


MrStevns (approved):
I can confirm that the bug is reproducible and has been fixed with the changes in the PR.

### Discussion
While it's fine to remove pointers in this case, I think in general that we need to figure out a way to cleanup old state when loading a new file and such.

One way could be to make all tools listen to Editor::objectLoaded and cleanup all state, like it was a de-initialization, however that's still very manual and error prone.

We also have ToolManager::cleanupAllToolsData although that's seemingly unused right now.

Another would be to initialize all tools again, that's a small overhead for the convenience of having a clean slate, and it only needs to be done once...
What do you think?

Issue #1792, opened 2023-10-06
After making a selection with the Select tool, auto-switch to Move tool


shieldgenerator7:
### Issue Summary

Related discussion:
https://discuss.pencil2d.org/t/suggestion-implementing-automatic-move-tool-on-bitmap-layers/609

When the user makes a selection with the Select tool, automatically switch to the Move tool so they can edit the picture.

### Actual Results

Currently the user has to manually switch to the Move tool every time.

### Expected Results

Ideally the program would auto switch to the Move tool for the user.

### Steps to reproduce

1. Select the Select tool
2. Use the Select tool to select an area
3. Result: The Select tool is still the active tool, but it should be the Move tool

shieldgenerator7:
I'm working on this already

Jose-Moreno:
@shieldgenerator7 Hi thank you for this!

There is certainly an issue with the overall select-transform workflow in Pencil2D. From a single responsibility POV each tool has a singular functionality. It wouldn't make sense for the select tool to switch automatically to the move tool without the user input.

However from an actual animation workflow, it is much desired to have this! Which is why most of the time there is a single tool implemented to perform both Selection & basic transformation operations on raster & vector artwork, particularly in software like Adobe Animate or Toonboom Harmony which Pencil2D shares commonalities with.

For this particular request I don't see any issue. Having such a helper function for those who want this would be probably a huge performance boost 🥳

PR #1793, opened 2023-10-06
#1792 Autoswitch to Move tool after selecting

shieldgenerator7:
The feature is implemented, and a checkbox option has been added to the Select tool to toggle this feature on and off. The default value is true.

I'd like this to be merged into the official program, and if not, I'd like feedback on how I can improve my code so it more closely follows the rules of this repository.


MrStevns:
Hi shieldgenerator7

First, thanks for contributing to Pencil2D.

Before reviewing the PR, I'd like to hear about why you think this is necessary and what you solve by implementing a checkbox to switch tool, that we can't do already.

For reference:
In the current version you can move a selection by holding down 'alt', which will temporarily activate another tool, in this case the move tool. 

My thoughts:
An improvement to that feature could be to allow the us to stay on the move tool until we're satisfied with the transformation and then go back to the selection tool or apply it, depending on the action. In some applications this is known as starting a transformation of the selection, allowing you to rotate, scale and move it around temporarily.

Implementation wise this could either by done via a button, something like a "Transform" button in the toolbox widget and maybe a shortcut as an alternative way to trigger the behavior. At that point though it would also make sense to create a new widget for handing the tool properties of the selection tool.

The overall task is more complicated as you now need to communicate back and forth between the tools but I do think that it makes more sense than a checkbox to decide to stay on the select tool or switch automatically to the move tool.

What do you think?

MrStevns:
Closing this PR due to lack of activity and feedback.

PR #1796, opened 2023-11-09
Fix deprecated top-level developer_name in AppData XML

musicinmybrain:
Use the `name` element in a `developer` block instead, as recommended by `appstreamcli` 1.0.0.

This fixes all warnings when validating the AppData XML file, although there are still informational messages:

```
I: org.pencil2d.Pencil2D:6: developer-id-missing
   The `developer` element is missing an `id` property, containing a unique string ID for the
   developer. Consider adding a unique ID.

✔ Validation was successful: infos: 1, pedantic: 2
```

### DIFF
--- a/app/data/org.pencil2d.Pencil2D.metainfo.xml
+++ b/app/data/org.pencil2d.Pencil2D.metainfo.xml
@@ -3,7 +3,9 @@
   <id>org.pencil2d.Pencil2D</id>
   <launchable type="desktop-id">org.pencil2d.Pencil2D.desktop</launchable>
   <name>Pencil2D</name>
-  <developer_name>The Pencil2D Team</developer_name>
+  <developer>
+    <name>The Pencil2D Team</name>
+  </developer>
   <metadata_license>CC0-1.0</metadata_license>
   <project_license>GPL-2.0</project_license>
   <summary>2D animation software supporting bitmap and vector graphics</summary>


musicinmybrain:
Let me clarify that `appstream` 1.0.0 has not yet been released, and my `appstream` is a pre-release as [currently packaged](https://src.fedoraproject.org/rpms/appstream/blob/51384dca2599ac6ff1895beb927ceb6e28018adc/f/appstream.spec) in Fedora Linux Rawhide (the development version of the OS), corresponding to https://github.com/ximion/appstream/commit/d88ed03cb5b3b1803bdee3528c9b99d528ceb065. See also https://github.com/ximion/appstream/issues/244.

musicinmybrain:
> Thanks for the PR. I see you also removed the previous developer_name tag, but won’t this degrade compatibility with current versions, considering that 1.0 has not been released yet? Would it be fine to keep developer_name for now (alongside the new tag) until more distributions have updated to 1.0?

Annoyingly, `appstreamcli validate` emits a warning (`W`) when `developer_name` is present, and adding `developer` doesn’t change that. Also annoyingly, it fails (exits with nonzero status) when there are warnings.

> On a completely unrelated note, the other day I saw a review (from ODRS I think) saying that the tool icons weren’t showing up on their Fedora system. Could it be that the Fedora package is missing a runtime dependency on the qtsvg 5 (specifically the plugins it provides)?

I just tested this by starting with a Fedora Server VM image from https://download.fedoraproject.org/pub/fedora/linux/releases/39/Server/x86_64/images/Fedora-Server-KVM-39-1.5.x86_64.qcow2, then installing the `Basic Desktop` group and `Pencil2D`. I was able to reproduce the problem, and installing `qt5-qtsvg` fixed it, so I’ll add a manual dependency. Normally, Qt library dependencies like this are handled automatically because something dynamically links one of their shared objects. I guess that didn’t happen because the plugins are conditionally `dlopen`’ed at runtime. Thanks for pointing this out.

PR #1798, opened 2023-11-24
Fix vector layer color import

J5lx:
This fixes the crash reported [over here](https://discuss.pencil2d.org/t/loading-layers-into-a-project-indicate-to-the-user-when-the-process-is-finished/7510?u=j5lx) on the forum. The crash was caused by the dangling pointer to the imported file’s `Object` dereferenced in the vector paint code in order to resolve colour indices.

This PR completely removes `Object` pointers from the vector classes and instead passes the object as a parameter at paint time. (Tbh I’m also thinking about splitting out the painting code altogether, but I didn’t want to let myself get derailed.) It also removes `Object` pointers from layer classes, which only used it in the constructor and `setObject` to generate layer ids, and instead passes those ids directly.

Lastly, a proper colour import mechanism for vector layer import is implemented as well.

### DIFF
--- a/tests/src/test_layer.cpp
+++ b/tests/src/test_layer.cpp
@@ -26,43 +26,37 @@ GNU General Public License for more details.
 
 TEST_CASE("LayerType")
 {
-    Object* object = new Object;
-
     SECTION("Bitmap Layer")
     {
-        Layer* bitmapLayer = new LayerBitmap(object);
+        Layer* bitmapLayer = new LayerBitmap(1);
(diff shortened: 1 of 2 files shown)


MrStevns (approved):
Excellent work figuring out that crash and how to fix it!
I also like the removal of Object in Layer, as I find that LayerManager fulfill that responsibility.

I've looked through the changes and I think they all look good. Feel free to merge when you're satisfied with the PR.

PR #1800, opened 2023-12-03
The effect is only evident when using an alpha value less than 100%

### LINKED ISSUE #1799: Pen dabs are being added on top of each other
### Expected Results
The pen dabs should merge, so they are not seen individually

 

### Video or Image Reference
pen-dab-stable

The video first shows the stroke in 0.6.6 stable and then switches to our nightly build

 

### Steps to reproduce
* Select pen tool
* Draw a stroke with 30%

 

### System Information
+ Pencil2D Version:
Nightly Build Oct 17 2023

commit: 127cd62212311b6b72a731c7a96b74021801bec4
date: 2023-10-17_21:30:25

Operating System: macOS 12.6
CPU Architecture: x86_64
Qt Version: 5.15.10

MrStevns:
Fixes #1799 which was reported on Discord


MrStevns:
Ah you're right, that seems to be the case with master as well though but it's likely my fault too and related, so I'll look into it. If only i had written some tests to avoid these cases 😔

There's still room for improvements regarding separation of concern 🙃

MrStevns:
I've had some time to think and experimented a bit.. try again with the latest changes.

Also thanks for reviewing and being meticulous 😄

PR #1801, opened 2023-12-04
Improve support for animated image importing

scribblemaniac:
This PR contains a few loosely related changes to image importing. Specifically:
- Add support for importing animated WebP images (same Qt code that handles animated GIF import)
- Progress dialog for animated image import with abort support
- Separate logic handling single and animated image importing
- Make movie export modal so the user can't perform other operations that might interfere with the export ([report](https://discuss.pencil2d.org/t/accidentally-using-a-drawing-tool-on-a-bitmapped-layer-whilst-pencil2d-is-expecting-an-mp4-video-file/7599?u=j5lx))
- Fix possible issue with error checking during image importing. I did not try to produce the issue, but based on the code it would probably handle some image corruption errors incorrectly.
- Add non-animated WebP image import/export

### DIFF
--- a/app/src/doubleprogressdialog.h
+++ b/app/src/doubleprogressdialog.h
@@ -18,14 +18,14 @@ GNU General Public License for more details.
 #ifndef DOUBLEPROGRESSDIALOG_H
 #define DOUBLEPROGRESSDIALOG_H
 
-#include <QDialog>
+#include <QProgressDialog>
 #include <QProgressBar>
 
 namespace Ui {
 class DoubleProgressDialog;
 }
 
-class DoubleProgressDialog : public QDialog
+class DoubleProgressDialog : public QProgressDialog
 {
     Q_OBJECT
 
@@ -63,8 +63,8 @@ public:
 
     ProgressBarControl *major, *minor;
 
-signals:
-    void canceled();
+//signals:
+    //void canceled();
 
 private:
     Ui::DoubleProgressDialog *ui;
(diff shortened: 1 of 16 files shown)


app/src/doubleprogressdialog.h
```diff
@@ -63,8 +63,8 @@ class DoubleProgressDialog : public QDialog
 
     ProgressBarControl *major, *minor;
 
-signals:
-    void canceled();
+//signals:
+    //void canceled();
```
MrStevns, line 67: Is this supposed to be commented out? if yes, let's remove the dead code 😃

app/src/importimageseqdialog.cpp
```diff
@@ -58,9 +58,15 @@ void ImportImageSeqDialog::setupLayout()
...
+    {
+    case FileType::GIF:
         setWindowTitle(tr("Import Animated GIF"));
-    } else {
+        break;
+    case FileType::IMAGE_SEQUENCE:
+        setWindowTitle(tr("Import animated image"));
+        break;
+    default:
         setWindowTitle(tr("Import image sequence"));
```
MrStevns, line 70: I think you accidentally switched these around, now the title will be "import animated image" for an image sequence, and "import image sequence" for animated images 🙃

MrStevns (changes requested):
Overall i like the changes, there's just one thing I bumped into that should be fixed.

The title for the import dialog is now saying "animated image" when it should be sequence and vise versa.

MrStevns (approved):
LGTM 👍

PR #1807, opened 2024-01-17
Fix stroke not being shown on a selection while drawing

MrStevns:
How to reproduce:
1. Create a selection
2. Draw a stroke inside it

Expected:
Stroke is drawn

Actual:
Stroke is not drawn until you stop drawing.

### DIFF
--- a/core_lib/src/canvaspainter.cpp
+++ b/core_lib/src/canvaspainter.cpp
@@ -315,7 +315,7 @@ void CanvasPainter::paintCurrentBitmapFrame(QPainter& painter, const QRect& blit
     }
 
     // We do not wish to draw selection transformations on anything but the current layer
-    if (isCurrentLayer && mRenderTransform) {
+    if (isCurrentLayer && mRenderTransform && !isDrawing) {
         paintTransformedSelection(currentBitmapPainter, paintedImage, mSelection);
     }
 


MrStevns:
Hi Jacob, I hope you're doing well. I know that life can get in the way, so don't feel bad about not having reviewed the other PR yet, we'll get there eventually. 😃 

You're right about that other bug! I completely forgot about the temporary move state 😅 ... The transformation should actually be in a identity state in cases where you are able to draw on the canvas again, because the "leavingThisTool" function should commit the transformation before going back to another tool.

I've pushed a fix for that too now.

scribblemaniac:
@MrStevns @J5lx I've had a chance to look at this. The tool change seem sensible. However, this isn't enough to solve the issue completely. If you create a selection, switch to a drawing tool, and move the selection with the arrow shortcuts, and then draw a stroke, then transform is not identity and isDrawing is true.

MrStevns:
@scribblemaniac

I am not able to reproduce that behavior, are you referring to the 'M' shortcut? Once you switch from the Move tool back to the drawing tool, using a shortcut or not, that should still trigger the LeavingTool function and apply the transformation.
Or is there another shortcut that I've forgotten about?

https://github.com/pencil2d/pencil/assets/1045397/dd38c2d7-6b57-4f15-986a-f26f6e97d530

MrStevns:
> If you create a selection, switch to a drawing tool, and move the selection with the arrow shortcuts, and then draw a stroke, then transform is not identity and isDrawing is true.

I've investigated this and can now say that it's a bug. Moving the selection when the current tool is Select, should not modify the keyframe, and therefore not the selection transform either, just the selection rect.
I'll address this in my upcoming PR for floating selection.

PR #1814, opened 2024-03-08
Limit StrokeManager use to stroke tools

J5lx:
While looking at the little glitch I discovered when reviewing #1806, I realised that there are quite a few components that indirectly rely on StrokeManager despite not even dealing with strokes and not requiring any StrokeManager functionality, in particular the non-stroke tools. In many of those cases, StrokeManager is used from somewhere within pointer event handling code to obtain the current cursor position even though that position is readily available from the event object.

This PR replaces those StrokeManager usages and moves StrokeManager instance from ScribbleArea to StrokeTool (where it is actually needed). This way:

- Non-stroke tools are more self-contained with one less dependency to worry about
- It is clearer where position data is coming from (e.g. `event->canvasPos()` vs `getCurrentPoint()`) as there is one less layer of indirection
- ScribbleArea no longer has to concern itself with the concept of strokes as it’s left entirely to StrokeTool and its subclasses

Additional changes:
- StrokeManager renamed to StrokeInterpolator to clearly distinguish it from BaseManager-derived classes and because SomethingManager is always a rather vague name
- For convenience, PointerEvent now has canvasPos in addition to posF (now viewportPos), similar to how QMouseEvent has localPos, windowPos and globalPos
- Fixed a bug where the camera tool showed the wrong cursor when dragging a path point after navigating away from a camera keyframe while hovering over one of the camera handles

Although the number of changes in this PR is relatively higher, many of them are fairly trivial. After submitting the PR, I’ll leave some comments on the few changes that might be less obvious.

### DIFF
--- a/core_lib/src/util/pointerevent.cpp
+++ b/core_lib/src/util/pointerevent.cpp
@@ -1,34 +1,27 @@
 #include "pointerevent.h"
 
-PointerEvent::PointerEvent(QMouseEvent* event)
+PointerEvent::PointerEvent(QMouseEvent* event, const QPointF& canvasPos)
 {
     mMouseEvent = event;
+    mCanvasPos = canvasPos;
 }
 
-PointerEvent::PointerEvent(QTabletEvent* event)
+PointerEvent::PointerEvent(QTabletEvent* event, const QPointF& canvasPos)
 {
     mTabletEvent = event;
+    mCanvasPos = canvasPos;
 }
 
 PointerEvent::~PointerEvent()
 {
 }
 
-QPoint PointerEvent::pos() const
+QPointF PointerEvent::canvasPos() const
 {
-    if (mMouseEvent)
-    {
-        return mMouseEvent->pos();
-    }
-    else if (mTabletEvent)
-    {
-        return mTabletEvent->pos();
-    }
-    Q_ASSERT(false);
-    return QPoint();
+    return mCanvasPos;
 }
 
-QPointF PointerEvent::posF() const
+QPointF PointerEvent::viewportPos() const
 {
     if (mMouseEvent)
     {
--- a/core_lib/src/util/pointerevent.h
+++ b/core_lib/src/util/pointerevent.h
@@ -21,19 +21,20 @@ public:
         Unmapped
     };
 
-    PointerEvent(QMouseEvent* event);
-    PointerEvent(QTabletEvent* event);
+    PointerEvent(QMouseEvent* event, const QPointF& canvasPos);
+    PointerEvent(QTabletEvent* event, const QPointF& canvasPos);
     ~PointerEvent();
 
     /**
-     * Returns QPoint of the device */
-    QPoint pos() const;
+     * Returns the QPointF of the device, in canvas coordinates
+     */
+    QPointF canvasPos() const;
 
     /**
-     * Returns the QPointF of the device
+     * Returns the QPointF of the device, in viewport coordinates
      * Returns pos() if used on mouse event
      */
-    QPointF posF() const;
+    QPointF viewportPos() const;
 
     /**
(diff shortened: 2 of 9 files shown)


core_lib/src/tool/stroketool.h
```diff
@@ -66,6 +68,13 @@ public slots:
     void onViewUpdated();
 
 protected:
+    QPointF getCurrentPressPixel() const;
+    QPointF getCurrentPressPoint() const;
+    QPointF getCurrentPixel() const;
+    QPointF getCurrentPoint() const;
+    QPointF getLastPixel() const;
+    QPointF getLastPoint() const;
```
MrStevns, line 76: I know that currently it's not like this but if we're moving things around anyway, then we might as well improve it a bit.
```suggestion
    const QPointF& getCurrentPressPixel() const;
    const QPointF& getCurrentPressPoint() const;
    const QPointF& getCurrentPixel() const;
    const QPointF& getCurrentPoint() const;
    const QPointF& getLastPixel() const;
    const QPointF& getLastPoint() const;
```

core_lib/src/util/pointerevent.cpp
```diff
@@ -1,34 +1,27 @@
...
-    {
-        return mMouseEvent->pos();
-    }
-    else if (mTabletEvent)
-    {
-        return mTabletEvent->pos();
-    }
-    Q_ASSERT(false);
-    return QPoint();
+    return mCanvasPos;
```
MrStevns, line 21: I think I understand why you did this, to avoid coupling PointerEvent to ViewManager right? I would prefer if it wasn't up to the init method to decide how this value is interpreted but i'm also not sure how to solve it without creating that coupling.

My issue with this right now is that one could be deceived to think that canvasPos is always the canvas position mapped based on ViewManager but in reality it could be something else, depending on the next time someone initializes a PointerEvent.

core_lib/src/util/pointerevent.h
```diff
@@ -21,19 +21,20 @@ class PointerEvent
...
+     */
+    QPointF canvasPos() const;
 
     /**
-     * Returns the QPointF of the device
+     * Returns the QPointF of the device, in viewport coordinates
      * Returns pos() if used on mouse event
      */
-    QPointF posF() const;
+    QPointF viewportPos() const;
```
MrStevns, line 37:
```suggestion
    const QPointF& viewportPos() const;
```

MrStevns (commented):
A much appreciated decoupling that I have thought about doing so many times.
Excellent that you got around doing it!

I've looked through the code and it looks good to me, well aside from some suggestions about reference access but that's minor stuff.

My reason for neither approving nor requesting changes is that I'd like to hear your thoughts about the PointerEvent part that I mentioned before concluding my review. Is it more important that we're decoupled from ViewManager but have higher potential for creating mapping bugs when initializing PointerEvent, than either introducing a interface that sets the expectation for how the points are mapped or depend on ViewManager explicitly? 🤔 

Speaking of testing the changes myself, I've tried to use all the different tools with layer combinations and it works as expected, good job 👍

core_lib/src/tool/stroketool.h
MrStevns: Ah you're right, because it's being mapped first. 💡

core_lib/src/util/pointerevent.cpp
MrStevns: Fair enough, we'll leave it as is 👍

core_lib/src/util/pointerevent.h
MrStevns: Disregard this, I forgot that we're creating an empty QPointF at the bottom, which wouldn't work with a temporary reference.

MrStevns (approved):
LGTM! If you're done, you may merge it

PR #1815, opened 2024-03-08
Fix AppStream developer id

musicinmybrain:
Per https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-developer, this should have been `org.pencil2d` instead of `pencil2d.org`. Technically, the ID can be any string, but the FreeDesktop documentation strongly recommends a reverse-DNS string.

This is a follow-up to https://github.com/pencil2d/pencil/pull/1796.


musicinmybrain:
Thanks for the quick review!

I agree, even as a downstream packager, the recent churn in the AppStream XML format and the way new warnings keep popping up that cause validation failures on previously-valid files has been a pain to keep up with. (AppStream 1.0.2 started warning when `<developer/>` has no `id`, so at least we were ahead of the game there.)

I hope things are starting to stablilize now!

PR #1818, opened 2024-04-10
Fix delayed vector transformation rendering

MrStevns:
This PR fixes a small cosmetic bug I stumbled upon while working on the new undo/redo feature.

How to reproduce:
1. Draw a vector stroke
2. Select it using the Select tool
3. Move or transform using the Move tool
4. undo/redo the operation and notice that it might either glitch and show the previous state then the new state, or it might not show the new state until you stop holding down ctrl or cmd on mac.

I also simplified the rendering logic a bit, check the commits for clarification about what does what.


chchwy:
Confirmed the issue is fixed.

chchwy (approved):

PR #1820, opened 2024-04-13
Fix crash when importing image to empty frame on hidden layer

J5lx:
This fixes the crash reported at <https://discuss.pencil2d.org/t/pencil2d-crashes-when-i-import-an-image-to-a-hidden-layer/8016?u=j5lx>.

- Also fixed a potential memory leak in Layer::addNewKeyFrameAt
- Also disabled timeline hover effect for hidden layers since double-click-to-add-keyframe is disabled for those
- Also implicitly fixed an issue where trying to import to hidden layers showed a warning for empty frames but not on keyframes


MrStevns (approved):
Crash confirmed fixed.

Additional changes looks good to me too 👍

PR #1822, opened 2024-04-15
Fix polyline buffer not always being cleared

scribblemaniac:
First reported here: https://discuss.pencil2d.org/t/problems-with-nightly-build-11-april-2024/8025

On the bitmap layer, applying a polyline path results in ScribbleArea::paintBitmapBuffer being called, which clears the mTiledBuffer. However, this function is never called when using the tool on the vector layer where the the curve is added by the tool. It will also never get cleared if clearToolData is called directly, but it seems that this function isn't currently being called outside the polyline tool. I've refactored it a bit so when clearToolData is called, the buffer is always cleared (and always after deactivating the tool so ScribbleArea::updateFrame works). One slight change is that isActiveChanged will only be triggered if mPoints is not empty, which makes sense because if it is empty, then the tool was not active and the state is not changing.

### DIFF
--- a/core_lib/src/tool/polylinetool.cpp
+++ b/core_lib/src/tool/polylinetool.cpp
@@ -100,7 +100,6 @@ bool PolylineTool::leavingThisTool()
     if (mPoints.size() > 0)
     {
         cancelPolyline();
-        clearToolData();
     }
     return true;
 }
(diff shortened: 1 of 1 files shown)


chchwy (approved):

chchwy:
I can confirm the issue is fixed.

BTW, is it considered a bug that polylines turn into smooth lines when using the Smudge tool?

scribblemaniac:
> BTW, is it considered a bug that polylines turn into smooth lines when using the Smudge tool?

Probably yes, but I would be hesitant to fix it when some of our users are currently relying on that behavior for their workflow.

chchwy:
> > BTW, is it considered a bug that polylines turn into smooth lines when using the Smudge tool?
> 
> Probably yes, but I would be hesitant to fix it when some of our users are currently relying on that behavior for their workflow.

Alright, I guess we are having another example of the classic software developer joke "it's not a bug, it's a feature!"
img

PR #1823, opened 2024-04-17
Remove "macdeployqtfix" from our deployment script

chchwy:
The use of `macdeployqtfix` is no longer needed as the official Qt deployment tool `macdeployqt` works as expected.

I carefully checked the Mac app bundles before and after applying `macdeployqtfix`, and I didn't see any differences between the two. Therefore, it should be safe to remove `macdeployqtfix` from our deployment process.

The only scenario where `macdeployqtfix` may be required is building the legacy Mac version, which is built with Qt 5.6 as far as I know.

### DIFF
--- a/.github/actions/create-package/create-package.sh
+++ b/.github/actions/create-package/create-package.sh
@@ -94,15 +94,7 @@ create_package_macos() {
   echo "::group::Deploy Qt libraries"
   macdeployqt Pencil2D.app
   echo "::endgroup::"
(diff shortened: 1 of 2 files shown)


scribblemaniac (approved):
I can't personally test this, but as long as you have I'm fine with this being merged. If there's a problem for some specific version of macOS we'll hopefully find out when someone complains about the nightly builds.

PR #1824, opened 2024-04-19
Make OnionSkins UI more compact 

chchwy:
Improve the OnionSkins UI by reducing the spacing between UI components.

Currently, on my HiDPI laptop displays (Win11 with 150% scaling), the OnionSkins options are barely visible. We can improve it by minimizing the spacing between UI elements, so more options can be displayed within the limited available space.

Please see the before/after screenshots as follows:

### Before
onionskins-old

### After
onionskins-new

### DIFF
--- a/app/ui/onionskin.ui
+++ b/app/ui/onionskin.ui
@@ -6,14 +6,14 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>213</width>
-    <height>382</height>
+    <width>219</width>
+    <height>238</height>
    </rect>
   </property>
   <property name="minimumSize">
    <size>
-    <width>187</width>
-    <height>122</height>
+    <width>190</width>
+    <height>124</height>
    </size>
   </property>
   <property name="floating">
@@ -30,6 +30,9 @@
     </sizepolicy>
    </property>
    <layout class="QVBoxLayout" name="verticalLayout">
+    <property name="spacing">
+     <number>0</number>
+    </property>
     <property name="leftMargin">
      <number>0</number>
     </property>
@@ -76,8 +79,8 @@
         <rect>
          <x>0</x>
          <y>0</y>
-         <width>213</width>
-         <height>360</height>
+         <width>219</width>
+         <height>214</height>
         </rect>
        </property>
        <property name="sizePolicy">
@@ -87,17 +90,20 @@
         </sizepolicy>
        </property>
        <layout class="QVBoxLayout" name="verticalLayout_3">
+        <property name="spacing">
+         <number>3</number>
+        </property>
         <property name="leftMargin">
-         <number>6</number>
+         <number>3</number>
         </property>
         <property name="topMargin">
-         <number>6</number>
+         <number>3</number>
         </property>
         <property name="rightMargin">
-         <number>6</number>
+         <number>3</number>
         </property>
         <property name="bottomMargin">
-         <number>6</number>
+         <number>3</number>
         </property>
         <item>
          <widget class="QGroupBox" name="prevFramesGroup">
@@ -108,6 +114,9 @@
            <bool>true</bool>
           </property>
           <layout class="QHBoxLayout" name="onionPrevLayout">
+           <property name="spacing">
+            <number>3</number>
+           </property>
            <property name="sizeConstraint">
             <enum>QLayout::SetDefaultConstraint</enum>
            </property>
@@ -115,13 +124,13 @@
             <number>0</number>
            </property>
            <property name="topMargin">
-            <number>6</number>
+            <number>3</number>
            </property>
            <property name="rightMargin">
             <number>0</number>
            </property>
            <property name="bottomMargin">
-            <number>6</number>
+            <number>3</number>
            </property>
            <item>
             <widget class="QSpinBox" name="onionPrevFramesNumBox">
@@ -189,6 +198,9 @@
            <bool>true</bool>
           </property>
           <layout class="QHBoxLayout" name="onionNextlLayout">
+           <property name="spacing">
+            <number>3</number>
+           </property>
            <property name="sizeConstraint">
             <enum>QLayout::SetDefaultConstraint</enum>
            </property>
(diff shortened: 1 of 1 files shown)


scribblemaniac (changes requested):
While I agree this change is an improvement on Windows, it doesn't look as good on Linux/GNOME. On the left is your version, the right is the old version (the font is slightly different in the AppImage apparently):
Screenshot from 2024-04-21 11-52-59
Because the border is drawn underneath the group box text instead of through it, it looks to me like there isn't enough padding on the top. I think we should add 3px to the `LayoutTopMargin` of `nextFramesGroup` and `prevFramesGroup` so that it's evenly spaced on Linux. While OS specific code could be added, I don't think it's worth it for 6px.

chchwy:
> While I agree this change is an improvement on Windows, it doesn't look as good on Linux/GNOME. On the left is your version, the right is the old version (the font is slightly different in the AppImage apparently): Screenshot from 2024-04-21 11-52-59 Because the border is drawn underneath the group box text instead of through it, it looks to me like there isn't enough padding on the top. I think we should add 3px to the `LayoutTopMargin` of `nextFramesGroup` and `prevFramesGroup` so that it's evenly spaced on Linux. While OS specific code could be added, I don't think it's worth it for 6px.

thanks for testing. I will tweak the spacing to see if I can make it look better.

chchwy:
BTW, this not only improves the Windows app but also enhances the Mac app, which by default uses a HiDPI display

Original
old

Compact
compact

scribblemaniac:
Maybe we shouldn't use QGroupBox here if the display is so inconsistent and keeping the layout compact is a priority?

chchwy:
@J5lx @scribblemaniac I've adjusted the margins, could you please test it again?

MrStevns:
It's not quite there imo, the padding is off on macOS.
image

I agree with Jacob that trying to sub-optimize this to shave off a few pixels is a bit silly, when we should instead consider either rethinking it or creating a custom UI that can be made smaller.

chchwy:
> It's not quite there imo, the padding is off on macOS. image
> 
> I agree with Jacob that trying to sub-optimize this to shave off a few pixels is a bit silly, when we should instead consider either rethinking it or creating a custom UI that can be made smaller.
> 
> Spacing should be consistent between our widgets, anything else is a UI blunder if you ask me. image
> 
> This, as it was, is better because it's aligned and there's space. Padding is important to make UI not feel cramped. image

chchwy:
> It's not quite there imo, the padding is off on macOS. image

I don't see obvious problems in this screenshot. can you elaborate more? 

> 
> I agree with Jacob that trying to sub-optimize this to shave off a few pixels is a bit silly, when we should instead consider either rethinking it or creating a custom UI that can be made smaller.
> 
> Spacing should be consistent between our widgets, anything else is a UI blunder if you ask me. image
> 
> This, as it was, is better because it's aligned and there's space. Padding is important to make UI not feel cramped. image

May I ask what your screen size is and the resolutions? The discussion will go in the wrong direction if we only talk about aesthetic. The change here want to strike a balance between the aesthetic and usability with HiDPI display.

chchwy:
> > Maybe we shouldn't use QGroupBox here if the display is so inconsistent and keeping the layout compact is a priority?
> 
> I don’t think this is something we should even have to worry about. To be completely honest, I’m not entirely convinced of this micromanagement of spacing and margin values in the first place. This will shave off a few pixels now, but the moment we make more meaningful changes (like the addition of [an onion skin global toggle](/pencil2d/pencil/issues/1085)), we’ll be back to square one. If we’re really that concerned about keeping all the stuff on screen, we should just arrange (some of) the tool windows horizontally next to each other by default, rather than stacking them all vertically, or reduce the overall UI scale, or lock the tool windows by default in order to conserve the space consumed by their titlebars. I worry that if we instead choose to micromanage our pixels like this we’ll just end up dealing with more edge-cases like these groupboxes in the future.

I don't really agree that space reduction is micro-management. 

The things you mentioned above, the global onion toggle and horizontal tools options, they are all valuable and should be considered seriously. However, these thins do not conflict with the space reduction. 

If you look at other tools such as Photoshop, Paint Tool SAI or Procreate, you may notice that they are extremely reluctant to put spaces between UI components in order to make the most use of limited screen spaces.

I want to emphasize again that I want to improve the usability with HiDPI display as they are becoming more and more common the current UI usability with HiDPI display is quite poor.

MrStevns:
> May I ask what your screen size is and the resolutions? The discussion will go in the wrong direction if we only talk about aesthetic. The change here want to strike a balance between the aesthetic and usability with HiDPI display.

My resolution is technically (2560 × 1600) but I have scaled UI, so it appears to be 1440x900. The screen is 13 inches.

I understand that you want to strike a balance between aesthetic and usability, which is fine, as long as it's consistent across our widgets. For example if we should have 2 or 3 px instead of 6 px padding everywhere, then that's fine too, as long as it's consistent.

You can see the padding difference here:
image

Looking at the code, the padding is 3 px around the overall group, yet there appears to be more spacing on the right side.
From a bit of digging, it seems to come from the checkboxes not being placed in a layout.

The interesting thing here is that if I tweak the spacing, the font size and try to squish the layout as much as possible, the layout still ends up being wider than your the changes you made, which just shows that tweaking this has all kinds of weird outcomes.

image

Top: My tweaked layout
Behind: Your tweaked layout

When I compare our layout with my CSP layout, ours is 25-30 px wider, but you can also see that they do all kinds of trickery to get it smaller, like squishing the size, potentially hiding it partially behind UI, having generally a more compact layout by more or less embedding the description into the slider component.

| CSP | Pencil2D (chchwy) | Pencil2D (MrStevns) |
| ---- | --------- | ------- |
| image | image | image |
| 165 px | 188 px | 192 px |

What we can do is change the minimum component size too, which apparently is 192 px in my case. Changing this manually to, we could probably go down to 140 pixels and it would look alright, that's of course with the tweaked font size too. Would that be sufficient?

| 11 point font size | 13 point font size | 
| ----------------- | ----------------- | 
image | image
| 142 px | 162 |

chchwy:
Thanks a lot for the efforts. I was thinking of starting small from a single widget and then step by step making the changes to other widgets. But you have done them all! That's fantastic. I will definitely check your branch out when I get time.

chchwy:
BTW, I also thinking of replacing the title bar of dock widgets. The default title bars eat a lot of spaces.

MrStevns:
I agree, changing the titlebar widgets or getting rid of them entirely in docked mode is also something I've experimented with during my early timeline mockups. Integrating the close and popup button directly into the given widget would make the UI a lot more compact vertically, on mac os at least.

this is a mockup from 2020, so the style does not fit anymore but the concept itself could probably still be applied.
image

PR #1825, opened 2024-04-23
Enable C++17 by default when building against Qt 6

J5lx:
Also cleans up other compiler options a bit:

- use strict_c/c++ QMake feature to disable compiler extensions instead of messing directly with compiler flags
- same with msvc_mp and utf8_source. the latter also does compiler detection so there should be no need for special handling of legacy builds
- -stdlib=libc++ for macOS is removed since it appears to be the default since Qt 5.1
- QMAKE_LINK{,_SHLIB} = $$QMAKE_CXX also appears to be the default

### DIFF
--- a/util/common.pri
+++ b/util/common.pri
@@ -18,32 +18,23 @@ greaterThan(QT_MAJOR_VERSION, 5) {
     CONFIG += c++11
 }
 
-win32-msvc* {
-    QMAKE_CXXFLAGS += /MP /utf-8
-    CONFIG(release,debug|release) {
-        QMAKE_CXXFLAGS += /Gy /GL
-        CONFIG += ltcg
-        CONFIG += force_debug_info
-    }
+# utf8_source is only for Qt 5, it is the default since Qt 6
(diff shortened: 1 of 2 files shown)


util/common.pri
```diff
@@ -11,39 +11,30 @@ PENCIL2D_RELEASE {
...
-        QMAKE_CXXFLAGS += /Gy /GL
-        CONFIG += ltcg
-        CONFIG += force_debug_info
-    }
+# utf8_source is only for Qt 5, it is the default since Qt 6
+CONFIG += msvc_mp utf8_source
+
+win32-msvc*:CONFIG(release,debug|release) {
+    QMAKE_CXXFLAGS += /Gy /GL
+    CONFIG += ltcg
```
chchwy, line 26: As you are working on this part, it would be nice to remove `/Gy /GL` and `CONFIG+=ltcg` since they don't make a noticable difference if Qt itself is not compiled with LTCG enabled,

chchwy (approved):

PR #1826, opened 2024-04-23
Enable newly added translations

J5lx:
This allows the translations added in 4d4ff5a007145f46e238e632ab61e06bcaaa998d to be used. I’m not 100% sure about 粵语 for Cantonese since I haven’t found very many references for that on the internet and our own Cantonese team on Transifex is currently without translators, so if anyone knows more about that I’d love to hear it.


chchwy:
粵语 is Cantonese, that's correct.

Issue #1248, opened 2019-08-20
Implement a discrete straight line constraint for drawing tools


Jose-Moreno:
### Issue Summary
As the name implies, having a discrete constraint (i.e variable degree from initial click placement) to create straight lines is a very widespread and useful feature that Pencil2D has been missing ever since it was created. 

Even the polyline tool could benefit from this behavior as creating straight lines at odd angles doesn't look really nice when you need precision drawing for creating certain shapes.

### Expected Results

To be able to create a straight line at regular or odd intervals depending on a modifier key (normally SHIFT). This would however have to be validated against the Quick-sizing behavior which already uses the hard coded SHIFT key to change the brush size when enabled

### Video or Image Reference

![image](https://user-images.githubusercontent.com/4912907/63309641-0df0c080-c2bd-11e9-9044-5349e2267d2f.png)

Red: odd angle straight lines (current)
Blue: discrete angle straight lines

davidlamhauge:
I don't know many of the programs that you (and others) refer to, and I don't understand 100% what you want here.
My guess is that you want a tool, or an extension to a tool, so it can draw straight lines. Right?
An these lines should be constrained to some predefined angles? Why? If I want a slightly tilted line (3-5 degrees) what then?
Well chances are that I've misunderstood it all, so please enlighten me :-)

Jose-Moreno:
@davidlamhauge Sure. So if you want a slightly tilted line you can do that already freehand or with the polyline tool. 

But If I want to draw a perfectly flat horizon line, delineate a set of buildings or mechanical elements, or better yet, to draw an isometric perspective grid at 45° angles, I can't without great effort or without using other software entirely.

So I'd like to alternate between drawing freehand lines and straight-angled lines however I choose for complex artwork. This is a simple tool meant to help in drawing inorganic, mechanical or repetitive patterns more efficiently, not to take away the artist freedom.

As you mentioned this is quite simply a request to extend all the drawing tools, and quite particularly to have the polyline tool to be able to lock the stroke placement on a specific axis or direction as determined by a custom angle.

Such behavior could also work similarly as it does in PhotoShop, perhaps just for the freehand drawings tools. In PS when you press & hold the modifier key, then click once on the canvas to establish the first point of the line, and afterwards click on a second point it creates a straight line in any direction as determined by those two points. Note that this will not provide necessarily provide perfect angles, but rather allow you to "eyeball" the angles yourself.

[...]

PR #1830, opened 2024-04-26
#1248 Horizontal and Vertical Line Snapping for Drawing Tools

ortolanmiles:
Added functionality for pen, pencil, brush and eraser to enter a snapping mode when drawing on a bitmap layer. Drawing movement will be locked to be either horizontal or vertical from the original click location. This covers issue number 1248 for all tools except the polyline tool. Preview of the line is generated, and stroke is only fully committed to the canvas after a pointer release event. RIght now the hotkey for this is '-'. 

I believe this change is ready to be merged into the project, but please let me know if there is anything that needs to be changed

https://github.com/pencil2d/pencil/assets/165717745/9122d51d-2e5e-4688-bf6c-ff56cb9ce9df


scribblemaniac:
Hello @ortolanmiles. I just wanted to thank you for your contribution. I know there has not been any activity on this the past month but rest assured, we will get to reviewing your contribution. We are just in the middle of [making a new release](https://github.com/pencil2d/pencil/issues/1831), which is always quite a process for us. We are also holding off on merging any new features like this until after the release to reduce the risk of introducing new bugs right before the release.

PR #1834, opened 2024-04-26
Fix crash when drawing on hidden layer

J5lx:
As reported at https://discuss.pencil2d.org/t/pencil2d-crashes-when-trying-to-drawing-to-a-hidden-layer/8038?u=j5lx


MrStevns (approved):
Fix confirmed and changes looks good to me 👍 
Merging.

PR #1836, opened 2024-04-28
Warn that translation changes require restart in new language

scribblemaniac:
When changing languages in the preferences window, this displays the warning in the target language instead of the current language so that people can actually read and understand it.


MrStevns (approved):
LGTM and nice improvement. 👍

PR #1840, opened 2024-04-30
Improve dock layout for lower resolution

MrStevns:
A follow up PR from #1824 

I've adjusted the margin as well as the min width for all the dock widgets so they are aligned.
In addition the font has also been adjusted from 13 pt to 11 pt.

The result looks like this:
image
Front: new layout
Back: old layout

The widgets that had the greatest impact adjusting this was the tool options widget as well as the onion skin widget.

Side by side:
| Old | New |
| ----| ----  |
| image | image |

Adjusting the font also helped quite a bit.
| 11 point font size | 13 point font size | 
| ----------------- | ----------------- | 
image | image
| 142 px | 162 px |

I've also added extra QLayout around certain elements, which is not a mistake. Doing so fixes some problems where Qt added extra margins around checkboxes and labels.


chchwy:
Would you be able to change the margins of the flowlayout in the toolbox as well?
It should be at `toolbox.cpp` line141 something like:

```
FlowLayout* flowlayout = new FlowLayout(0, 3, 3);
```

MrStevns:
I played around with that too, though it causes the widget to never go into one vertical line, at least on mac.

| 3 px margin | 0 px margin |
| ------------ | ------------ |
| image | image |

It bothers me a bit but I can't figure out how to force the layout into submission. For some reason it won't get smaller than 60 px or so... even though I believe i've tweaked all possible values. I must be missing something.

chchwy (approved):

MrStevns:
Small improvement. The toolbox can now also be shrunk down when laid out horizontally.
dock-widget-layout2

MrStevns:
Hmm not sure what's up but the font looks larger on some of the widgets on my windows machine.
it looks like setting the font point size specifically almost has the opposite effect on windows... sigh 😩 

https://forum.qt.io/topic/125993/font-point-size-weird-behaviour-on-windows I found this and some other threads mentioning similar problems... what a can of worms.

image

Sure we could just go back to the old font but seriously... Qt, you had one job.

MrStevns:
Fixed. The "solution" is to use qstylesheet instead. In this case i'm setting the font-size of all widgets that are children of BaseDockWidget.

image

MrStevns:
I have also tested these changes on a linux distro now, change still looks good to me.
image

MrStevns:
Thanks for testing it out. 👍 

> On Windows, it looks practically unchanged except for the title bars, whereas on Linux the text is painfully tiny for me, much smaller than the default font size (note the status bar for comparison):

I've reverted the font change again.
This seems to be one of the those things that require a greater understanding of Qt in order to change properly, so we'll keep the same font for now. I don't like the idea of scaling the font with some arbitrary number. It must be possible to use proper px or pt values and then it's a matter of figuring out how those work in Qt compared to web, if it's css they're basing their logic of.

> Also, on Qt 6, I now get a new warning in the console:

I'll fix that.

MrStevns:
> “Scaling the font with some arbitrary number” is pretty standard on the web and generally recommended for accessibility, I’m not sure what’s the problem with it. E.g. the main font size on our forum is 0.938em, and I haven’t changed it. If you have something else in mind, sure. But please consider taking the system default as a base in one way or another since we mainly use native styling.

Alright fair enough, I'm used to mobile development where it's not like this but if that's how it's done, then we'll do it like that.
I've reverted the font changes now though, instead I think it would be wiser to create a font size preference, so the user can control this themselves. That's out of scope for this PR though.

> And lastly, your most recent commit seems to have broken the colour inspector preview somehow, it now defaults to the standard widget background on startup. Not sure what exactly is going on there…

This was caused by a style override which has been reverted.

Would you mind reviewing again @J5lx

MrStevns:
Hmm strange, it's not something i'm seeing in my build anymore. 
I'm using Qt 6.7.0, what version are you on?

Also, is this with the default layout or have you moved the docks around?

MrStevns:
Ah... good point regarding `mDockArea` being uninitialized, I forget that it was an enum 😅 

I've made a few changes, though i'm unsure about whether we should simply remove the -1 in BaseDockWidget rather than me simply not calling the base class. What do you think?

MrStevns:
Thanks Jakob! 🙏

PR #1842, opened 2024-05-23
Improve dev docs styling

J5lx:
I gave the dev docs a little makeover to bring it more in line with the rest of the site, prompted by this little conversation on Discord:

> **scribble:** I think reviewing the Contributing page is also in order. […]
> **me:** I’ve thought about that as well. I’ve also been wondering whether it might be a good idea to merge it into the https://dev.pencil2d.org/ front page and then use that as the new contribute page, so that there’s a single place for all the contribution docs.
> **scribble:** Seems like a good idea. I just wish our docs didn't look so ugly. Not exactly putting our best foot forward for potential new contributors.

Although the diff is a bit on the larger side, a fair portion of it is simply boilerplate to nullify the default styles.

Before:

image

After:

image


scribblemaniac:
Already looks much more cohesive now. Thanks!

PR #1843, opened 2024-05-24
Prevent Pencil2D projects from referencing external files

scribblemaniac:
This PR adds additional checks to the loading of bitmap/vector/sound layers so that they will not be loaded if they are outside of the data directory. This can occur with specially crafted main.xml files, or with symlinks. This closes multiple admittedly unlikely security threats in the program that can result in exfiltrating images from arbitrary locations though animations, and even duplicating arbitrary files under some circumstances. Tests have also been written for the scenarios this PR covers with the exception of symlinks (and possibly Windows shortcuts) because Qt has very bad support for symlink creation.

This also incidentally fixes an issue where an infinite loop will occur when loading a main.xml file that contains a non-element node (ex. an xml comment) in the sound layer.

### DIFF
--- a/core_lib/src/structure/layerbitmap.cpp
+++ b/core_lib/src/structure/layerbitmap.cpp
@@ -21,7 +21,7 @@ GNU General Public License for more details.
 #include <QFile>
 #include "keyframe.h"
 #include "bitmapimage.h"
-
+#include "util/util.h"
 
 LayerBitmap::LayerBitmap(int id) : Layer(id, Layer::BITMAP)
 {
@@ -204,31 +204,9 @@ void LayerBitmap::loadDomElement(const QDomElement& element, QString dataDirPath
         QDomElement imageElement = imageTag.toElement();
         if (!imageElement.isNull() && imageElement.tagName() == "image")
         {
-            bool shouldLoad = true;
-
-            // Make sure src path is relative
-            shouldLoad &= QFileInfo(imageElement.attribute("src")).isRelative();
-            QFileInfo fi(dataDirPath, imageElement.attribute("src"));
-            // Make sure file is in data directory after resolving relative components and symlinks
-            QString canonicalPath = fi.canonicalFilePath();
-            fi.setFile(!canonicalPath.isEmpty() ? canonicalPath : fi.absoluteFilePath());
-            QDir dataDir(dataDirPath);
-            QDir ancestor = fi.dir();
-            while (ancestor != dataDir)
+            QString path = validateDataPath(imageElement.attribute("src"), dataDirPath);
+            if (!path.isEmpty())
             {
-                QDir newAncestor = QFileInfo(ancestor.absolutePath()).dir();
-                if (ancestor == newAncestor)
-                {
-                    // Data dir was not found in ancestors of the src path
-                    shouldLoad = false;
-                    break;
-                }
-                ancestor = newAncestor;
-            }
-
-            if (shouldLoad)
-            {
-                QString path = fi.absoluteFilePath();
                 int position = imageElement.attribute("frame").toInt();
                 int x = imageElement.attribute("topLeftX").toInt();
                 int y = imageElement.attribute("topLeftY").toInt();
--- a/core_lib/src/structure/layerbitmap.cpp
+++ b/core_lib/src/structure/layerbitmap.cpp
@@ -202,24 +202,41 @@ void LayerBitmap::loadDomElement(const QDomElement& element, QString dataDirPath
     while (!imageTag.isNull())
     {
         QDomElement imageElement = imageTag.toElement();
-        if (!imageElement.isNull())
+        if (!imageElement.isNull() && imageElement.tagName() == "image")
         {
-            if (imageElement.tagName() == "image")
+            bool shouldLoad = true;
+
+            // Make sure src path is relative
+            shouldLoad &= QFileInfo(imageElement.attribute("src")).isRelative();
+            QFileInfo fi(dataDirPath, imageElement.attribute("src"));
+            // Make sure file is in data directory after resolving relative components and symlinks
+            QString canonicalPath = fi.canonicalFilePath();
+            fi = !canonicalPath.isEmpty() ? canonicalPath : fi.absoluteFilePath();
+            QDir dataDir(dataDirPath);
+            QDir ancestor = fi.dir();
+            while (ancestor != dataDir)
             {
-                QString path = dataDirPath + "/" + imageElement.attribute("src"); // the file is supposed to be in the data directory
-                QFileInfo fi(path);
-                if (!fi.exists()) path = imageElement.attribute("src");
+                QDir newAncestor = QFileInfo(ancestor.absolutePath()).dir();
+                if (ancestor == newAncestor)
+                {
+                    // Data dir was not found in ancestors of the src path
+                    shouldLoad = false;
+                    break;
+                }
+                ancestor = newAncestor;
+            }
+
+            if (shouldLoad)
+            {
+                QString path = fi.absoluteFilePath();
                 int position = imageElement.attribute("frame").toInt();
                 int x = imageElement.attribute("topLeftX").toInt();
                 int y = imageElement.attribute("topLeftY").toInt();
-                qreal opacity = 1.0;
-                if (imageElement.hasAttribute("opacity")) {
-                    opacity = imageElement.attribute("opacity").toDouble();
-                }
+                qreal opacity = imageElement.attribute("opacity", "1.0").toDouble();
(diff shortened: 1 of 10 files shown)


core_lib/src/structure/layerbitmap.cpp
```diff
@@ -202,24 +202,41 @@ void LayerBitmap::loadDomElement(const QDomElement& element, QString dataDirPath
...
+
+            // Make sure src path is relative
+            shouldLoad &= QFileInfo(imageElement.attribute("src")).isRelative();
+            QFileInfo fi(dataDirPath, imageElement.attribute("src"));
+            // Make sure file is in data directory after resolving relative components and symlinks
+            QString canonicalPath = fi.canonicalFilePath();
+            fi.setFile(!canonicalPath.isEmpty() ? canonicalPath : fi.absoluteFilePath());
+            QDir dataDir(dataDirPath);
+            QDir ancestor = fi.dir();
+            while (ancestor != dataDir)
```
MrStevns, line 217: Couldn't we move this to a method in Layer, so we don't have to have it duplicated three times?

core_lib/src/structure/layervector.cpp
```diff
@@ -155,31 +155,53 @@ void LayerVector::loadDomElement(const QDomElement& element, QString dataDirPath
...
-                    getVectorImageAtFrame(position)->setOpacity(1.0);
-                progressStep();
             }
+            else
+            {
+                position = imageElement.attribute("frame").toInt();
+                addNewKeyFrameAt(position);
+                getVectorImageAtFrame(position)->loadDomElement(imageElement);
+                getVectorImageAtFrame(position)->setOpacity(imageElement.attribute("opacity", "1.0").toDouble());
+            }
```
MrStevns, line 200: I know that we had similar code but this feels wrong to me, why are we instantiating a keyframe if there's no src for it? Or let me rephrase, why do we create a new keyframe here at all when this method should only be about loading data if we have the necessary data.

It's not critique as much as a question, none of our other Layer classes have this so it seems weird to me that the Vector class has it, or do you disagree?

scribblemaniac: It is weird to me that they are handled differently between layers yes. I did not change this behavior because it's harmless and I wanted to minimize the changes I make so it can be merged quicker and with lower risk of introducing a bug in this critical file loading code.

I don't think it is wrong to instantiate an empty frame if there is no src. If there is a bitmap key with no data, it does not save an image for that frame, but it still sets the src. For a vector key a .vec file is written and src is set even if there is no data. An src attribute pointing to a file that does not exist and a file that contains no data both feel wrong to me. If there's no data that needs to be stored for a key, we shouldn't need to create a data file. If there's no data file, we should not need to specify the path to the non-existent file. But users still may want to save an emtpy key (to stop holding the previous key, or for planning/timing purposes).

core_lib/src/structure/layerbitmap.cpp
scribblemaniac: Yes very good point. I have refactored this into a separate function. It is in util because this is not exclusively used by layers now.

MrStevns (commented):
I was about to approve this as it looks good to me now but i noticed that you haven't made this change for the camera layer, is that intentional?

scribblemaniac:
> I was about to approve this as it looks good to me now but i noticed that you haven't made this change for the camera layer, is that intentional?

Yes this is intentional. As far as I know, the camera layer doesn't read any data from other files (all the camera data is the in the main.xml file) so it does not need these checks.

MrStevns (approved):
LGTM 👍

core_lib/src/util/util.cpp
```diff
@@ -126,3 +128,59 @@ QString uniqueString(int len)
...
+    // Make sure src path is relative
+    if (!QFileInfo(filePath).isRelative()) return QString();
+
+    QFileInfo fi(dataDirPath, filePath);
+    // Recursively resolve symlinks
+    QString canonicalPath = fi.canonicalFilePath();
+
+    // Iterate over parent directories of the file path to see if one of them equals the data directory
+    QDir dataDir(dataDirPath);
+    if (dataDir.exists()) dataDir.setPath(dataDir.canonicalPath());
```
chchwy, line 143: I don't really get what this line does? is it for resolving the symlinks?
Also could you add curly brackets?

core_lib/src/util/util.cpp
```diff
@@ -126,3 +128,59 @@ QString uniqueString(int len)
...
+    if (canonicalPath.isEmpty())
+    {
+        // File does not exist, use absolute path and attempt to resolve symlinks again for each parent directory
+        fi.setFile(fi.absoluteFilePath());
+        QDir ancestor(fi.absoluteFilePath());
+        while (true) {
```
chchwy, line 149: I don't quite understand this while loop. Could you please elaborate a bit more? probably with some examples?

core_lib/src/util/util.cpp
scribblemaniac: Yes. The canonical path is effectively the absolute path with all symlinks and relative components ('.' and '..') resolved. This particular line is necessary to deal with the case when one or more of the directories in the data dir's path is a symlink. For example, imagine if the data dir for a project is in `/tmp/Pencil2D/Project_Y2xD_012345/data`, and `/tmp` is actually a symlink to `/custom_temp_dir`. In this case, a vector file defined in main.xml could be `002.001.vec`, which when that file is fully resolved below, it will result in a path of `/custom_temp_dir/Pencil2D/Project_Y2xD_012345/data/002.001.vec`. This will not match the data dir value of `/tmp/Pencil2D/Project_Y2xD_012345/data` unless the symlinks are also resolved for the data dir path.

core_lib/src/util/util.cpp
scribblemaniac: For an example, let's say the data directory is at the path `/a/b/c` and the file path being validated is at `/a/b_symlink/c/d/file.txt`. The directory `b_symlink` is obviously a symlink to `/a/b`. We will also assume the directory `/a/b/c/d` and `file.txt` do not exist and directories `/a`, `/a/b`, and `/a/b/c` do exist. I will be walking through the logic based on the slightly rearranged code I have pushed since your review so pull the latest code before following along. 

This loop iterates up the directory hierarchy of the file path. Keep in mind that it only runs if the canonical path of the file is empty, which means the file and possibly some of its parent directories don't exist or are dangling symlinks (symlinks where the target location does not exist, QDir::exists() also returns false in this case).

Iteration 1:
1. ancestor is `/a/b_symlink/c/d/file.txt`
2. ancestor is not equal to data dir
3. newAncestor is set to the parent of ancestor: `/a/b_symlink/c/d`
4. newAncestor does not exist
5. newAncestor is different from ancestor
6. Set ancestor to newAncestor

Iteration 2:
1. ancestor is `/a/b_symlink/c/d`
2. ancestor is not equal to data dir
3. newAncestor is set to the parent of ancestor: `/a/b_symlink/c`
4. newAncestor does exist, so set new ancestor to the canonical path, which resolves the symlink b_symlink. newAncestor is now: `/a/b/c`
5. newAncestor is different from ancestor
6. Set ancestor to newAncestor

Iteration 3:
1. ancestor is `/a/b/c`
2. ancestor is equal to data dir, that means the file would be within the data directory after resolving all symlinks  if it and all of its parent directories existed. As a positive result, return the file path.

If the data dir was not `/a/b` and instead something like `/x`, then the loop would continue with ancestor = `/a/b` and then `/a` and then `/`. When ancestor is the root directory, calling dir on it will again return the root directory `/` and newAncestor will equal ancestor. As a negative result, return an invalid string because we now know that the file would not be within the data directory after resolving symlinks if it and all of its parent directories existed.

core_lib/src/util/util.cpp
chchwy: good thank you for explaining this.

core_lib/src/util/util.cpp
chchwy: Alright, thanks for the thorough explanation.

I now understand the nuance here. The condition `if (ancestor == newAncestor)` indicates that the path has reached the root directory, as the parent directory of the root is itself. This wasn’t immediately obvious to me, so I couldn't grasp it at first.

Would you be able to put this explanation into the code as a comment or a util function so the code can express its intent clearly? can `QDir::isRoot()` be used here?

chchwy (approved):

scribblemaniac:
Thanks for the reviews @MrStevns and @chchwy

chchwy:
All good. BTW, please use "Squash and Merge" next time. It gives a proper granularity of commits, making it clearer when reviewing the git history.

PR #1851, opened 2024-06-24
Do not make a new keyframe if double clicking on an existing keyframe

khoidauminh:
Based on the message from the Software's discord server, quoted:

> Not sure if it's been discussed here but:
First, double clicking in anywhere in the timeline gives you a new keyframe. Double clicking on a keyframe appends one after it.
When you need to move a keyframe, you need to click on it first before you can drag it around. If I happen to do this too fast, pencil2d registers this as a double click and does the keyframe appending. This can be annoying 

> I think it'll help a lot if pencil2d distinguishes between double clicking on an empty space and on a keyframe

This PR does exactly what the message above wants to. When double clicking on an existing keyframe, the program won't make a new keyframe, instead only selects it. Beneficial for those who want to drag the keyframe around in a quick way.

### DIFF
--- a/app/src/timelinecells.cpp
+++ b/app/src/timelinecells.cpp
@@ -1130,7 +1130,10 @@ void TimeLineCells::mouseDoubleClickEvent(QMouseEvent* event)
         if (mType == TIMELINE_CELL_TYPE::Tracks && (layerNumber != -1) && (frameNumber > 0) && layerNumber < mEditor->object()->getLayerCount())
         {
             mEditor->scrubTo(frameNumber);
-            emit insertNewKeyFrame();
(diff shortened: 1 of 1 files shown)


MrStevns:
Hi khoidauminh, thanks for the PR. 

I implemented this feature with the goal that it should be easier to add addition frames to a Keyframe without having to move the existing ones on the timeline. From the thread though I realize that in its current state, the functionality is too easy to trigger.

Based on the feedback from Discord, I'll accept the PR for now.

MrStevns (approved):

MrStevns:
A made a small change to account for soundclip length and also put the scrub call inside the branch.
Will merge PR now.

PR #1853, opened 2024-06-28
Temporarily modify width and feather during Quick Sizing

khoidauminh:
During Quick Sizing, Pencil2d tries to write changes to disk every time the width/feather changes. Some devices, particularly drawing tablets, send events at a high rate, causing Pencil2d to fail to keep up with with the events and hang for a moment.

This PR temporarily fixes the issue by introducing new methods that only changes the properties, then calling the original methods (which includes writing to disk) once we're done doing Quick Sizing.

### DIFF
--- a/core_lib/src/managers/toolmanager.cpp
+++ b/core_lib/src/managers/toolmanager.cpp
@@ -153,7 +153,6 @@ void ToolManager::setWidth(float newWidth)
     }
 
     currentTool()->setWidth(static_cast<qreal>(newWidth));
-    emit penWidthValueChanged(newWidth);
     emit toolPropertyChanged(currentTool()->type(), WIDTH);
 }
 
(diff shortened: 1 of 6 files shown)


MrStevns:
Hi khoidauminh 

Thanks for the contribution.

While the idea that we don't write to disk on every change is fine, I think the name of the function is misleading as it still changes the width property. I'd prefer that we didn't come up with a "temporary" bandaid solution but rather figure out how we can fix this properly. 

As such, would you mind trying something?

I noticed that we call QSettings::sync() explicitly in all the tools after updating a setting, which might be overkill. Further more, this could also be a what's causing the needless amount of writes because sync will force an update every time it's called, where as normally setValue will simply add a update to the event queue, and apply it sometime later.

What happens if you remove the sync call, does the performance problem persist?
For example in PencilTool::setWidth, line: 77.

scribblemaniac:
@MrStevns Based on the Qt [documentation](https://doc.qt.io/qt-5/qsettings.html#sync) for `QSettings::sync`, it gets called by the `QSettings` destructor., although the documentation for the destructor itself is less clear. While calling it explicitly might be overkill, I suspect it won't fix the performance issue here.

While a "proper" fix might address the possibility of this being an issue for all settings with some debouncing, I don't think this is a bandaid solution. It makes some sense to me that the width/feathering would not be applied by quick sizing until you release the cursor. If the property value is not set at all though, then you won't be able to see live updates in the options panel.

khoidauminh:
@MrStevns I've tested this and the main bottleneck is actually the QSettings::setValue() function. I think it tries to scan through the keys and set the value, which is very expensive.

MrStevns (approved):
Thanks for taking the initiative to fix this. 

I've made some changes to simplify the logic.
Since the UI only had to be notified to update the slider, you could call ToolPropertyChanged from StrokeTool directly which avoids exposing the tempWidth logic to the manager.

MrStevns (approved):

PR #1856, opened 2024-06-29
Improve preferences layout for long translations

### LINKED ISSUE #1850: app crashes when opening settings
### Issue Summary
Pencil2d crashes when trying to open settings menu 
 

### Actual Results
-
 

### Expected Results
-
 

### Steps to reproduce
Start the program --> Edit --> Settings  -->...

 

### System Information
+ Pencil2D Version: 0.7.0.911

 

+ Operating System: Fedora Linux 40 , KDE Plasma 6.1 (Wayland)

 

+ RAM Size: 16Gb

 

+ Graphics Tablet: HUION GT220 v2

J5lx:
Fixes #1850 by simply removing the size restriction altogether. I couldn’t think of a good reason to keep it. I also made the dialog a little bigger by default and made some changes to the “tab bar” to avoid cutting off long translations.

Before:

image

After:

image

### DIFF
--- a/app/src/preferencesdialog.cpp
+++ b/app/src/preferencesdialog.cpp
@@ -23,6 +23,13 @@ PreferencesDialog::PreferencesDialog(QWidget* parent) :
     ui(new Ui::PreferencesDialog)
 {
     ui->setupUi(this);
+
+    for (int i = 0; i < 0 + 1 * ui->contentsWidget->count(); i++) {
+        QListWidgetItem* item = ui->contentsWidget->item(i);
+        // Fill entire width
+        item->setSizeHint({std::numeric_limits<int>::max(),
+                           ui->contentsWidget->visualItemRect(item).height()});
+    }
 }
 
 PreferencesDialog::~PreferencesDialog()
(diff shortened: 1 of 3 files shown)


MrStevns (changes requested):
The removal of the size constraint on the dialog is fine by me, not sure why we had it either.

The code you added to improve on the tab view placing completely broke it for me though. 

loooong

MrStevns:
> That’s odd, I’m definitely not seeing that on Linux. On Windows, Qt 5 has that issue but not Qt 6, maybe that’s the issue? I’ll see if I can figure out what’s going on there.

Ah good point, I was testing with Qt 5 but even with my Qt 6 build the issue remains. I've tested with 6.6.2 and 6.7.0.

MrStevns (approved):
That did the trick, both qt 5 and qt 6 works now, as such I think we can approve of these changes.

MrStevns:
btw. shouldn't the PR point to the release branch or are we simply not applying more bug fixes to the release?

chchwy:
> btw. shouldn't the PR point to the release branch or are we simply not applying more bug fixes to the release?

Target the master branch since the change was made based on the master branch. And then I will cherry-pick this to the release branch later on.

chchwy (approved):

PR #1863, opened 2024-07-21
Closed Polyline path

HeCorr:
This PR adds the `Closed` option to the Polyline tool which automatically closes it's path as it's being drawn.

The `Ctrl` key can be held to temporarily override the closed state:
- If the `Closed` option is _on_, holding `Ctrl` temporarily _disables_ it
- If the `Closed` option is _off_, holding `Ctrl` temporarily _enables_ it

### Video

https://github.com/user-attachments/assets/fb6ad3e7-4c21-4311-acb2-80e9689dc216

It works with both double-clicking and pressing Enter to apply the Polyline.


HeCorr:
This has been discussed on Discord and I decided to add a tool option for this which inverts the behavior of the `Ctrl` key, so that:

- If the `Closed` option is _on_, holding `Ctrl` temporarily _disables_ it
- If the `Closed` option is _off_, holding `Ctrl` temporarily _enables_ it

So that users who prefer to use the `Ctrl` key can leave the option off and those who don't can leave the option on with the added benefit of temporarily disabling it if necessary without having to toggle the tool option.

core_lib/src/tool/basetool.cpp
```diff
@@ -148,6 +149,11 @@ void BaseTool::setBezier(const bool _bezier_state)
     properties.bezier_state = _bezier_state;
 }
 
+void BaseTool::setClosed(const bool _closed)
+{
+    properties.closedPolylinePath = _closed;
```
MrStevns, line 154:
```suggestion
void BaseTool::setClosed(const bool closed)
{
    properties.closedPolylinePath = closed;
```

There's no need for underscore here

core_lib/src/tool/basetool.h
```diff
@@ -109,6 +110,7 @@ class BaseTool : public QObject
     virtual void setFeather(const qreal feather);
     virtual void setInvisibility(const bool invisibility);
     virtual void setBezier(const bool bezier_state);
+    virtual void setClosed(const bool closed);
```
MrStevns, line 113: Maybe we could call this setClosedPath instead so it's more consistent with the other places, that would also make the name less ambiguous.

core_lib/src/tool/polylinetool.h
```diff
@@ -51,6 +52,7 @@ class PolylineTool : public StrokeTool
 
 private:
     QList<QPointF> mPoints;
+    bool mClosedOverride = false;
```
MrStevns, line 55: Consider using enabled or on here to make it more explicit that we're dealing with a bool.

i'd also appreciate if we could name it something like *mClosedPathEnabled* instead, to make it less ambiguous.

MrStevns, line 55: I'd prefer if we could make the naming similar to the calls, like "mClosedPathOverrideEnabled"

app/ui/tooloptions.ui
```diff
@@ -187,6 +187,16 @@
          </property>
         </widget>
        </item>
+       <item>
+        <widget class="QCheckBox" name="useClosedBox">
+         <property name="toolTip">
+          <string>Close Polyline path (hold Ctrl to temporarily invert)</string>
+         </property>
+         <property name="text">
+          <string>Closed</string>
```
MrStevns, line 196: I think that closed might be a tad too vague, maybe we could write "Closed path" instead

core_lib/src/tool/basetool.cpp
HeCorr: I had simply copied the code from right above 😛
https://github.com/pencil2d/pencil/blob/327ccbc6be4a8a6244f9a2d793ec545ac6b58fb9/core_lib/src/tool/basetool.cpp#L147-L150

HeCorr:
The requested changes have been committed!

app/ui/tooloptions.ui
```diff
@@ -187,6 +187,16 @@
          </property>
         </widget>
        </item>
+       <item>
+        <widget class="QCheckBox" name="useClosedBox">
+         <property name="toolTip">
+          <string>Close Polyline path (hold Ctrl to temporarily invert)</string>
```
MrStevns, line 193: Although I now that we do this elsewhere in the file, referring to ctrl is actually not correct, because Control is substituted by ⌘ or Cmd on macOS.

I considered changing this in your PR but since we do it in various places, we'll keep it as is.

The correct way to fix this though would be something like:
```
QString controlShortcutText = QKeySequence(Qt::Key_Control).toString(QKeySequence::NativeText);
ui->useClosedBox->setToolTip(tr("Close Polyline path (hold %1 to temporarily invert)").arg(shortcutText));
```

MrStevns (approved):
I took the PR for another spin and noticed a few other things I forgot to check the last time, sorry about that.

Things that wasn't working that I've fixed for you:
* The check state would not be saved to our settings file, meaning `closedPolylinePath` would always return false.
* The checkbox would not be updated on startup.

I also took the liberty to rename useClosedBox to useClosedPathBox.

With those things taken care of, this PR should imo. be good to go.
Feel free to merge the PR if you're satisfied with those changes.

HeCorr:
> The check state would not be saved to our settings file

I did notice the state wasn't being saved, but then I tested the Bézier option and it didn't save either hahah

Thanks for the fixes though!

MrStevns:
Yeah it's not consistent... ideally all tool option properties should be saved to settings.

btw. are you allowed to merge the PR yourself?

HeCorr:
> are you allowed to merge the PR yourself?

nope 
Screenshot_2024-07-24-15-15-40-988

PR #1867, opened 2024-07-24
Maintain active layer track in view

### LINKED ISSUE #1858: Layer list does not scroll when using arrow keys
### Issue Summary
When selecting different layers with the Up and Down arrow keys, the list doesn't scroll automatically, forcing the user to scroll it manually with the mouse wheel to see which layer is currently active.
 

### Actual Results
The timeline (specifically the layer list) stays in place:

https://github.com/user-attachments/assets/6d0b6289-61ab-4633-9071-6e32b6899fb7

 

### Expected Results
I expected the timeline to scroll vertically automatically, so the active layer stays visible:
(manually scrolled for the sake of illustration)

https://github.com/user-attachments/assets/59d00ce0-11ea-4727-a111-1ea68032c2d2

 

### Steps to reproduce
1. Open Pencil2D
2. Create multiple layers
3. Change the active layer by pressing the Up and Down arrow keys
 

### System Information
```
Nightly Build Jun 17 2024

commit: b28357852d6b4467c7e7cbca4179e166d62d7c9b
date: 2024-06-17_18:35:38

Operating System: Windows 10 Version 1909
CPU Architecture: x86_64
Qt Version: 5.15.2
RAM Size: 64 GB
```

HeCorr:
Fixes #1858.

This change maintains the currently selected layer visible when the active layer changes, so that using the Up and Down arrow keys works as expected:

https://github.com/user-attachments/assets/26c97fea-6aa4-4d3b-a56c-f391d74162f4

It also works when the layer is far away from view, not just right above/below:

https://github.com/user-attachments/assets/8183b8cb-e94b-4c7d-8847-d7facfa76b48

## Caveat
When clicking on a layer that is barely in view (which isn't considered as in view by the code), it is moved due to the sudden scroll as the cursor is still being held:

https://github.com/user-attachments/assets/2732bebc-9a8e-454e-aab0-612255185fbd

I thought of making it so that the layer is only selected upon mouse _release_ but I'm willing to bet the code responsible for reordering layers decides which one to move based on which one is active, but I could be wrong.

Another potential solution would be to snap the timeline height so that it never shows half a layer:

### DIFF
--- a/app/src/timeline.cpp
+++ b/app/src/timeline.cpp
@@ -193,6 +193,8 @@ void TimeLine::initUI()
 
     setWindowFlags(Qt::WindowStaysOnTopHint);
 
+    connect(editor()->layers(), &LayerManager::currentLayerChanged, this, &TimeLine::currentLayerChanged);
+
     connect(mHScrollbar, &QScrollBar::valueChanged, mTracks, &TimeLineCells::hScrollChange);
     connect(mTracks, &TimeLineCells::offsetChanged, mHScrollbar, &QScrollBar::setValue);
     connect(mVScrollbar, &QScrollBar::valueChanged, mTracks, &TimeLineCells::vScrollChange);
@@ -282,6 +284,25 @@ void TimeLine::wheelEvent(QWheelEvent* event)
     else
     {
         mVScrollbar->event(event);
+    }
+}
+
+void TimeLine::currentLayerChanged(int layerIndex)
+{
+    // invert index so 0 is at the top
+    int idx = mNumLayers - layerIndex - 1;
+    // number of visible layers
+    int height = mNumLayers - mVScrollbar->maximum();
+    // scroll bar position/offset
+    int pos = mVScrollbar->value();
+
+    if (idx < pos) // above visible area
+    {
+        mVScrollbar->setValue(idx);
+    }
+    else if (idx >= pos + height) // below visible area
+    {
+        mVScrollbar->setValue(idx - height + 1);
     }
 }
 
--- a/app/src/timeline.cpp
+++ b/app/src/timeline.cpp
@@ -197,12 +197,11 @@ void TimeLine::initUI()
 
     setWindowFlags(Qt::WindowStaysOnTopHint);
 
-    connect(editor()->layers(), &LayerManager::currentLayerChanged, this, &TimeLine::currentLayerChanged);
     connect(mHScrollbar, &QScrollBar::valueChanged, mTracks, &TimeLineCells::hScrollChange);
     connect(mTracks, &TimeLineCells::offsetChanged, mHScrollbar, &QScrollBar::setValue);
     connect(mVScrollbar, &QScrollBar::valueChanged, mTracks, &TimeLineCells::vScrollChange);
     connect(mVScrollbar, &QScrollBar::valueChanged, mLayerList, &TimeLineCells::vScrollChange);
-    connect(mVScrollbar, &QScrollBar::valueChanged, this, &TimeLine::scrollbarValueChanged);
+    connect(mVScrollbar, &QScrollBar::valueChanged, this, &TimeLine::onScrollbarValueChanged);
     connect(mScrollingStoppedTimer, &QTimer::timeout, mLayerList, &TimeLineCells::onScrollingVerticallyStopped);
 
     connect(splitter, &QSplitter::splitterMoved, this, &TimeLine::updateLength);
@@ -239,7 +238,7 @@ void TimeLine::initUI()
 
     LayerManager* layer = editor()->layers();
     connect(layer, &LayerManager::layerCountChanged, this, &TimeLine::updateLayerNumber);
-    connect(layer, &LayerManager::currentLayerChanged, this, &TimeLine::onLayerChanged);
+    connect(layer, &LayerManager::currentLayerChanged, this, &TimeLine::onCurrentLayerChanged);
     mNumLayers = layer->count();
 
     scrubbing = false;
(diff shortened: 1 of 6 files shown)


MrStevns (approved):
Thanks for the contribution, this is an excellent improvement to the timeline.

Changes looks good to me, and it works as expected for me.
Good job 👍 

Feel free to merge if you're happy with the changes.

MrStevns (changes requested):
My bad, I missed the quite important caveat, which I can confirm happens. We need to take care of that.
Unfortunately the timeline cell logic is such a mess, it might not be trivial to fix without introducing some other bug.

You don't need to click on a half hidden layer to make it happen either, if you add 5 layers and click on layer 2, then it also happens.

HeCorr:
Maybe instead of listening to `currentLayerChanged` which is emitted with both arrows and cursor clicks, I could emit a new signal from `gotoPreviouslayer` and `gotoNextlayer`, so that this only runs when using the arrows.

Or a simpler way, which I couldn't figure out is to directly listen to arrow key presses from `timeline.cpp`.

HeCorr:
@MrStevns is the signature change of `currentLayerChanged` suggested above acceptable? If so I'm gonna give that a go soon.

MrStevns:
> Maybe instead of listening to currentLayerChanged which is emitted with both arrows and cursor clicks, I could emit a new signal from gotoPreviouslayer and gotoNextlayer, so that this only runs when using the arrows.

> Or a simpler way, which I couldn't figure out is to directly listen to arrow key presses from timeline.cpp.

> Or maybe even change the currentLayerChanged signal itself, adding an enum which specifies the source of the change (mouse or keyboard).

I'm not convinced that either of these will fix the problem. Why should changing the signature of `currentLayerChanged` change the behavior, when what's causing the problem is most likely the logic in timelinecells or the TimeLine::updateLayerView which controls the maximum amount of steps of the scrollbar?

HeCorr:
> Why should changing the signature of currentLayerChanged change the behavior?

Consider the following:
```c++
void TimeLine::currentLayerChanged(int layerIndex, LayerChangeSource source)
```
I can check whether `source == LayerChangeSource::ArrowKey` and only scroll if it's true, which would ignore mouse clicks (...unless clicking a half-visible layer already scrolls by default/outside this branch, which I honestly haven't tested).

> For this to work properly, I think you'll have to add some logic that prevents it from doing the layer swapping on mouse release when you're not explicitly dragging.

I'll have to see what I can do about that.

HeCorr:
Just checked on Nightly and indeed it does not scroll by default when clicking, which means my theoretical fix could work.

https://github.com/user-attachments/assets/8915e0ad-edc3-4cf1-a90a-a72ded4d6301

HeCorr:
I did a quick test and it did work, but I ended up changing too many places that use the `currentLayerChanged` signal, so I'm going to try to create a new signal instead.

https://github.com/user-attachments/assets/7487f98b-f749-4184-bc0f-5a959252810e

MrStevns:
> Consider the following:

> void TimeLine::currentLayerChanged(int layerIndex, LayerChangeSource source)

> I can check whether source == LayerChangeSource::ArrowKey and only scroll if it's true, which would ignore mouse clicks (...unless clicking a half-visible layer already scrolls by default/outside this branch, which I honestly haven't tested).

The layer manager should not have to be aware of what is triggering it, be it keyboard or mouse, that would be a GUI thing, meant for a the app side, for example the timeline widget or cells widget. To make such a change to fix what is essentially a GUI bug, would be wrong.

It's a good thing that you're experimenting and that you've found a few solutions that works. However, I still think we should deal with this through the interface code, rather than adding an additional emitter logic to the LayerManager, when the source of the problem comes from the timeline logic.

HeCorr:
> I can understand if you're reluctant to add more code to the timeline cell logic, as it's a mess

I wouldn't even know where to begin.

I understand that this simple fix is not the ideal solution but it works and unfortunately this is all I can provide at the moment.

MrStevns:
> > I can understand if you're reluctant to add more code to the timeline cell logic, as it's a mess
> 
> I wouldn't even know where to begin.
> 
> I understand that this simple fix is not the ideal solution but it works and unfortunately this is all I can provide at the moment.

No worries, let me help you then 😄 


> > I can understand if you're reluctant to add more code to the timeline cell logic, as it's a mess
> 
> I wouldn't even know where to begin.
> 
> I understand that this simple fix is not the ideal solution but it works and unfortunately this is all I can provide at the moment.

No worries, I took a stab at it which required minimal amount of changes to TimeLineCells.
You can check out the commits yourself but the gist of it is that we've introduced a `mScrollingVertically` so we can avoid moving the layers while actively scrolling. Once that's done, we still have one problem to solve though, which is how do we reset the value?

I figured that a timer that triggers after x milliseconds would do the trick. It will only trigger once after we've scrolled one way or another.

After inspecting the TimeLine class myself I also noticed that we had a ´layerChanged` slot already, so I've refactored your logic into that method instead of having an additional signal.

Let me know what you think.

HeCorr:
I personally don't understand why you would opt for a more complex implementation that uses a timer over simply emitting two signals and picking one to listen to depending on what makes the most sense.

Sure, it wasn't an elegant solution, but it was simple and worked pretty well. But that's just my opinion.

MrStevns:
I get your point but what might seem like a simpler solution at first, doesn't make sense in the long run.

Your solution adds seemingly abysmal complexity to the core layer manager, an object that is tied to everything, to fix a specific scenario in a specific GUI component, while I add logic to the timeline, which only affects that specific component and lives in the APP part of the code base.

Consider this example. Another person writes their own app client for Pencil2D; possibly for mobile, and write a better timeline implementation. Why should they have to deal with this new signal, when it's a problem in our GUI code?

Small things like this add up over time and makes the code base less maintainable. Even though we're open source, doesn't mean we don't want the code to be as good as it can be.

It's not a rant, truthfully, I'm just trying to voice why I think it's important to fix bugs where they appear. :)

MrStevns (approved):

PR #1870, opened 2024-07-26
Fixed a few warnings

HeCorr:
This PR fixes a handful of simple warnings, mostly missing `emit` keywords and unused QStrings.

There were a lot more warnings but unfortunately my Qt and C++ knowledge isn't good enough to handle those so I decided not to risk breaking something. At least I fixed more than just the missing `emit`s.

### DIFF
--- a/app/src/importimageseqdialog.cpp
+++ b/app/src/importimageseqdialog.cpp
@@ -272,7 +272,7 @@ const PredefinedKeySetParams ImportImageSeqDialog::predefinedKeySetParams() cons
     dot = finalList[0].lastIndexOf(".");
 
     QStringList absolutePaths;
-    for (const QString& fileName : qAsConst(finalList)) {
+    for (const QString& fileName : finalList) {
         absolutePaths << path + fileName;
     }
 
--- a/app/src/importimageseqdialog.cpp
+++ b/app/src/importimageseqdialog.cpp
@@ -272,7 +272,7 @@ const PredefinedKeySetParams ImportImageSeqDialog::predefinedKeySetParams() cons
     dot = finalList[0].lastIndexOf(".");
 
     QStringList absolutePaths;
-    for (const QString &fileName : finalList) {
+    for (const QString& fileName : qAsConst(finalList)) {
         absolutePaths << path + fileName;
     }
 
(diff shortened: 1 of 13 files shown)


app/src/importimageseqdialog.cpp
```diff
@@ -275,7 +272,7 @@ const PredefinedKeySetParams ImportImageSeqDialog::predefinedKeySetParams() cons
     dot = finalList[0].lastIndexOf(".");
 
     QStringList absolutePaths;
-    for (QString fileName : finalList) {
+    for (const QString &fileName : finalList) {
```
MrStevns, line 275:
```suggestion
    for (const QString &fileName : qAsCost(finalList)) {
```

By adding const here, you'll probably get an additional warning for 'Might detach container', as sch we use qAsConst(...) to make the list const also.

core_lib/src/structure/filemanager.cpp
```diff
@@ -562,7 +562,7 @@ int FileManager::countExistingBackups(const QString& fileName) const
     const QString& baseName = fileInfo.completeBaseName();
 
     int backupCount = 0;
-    for (QFileInfo dirFileInfo : directory.entryInfoList(QDir::Filter::Files)) {
+    for (const QFileInfo &dirFileInfo : directory.entryInfoList(QDir::Filter::Files)) {
```
MrStevns, line 565:
```suggestion
    for (const QFileInfo& dirFileInfo : qAsConst(directory.entryInfoList(QDir::Filter::Files))) { {
```

By adding const here, you'll get an additional warning saying 'Might detach container', as such we use qAsConst(...).

app/src/importimageseqdialog.cpp
HeCorr: this was a separate warning with many occurrences but not all of them went away by adding `qAsCost()` so I just skipped it entirely. thanks for the info.

core_lib/src/structure/filemanager.cpp
HeCorr: adding it like that fails:
image

adding it just around `directory` doesn't make the warning go away:
image

app/src/importimageseqdialog.cpp
HeCorr: nope, the CI doesn't like that.
image

MrStevns: My bad, I recalled that we used it elsewhere in the code but it doesn't seem so.. it's likely because of backward compatibility then.
Alright ignore those comments 😅

MrStevns (approved):
It's always nice to get rid of warnings, thanks for taking the initiative to fix them.
I will merge the PR now.

PR #1875, opened 2024-08-25
iss: 1873 - Fix frame not being drawn on the timeline

### LINKED ISSUE #1873: Timeline frame disappear
### Issue Summary

https://github.com/user-attachments/assets/b9412e71-63cc-4c0f-b524-2d785d178a1e

In the video above you can see that after the timeline selection is deselected, the current frame disappear from the timeline.
Although the frame will re-appear if you click on the frame.

### Actual Results
The last frame disappear when the timeline selection is undone

### Expected Results
The last frame not disappearing

### Steps to reproduce
1. Draw a stroke on keyframe 1
2. Create a new keyframe on pos 2
3. Select both keyframes on the timeline
4. Draw a stroke on pos 2
5. undo the action

### System Information
+ Pencil2D Version: 0.7.0

MrStevns:
The reason for this is that while we keep a list of selected keyframes to move and paint on the timeline, a keyframe also has its own "isSelected" state. This state is backed up through undo/redo operations while the layer::getSelectedFrames is not.
We use the isSelected state to ignore selected frames in the TimeLineCells::paintFrames while we use Layer::getSelectedFrames() to paint the selected frames. 

This means that we can get into a situation where the frame is neither painted as selected nor non selected because of the two different sources of truth.

The solution is to remove one of the sources. As such, I went ahead and removed isSelected from Keyframe, so there's only one truth. By doing so, said frames can no longer be put in a selected state through undo/redo operations, which I personally think is good. I've inspected a few other applications which does also not keep selections as undo/redo'able state.

There's also the benefit that we don't have to iterate through all frames to know which is selected.

To that end my conclusion is that we shouldn't have to do that either.

fixes #1873

### DIFF
--- a/app/src/timelinecells.cpp
+++ b/app/src/timelinecells.cpp
@@ -444,11 +444,18 @@ void TimeLineCells::paintFrames(QPainter& painter, QColor trackCol, const Layer*
 
     int recHeight = height - 4;
 
-    layer->foreachKeyFrame([&](KeyFrame* key)
+    const QList<int> selectedFrames = layer->getSelectedFramesByPos();
+    for (auto pair : layer->keyframes())
     {
+        const KeyFrame* key = pair.second;
         int framePos = key->pos();
         int recWidth = standardWidth;
         int recLeft = getFrameX(framePos) - recWidth;
+
+        // Selected frames are painted separately
+        if (selectedFrames.contains(framePos)) {
+            continue;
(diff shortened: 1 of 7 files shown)


MrStevns:
Replaced by #1887 as I made these changes with master as origin, so it can't go into an eventual 0.7.1 branch

Issue #1849, opened 2024-06-18
Improvement for working with Layers (naming and creating a new layer)


DearFox:
Stopping the workflow when creating a new layer by opening the Layer Properties window is a bad decision.
![изображение](https://github.com/pencil2d/pencil/assets/59219907/12cef8c0-b7b5-4b22-b3ed-78f0dc63b18f)

Usually an artist doesn't care what his layer is called, so it would be better to skip the display of this window when creating a new layer.
 To rename an already created layer, it would be more convenient to double-click on its name in order to call up in place of its name an input field containing the current name of the layer with all the characters highlighted.  The same thing can be called by pressing F2 (the standard key for renaming in many programs).
![изображение](https://github.com/pencil2d/pencil/assets/59219907/bf4a470a-6f67-4671-8cea-009cfee4a4e4)
To bring up the current "Layer Properties" window, double-click on the layer icon, or use RMB to bring up a context menu with options like: "Rename" and "Properties"

Jose-Moreno:
@DearFox Hi! Thank you for proposal.

While I personally do care about naming when working solo or with an animation team (in case the file moves down the pipeline so other people have to clean-up the roughs or color the final lineart) I definitely agree the modal window that appears when creating any layer is a workflow blocker.

There has been discussions about this in the past and we did come to the conclusion that it should be removed eventually, but so far it's been a problem of lacking time, rather than anything else.

With this ticket it should be easier for us to track this request. Thanks!

PR #1880, opened 2024-09-26
TimeLineCells rewrite part 1

MrStevns:
This PR consists of a rewrite of the timeline cells part where:

- TimeLineCells has been split into numerous components:
  - TimeLineLayerHeaderWidget
    This component currently consist of a single button which is the layer visibility dot but has been built so it can be extended trivially
      - LayerVisibilityButton
        A shared component that can either be used to control the local visibility state or the global, depending on the context it's created with.
  - TimeLineLayerList
    a full rewrite of the layer logic of TimeLineLayerCells, now built using the following widgets.
    The layer list also takes care of delegating information back to the track list and handles the dragging functionality of each cell.
    - TimeLineBaseCell
      A base call for the different potential cell types, although right now it's only used for TimeLineLayerCell.
    - TimeLineLayerCell
      A derivetive of TimeLineBaseCell that is mostly a container for the TimeLineLayerCellEditorWidget. It sets the base geometry for the editor widget when created and takes care of cleaning it up when removed
    - TimeLineLayerCellEditorWidget
      The component that actually draws the layer cell. Now a QWidget with a simple QHBoxLayout. The editor widget also contains a new widget, namely LineEditWidget.
      - LineEditWidget
        A generic component that can be used to edit a line of text, exists because Qt's own QLineEdit does not auto select the entire text when used...
        For the timeline layer component, this widget can be used to edit the name of the layer, by double-clicking.
    - TimeLineLayerCellGutterWidget
      The component that draws the layer gutter, which is either shown or hidden but always exist as a overlay widget on top of the list.
  - TimeLineTrackHeaderWidget
    The header component consists of drawing of track numbers, small scrubber and the onion skin logic.
    The logic has been left mostly intact.
  - TimeLineTrackList
    TimelineTrackList is mostly a refactoring of the existing logic in TimeLineCells that handled the track, frame visualization and selection logic.

Here's what the timeline look like now:
image

A bunch of other changes and improvements made along the way:
- The vertical scrollbar has been moved so it's next to the track cells rather than the entire timeline.
- The visibility dot and the layer type icon are now drawn in higher resolution for HDPI displays.
- The header visual has been made gray across both lists, separating the track numbers from the track cells a bit more.
- The visibility dot has been made a tad bigger, 10 pixels instead of nine, which aligns better when the widgets are placed on whole pixels. Before it looked slightly offset
- The layer gutter has been improved in a few aspects, both visually and functionally. The gutter will only be displayed now when you can move the layer to a different position. It has also been made slightly larger and easier to detect.
- The camera layer at the bottom can no longer be dragged.
- The input dialog no longer appears when creating a new layer, instead you have to double-click on the layer label to change it. For the camera, you can still double-click on the camera icon to change its camera settings and name.

You can see the improvements in action here:
timeline-rewrite

### TLDR
The left part of the timeline, namely the layer cells has been rewritten to use widgets, are now easier to maintain and extend.
The right part of the timeline, namely the track cells have been left mostly untouched and in large parts have just been moved into it's own component but still uses QPainter and contain the same logic as before.

fixes https://github.com/pencil2d/pencil/issues/1849

### DIFF
--- a/app/src/timelinetrackheaderwidget.cpp
+++ b/app/src/timelinetrackheaderwidget.cpp
@@ -51,8 +51,6 @@ void TimeLineTrackHeaderWidget::paintEvent(QPaintEvent*)
         paintOnionSkin(painter);
     }
     paintTicks(painter, palette);
-    paintSplitter(painter, palette);
-
     paintScrubber(painter, palette);
 }
 
@@ -77,16 +75,6 @@ void TimeLineTrackHeaderWidget::paintBackground(QPainter& painter, const QPalett
     painter.setPen(Qt::NoPen);
     painter.setBrush(palette.color(QPalette::Base));
     painter.drawRect(QRect(0, 0, width() - 1, 0));
-}
-
-void TimeLineTrackHeaderWidget::paintSplitter(QPainter& painter, const QPalette& palette) const
-{
-    // --- draw bottom line splitter for track bar
-    QPen pen;
-    pen.setBrush(palette.color(QPalette::Mid));
-    pen.setWidthF(1);
-    painter.setPen(pen);
-    painter.drawLine(0, rect().bottom(), width() - 1, rect().bottom());
 }
 
 void TimeLineTrackHeaderWidget::paintOnionSkin(QPainter& painter) const
--- a/app/src/timelinetrackheaderwidget.h
+++ b/app/src/timelinetrackheaderwidget.h
@@ -47,7 +47,6 @@ private:
     void paintBackground(QPainter& painter, const QPalette& palette) const;
     void paintTicks(QPainter& painter, const QPalette& palette) const;
     void paintScrubber(QPainter& painter, const QPalette& palette) const;
-    void paintSplitter(QPainter& painter, const QPalette& palette) const;
 
     int getFrameNumber(int x) const;
     int getFrameX(int frameNumber) const;


DearFox:
Oh, that sounds cool!
Can I download build this somewhere to try it out?

MrStevns:
> Oh, that sounds cool! Can I download build this somewhere to try it out?

Not at the moment but I'll look into it.

DearFox:
swr45udtyi
Renaming layers works a little weird and unintuitive at the moment.
You can select text with the mouse cursor outside of the rename state, but during renaming, attempting to select text or move the text pointer will simply reset the layer's rename state.
(on Windows)

MrStevns:
Nice catch! I'll look into fixing that. Thanks for testing @DearFox

MrStevns:
Should be fixed now.

I've also made some other changes so that:
* You cannot select the text while you're not editing
* The selection and selected background now have better contrast when the item is already selected.
* The context menu is only available in edit mode.
* You can remove focus by clicking anywhere.

Builds:
Windows
macOS
Linux

DearFox:
Thanks, it works pretty intuitively now.
Here's a little demo for anyone who wants to see how it works:
se4ru6dti

Also, I just noticed that for the camera layer you can still open the dialog to change the name and resolution of the frame.
изображение
To do this, you need to double-click on the layer icon, or on the space above or below the text. I think there are about a couple of pixels there that are not affected by the text hitbox.
изображение

I also just discovered a slightly strange feature, when changing text the ESC key in some conditions works like Ctrl + z
ewr4u6dtuki
In this example, first I rename the layer several times and press Enter to confirm the rename. But then I start renaming and reset it using ESC. You can see that the text has a gray background. At this point, if the layer name was not changed, it will revert to the previous version, as if the user used ctrl+z.

DearFox:
изображение
I think it might be a good idea to move the resolution change settings to this settings menu.

Although I don't really like this logic of using tools to access some layer functions. But then we could remove this weird pop-up menu that blocks work with the program.

Ideally, there should be a separate floating window with layer and tool settings, roughly as it is implemented in Flash\Animate. At least, it seems to me that such logic is a bit more intuitive. Although in the case of Flash\Animate, there is a kind of division of focus. That is, when the user selects a tool, he sees the settings of this tool. When the user clicks on a layer in the timeline window, he sees the settings of this layer, but he still has his tool selected. And the user only needs to start drawing or using the previously selected tool on the canvas, so that the settings change again from the layer settings to the settings of the selected tool.
Example:
e56uetdyj
Sorry, I don't know how in Animate I can change the interface language to English.

MrStevns:
I've made another fix for the undo behavior after editing finished. Now it should behave more like a cancel function.

As for the camera dialog, yes it's intentionally still accessible through the the camera layer, by pressing on the icon or near it. While I know that some applications like Flash, OpenToonz and possibly also ToonBoom just add another panel, I don't want to add that kind of clutter to the interface, even if it is temporary.

How we'll fix the infamously hidden camera settings dialog, is still unsolved for now.

MrStevns:
PR is now up to date again

PR #1882, opened 2024-09-29
iss-1881: Move frames as far as possible

### LINKED ISSUE #1881: 0.7 Keyframe Moving Beyond Layer Names
### Issue Summary
In 0.7 moving keyframes to the left beyond frame 1 cuts them back to where they were before you start moving them.
 

### Actual Results
Back in 0.6.6, no matter how far left you moved the keyframes, they would stop at frame 1 and would stay there, in 0.7, they no longer do and when you do move them leftwards beyond frame 1 and stop holding left click they just cut back to where they were before you start moving them.
 

### Expected Results
Basically I expect it to work how it did before back in 0.6.6. The video shows the example (seen below)
 

### Video or Image Reference

https://github.com/user-attachments/assets/e3617d4d-d506-4871-8ce3-9777d416e457

 

### Steps to reproduce
Go on Pencil2d 0.7, make a keyframe at frame 1 so now there's 2 keyframes (since when opening there's already one), select them both (making them both gray) drag them right, deselect them (so they become blue), select them again (they become gray), drag them as far as you can to the right, then unclick, do the same in 0.6.6 and you'll see what i mean.
 

### System Information
+ Pencil2D Version: v0.6.6
+ Pencil2D Version: Nightly Build 2024-08-07_22:51:33 (this is what i've been calling 0.7)
(these two titles are what is shown on the top left of the screen upon opening their versions)

 

+ Operating System: 
Windows 10
 

+ RAM Size:
I have no idea
 

+ Graphics Tablet:
nope

MrStevns:
This fixes #1881 

Instead of moving keyframes back to where they were, the frames will now be shifted to the right until there's room for all selected frames.

issue-1881

### DIFF
--- a/core_lib/src/structure/layer.cpp
+++ b/core_lib/src/structure/layer.cpp
@@ -503,14 +503,25 @@ void Layer::deselectAll()
     }
 }
 
-bool Layer::canMoveSelectedFramesToOffset(int offset) const
+bool Layer::canMoveSelectedFramesToOffset(int fromIndex, int offset) const
 {
     QList<int> newByPositions = mSelectedFrames_byPosition;
 
-    for (int pos : newByPositions)
+    int step = -1;
+
+    if (fromIndex < 0 || fromIndex > newByPositions.count() - 1) {
+        return false;
+    }
+
+    if (offset < 0) {
+        step = 1;
+    }
+
+    for (; fromIndex> -1 && fromIndex < newByPositions.count(); fromIndex += step)
     {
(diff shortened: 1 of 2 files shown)


MrStevns:
Thanks for reviewing Jacob. If there was a reason for making it so elaborated, I cannot remember it anymore.

Your suggested diff does indeed work. I don't see a reason to complicate the logic further if we can avoid it. I'll revert my changes and apply yours instead.

PR #1885, opened 2024-10-19
Feature/iss 748 - undo redo keyframes

### LINKED ISSUE #748: Allow to UNDO / REDO "deleting" new frames and layers
### --Issue Summary--
Is it normal behaviour that the UNDO feature is unable to UNDO the addition of new frames, even the addition of new Layers?

MrStevns:
This PR adds the following to the undo/redo system:

- Adding keyframes
- Deleting keyframes
- Moving keyframes

In addition I had to make some changes to the structure of the undoredomanager, as it wasn't as flexible as I needed it to be. The change is for the better though.

I've also introduced a set of guards to avoid dealing with an undo/redo stack that is potentially useless, for example when you delete the layer where there were once keyframes.
Since we don't support that yet, we have to guard against it to avoid crashing.

partially fixes #748

### DIFF
--- a/core_lib/src/managers/undoredomanager.h
+++ b/core_lib/src/managers/undoredomanager.h
@@ -41,9 +41,7 @@ class UndoRedoCommand;
 /// The undo/redo type which correspond to what is being recorded
 enum class UndoRedoRecordType {
     KEYFRAME_MODIFY, // Any modification that involve a keyframe
-
-    // Possible future actions
-    // KEYFRAME_REMOVE, // Removing a keyframe
+    KEYFRAME_REMOVE, // Removing a keyframe
     // KEYFRAME_ADD, // Adding a keyframe
     // SCRUB_LAYER, // Scrubbing layer
     // SCRUB_KEYFRAME, // Scrubbing keyframe
@@ -156,7 +154,9 @@ private:
     void replaceBitmap(const UndoSaveState& undoState, const QString& description);
     void replaceVector(const UndoSaveState& undoState, const QString& description);
 
-    const UndoSaveState* savedKeyFrameState() const;
+    void removeKeyFrame(const UndoSaveState& undoState, const QString& description);
+
+    const UndoSaveState* savedKeyFrameState(const UndoRedoRecordType& type) const;
 
     void pushCommand(QUndoCommand* command);
 
--- a/core_lib/src/managers/undoredomanager.h
+++ b/core_lib/src/managers/undoredomanager.h
@@ -43,6 +43,7 @@ enum class UndoRedoRecordType {
     KEYFRAME_MODIFY, // Any modification that involve a keyframe
     KEYFRAME_REMOVE, // Removing a keyframe
     KEYFRAME_ADD, // Adding a keyframe
+    KEYFRAME_MOVE,
     // SCRUB_LAYER, // Scrubbing layer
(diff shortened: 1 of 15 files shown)


nonemec-havok:
Just taking my first steps with Pencil2D and this turns out to be one of the most painful missing features: Copy&Paste of keyframes is a tricky operation and often goes wrong. That on its own would be annoying. The fact that the operation is then not undoable is horribly painful because it can leave the timeline completely messed up. Hope this feature gets completed at some point!

MrStevns:
When i wrote this PR and also back when I started on the undo/redo rewrite I had my doubts about undoSaveState implementation, specifically the way we deal with the lifetime of the current save state and delete it. I was never fully happy with that implementation and just recently bumped into the reason why it was indeed a bad idea. Dangling ptrs...

I knew about the problem but at the time i couldn't figure out a better solution, so I left it there...

With this PR and especially the latest changes, I think I have finally come up with a solution that I like and is safer to use.

Instead of getting access to the raw ptr, we get a handle to the save state called SAVESTATE_ID. The id is used to fetch a given saveState stored temporarily in undoRedoManager. This now also means that we can create multiple save states and apply all of them without having to think about dealing with their lifetime. Once the state has been recorded, it will deleted automatically.

In most cases the save state will almost always be accompanied by record(...) right away but I did bump into a few scenarios I hadn't thought of, which showed the flaws in the previous implementation. By keeping a list of save states rather than only one, we fix this problem. I thought about using a circular buffer instead since it's not likely we'll need more than 5 slots but then again, KISS until proven otherwise...

With that I think the PR is ready to be reviewed.

chchwy:
Will be reviewing this PR soon.

core_lib/src/managers/undoredomanager.h
```diff
@@ -38,20 +39,22 @@ class KeyFrame;
 class LegacyBackupElement;
 class UndoRedoCommand;
 
+typedef int SAVESTATE_ID;
```
chchwy:
```suggestion
using SAVESTATE_ID = int;
```

MrStevns:
Thanks for reviewing chchwy.
Your comments has been addressed.

I thought about using `using` instead of typedef but I wasn't sure if we were fully onboard c++11 yet.

chchwy:
> Thanks for reviewing chchwy. Your comments has been addressed.
> 
> I thought about using `using` instead of typedef but I wasn't sure if we were fully onboard c++11 yet.

We are using C++14 as it's required for the catch2 test framework. And as for the master branch, I believe it's time to move to C++17 or newer.

chchwy:
Hey @MrStevns I gave this branch a spin and found a bug, please see the sceen recording:
https://drive.google.com/file/d/1czqN8ENiHJ-36NkxkBqg0_XE8C7tCjN0/view?usp=sharing

The steps to replicate:

1. Create frame 5, Create frame 6
2. Draw on frame 5 and frame 6
3. Select both frame 5 and frame 6 in Timeline
4. Move the 2 frames by dragging left
5. Undo
6. The undo doesn't restore the frames properly

MrStevns:
Ah nice spotted.. I tested all kinds of moves except one that overlaps 🤦 

I've changed it to use moveSelectedFrames now which is actually also more correct, as it now corresponds to the same action that was initially performed

chchwy:
All good. Thanks! I am going to merge this one now.

PR #1887, opened 2024-11-04
Fix #1873 - Fix frame not being drawn on the timeline

### LINKED ISSUE #1873: Timeline frame disappear

MrStevns:
The reason for this is that while we keep a list of selected keyframes to move and paint on the timeline, a keyframe also has its own "isSelected" state. This state is backed up through undo/redo operations while the layer::getSelectedFrames is not.
We use the isSelected state to ignore selected frames in the TimeLineCells::paintFrames while we use Layer::getSelectedFrames() to paint the selected frames.

This means that we can get into a situation where the frame is neither painted as selected nor non selected because of the two different sources of truth.

The solution is to remove one of the sources. As such, I went ahead and removed isSelected from Keyframe, so there's only one truth. By doing so, said frames can no longer be put in a selected state through undo/redo operations, which I personally think is good. I've inspected a few other applications which does also not keep selections as undo/redo'able state.

There's also the benefit that we don't have to iterate through all frames to know which is selected.

To that end my conclusion is that we shouldn't have to do that either.

fixes https://github.com/pencil2d/pencil/issues/1873

### DIFF
--- a/app/src/timelinecells.cpp
+++ b/app/src/timelinecells.cpp
@@ -446,16 +446,15 @@ void TimeLineCells::paintFrames(QPainter& painter, QColor trackCol, const Layer*
     int recHeight = height - 4;
 
     const QList<int> selectedFrames = layer->getSelectedFramesByPos();
-    for (auto pair : layer->keyframes())
+    layer->foreachKeyFrame([&](KeyFrame* key)
     {
-        const KeyFrame* key = pair.second;
         int framePos = key->pos();
         int recWidth = standardWidth;
         int recLeft = getFrameX(framePos) - recWidth;
 
         // Selected frames are painted separately
         if (selectedFrames.contains(framePos)) {
-            continue;
+            return;
         }
 
         if (key->length() > 1)
@@ -475,7 +474,7 @@ void TimeLineCells::paintFrames(QPainter& painter, QColor trackCol, const Layer*
         }
 
         painter.drawRect(recLeft, recTop, recWidth, recHeight);
-    }
+    });
 }
 
 void TimeLineCells::paintCurrentFrameBorder(QPainter &painter, int recLeft, int recTop, int recWidth, int recHeight) const
--- a/app/src/timelinecells.cpp
+++ b/app/src/timelinecells.cpp
@@ -445,11 +445,18 @@ void TimeLineCells::paintFrames(QPainter& painter, QColor trackCol, const Layer*
 
     int recHeight = height - 4;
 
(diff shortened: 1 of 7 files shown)


chchwy:
Totally agree with you about removing "isSelected" from Keyframe.

app/src/timelinecells.cpp
```diff
@@ -444,12 +444,19 @@ void TimeLineCells::paintFrames(QPainter& painter, QColor trackCol, const Layer*
 
     int recHeight = height - 4;
 
-    layer->foreachKeyFrame([&](KeyFrame* key)
+    const QList<int> selectedFrames = layer->getSelectedFramesByPos();
+    for (auto pair : layer->keyframes())
```
chchwy, line 448: Doesn't the original `layer->foreachKeyFrame()` fit your need here?

MrStevns: Ah.. yes that's right.. I forgot about that.
I've fixed it now.

core_lib/src/structure/layer.h
```diff
@@ -170,6 +170,8 @@ class Layer
     /** Clear the list of dirty keyframes */
     void clearDirtyFrames() { mDirtyFrames.clear(); }
 
+    std::map<int, KeyFrame*, std::greater<int>> keyframes() const { return mKeyFrames; }
```
chchwy, line 173: Could you please remove this unused function?

I don't want to expose the internal data structure outside the Layer class. Looking back, my decision to use std::map instead of std::vector wasn't very smart in hindsight.

chchwy (approved):

chchwy:
Apart from the clean up mentioned above, I don't have further questions. Thanks and well done @MrStevns

core_lib/src/structure/layer.h
MrStevns: Done

chchwy: thanks!

PR #1892, opened 2024-11-17
Use keyPressEvent rather than shortcut for polyline delete segment

scribblemaniac:
#1861 Introduced the ability to delete segments of an active polyline with a shortcut (the backspace key by default). This prevented the keypress event from being received by ScribbleArea::keyPressEvent, which in turn broke the delete selection action. This PR reverts the creation of the new shortcut, and uses PolylineTool::keyPressEvent instead to trigger this behavior. This allows for both the polyline tool and selection to use the backspace shortcuts. I also think this is a better way to do things code-wise.

However, it does mean that the polyline delete segment action is another one of shortcuts which is currently hardcoded. I do think it was a nice idea to have this as a customizable shortcut, but I think window-level shortcuts were the wrong way to go about this, or at the very least, *all* shortcuts would have to be handled this way to prevent some from getting overridden the delete selection action was.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -17,6 +17,7 @@
 + Fix hanging when adjusting cursor because of too many writes to disk [#1853](https://github.com/pencil2d/pencil/pull/1853)
 + Avoid updating width/feather sliders for tools that don’t use them [cce3107](https://github.com/pencil2d/pencil/commit/cce31079c871fcc04e957c44d5c6e65990f635f1)
 + Fix fill misbehaving when drawing was partly outside border [#1865](https://github.com/pencil2d/pencil/pull/1865)
++ Fix clearing selection with the delete shortcut [#1892](https://github.com/pencil2d/pencil/pull/1892)
 
 ## Pencil2D v0.7.0 - 12 July 2024
 
--- a/app/src/mainwindow2.cpp
+++ b/app/src/mainwindow2.cpp
@@ -40,7 +40,6 @@ GNU General Public License for more details.
 #include "pencilsettings.h"
 #include "object.h"
 #include "editor.h"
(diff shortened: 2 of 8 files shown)


MrStevns:
It is supposedly possible to have both hardcoded events and shortcuts, using event filters and listening to the QShortcutOverride property type. Was that something you looked into?

Saying that we either do hardcoded or customized is not easy nor entirely possible elther. For example what about modifiers which can't be bound to shortcuts on their own?

scribblemaniac:
QShortcutOverride is not something I've looked into yet, and does look like it could be helpful. It is possible to customize modifiers-only shortcuts with code similar this in keyPressEvent:
```c++
QSettings settings(PENCIL2D, PENCIL2D);
int keyCode = settings.value(KEY_EYEDROPPER, Qt::Key_Alt).toInt();
if (event.key() == keyCode) { doStuff() }
```
Of course this won't work with multiple keys, and Qt has no way to retrieve all currently held buttons (aside from modifier keys). Similarly we could store and retrieve QKeySequences and compare them in QShortcutEvent events, but that won't work for modifier-only shortcuts. There are other things to consider too such as interactions between shortcuts, loosing window focus, etc. It's a problem that requires more thought to properly solve. For the immediate issue with the selection deletion, I think we should keep it simple.

MrStevns (approved):
You're right, the whole shortcut, focus, and in general keyboard interaction requires a better foundation to get working properly.

I don't expect that to be sorted in this PR, I'll accept the PR, then we'll have to figure out how to deal with these combinations in the future.

Maybe we can use QShortcutOverride or maybe we'll have to figure out another way.

PR tested. Deletion of selections work again and deleting polylines using escape Key also still works.

PR #1895, opened 2024-11-24
Add more details to FileManager error report

MrStevns:
The purpose of this PR has been to make the error report cover more steps, introduce "Error" to make it more explicit what's a information, a warning and an error.

I've also tried to make it fail faster rather than spitting out all kinds of irrelevant information when the actual error happened 4 lines above.

Here's an example of what our current report looks like:
```
Raw file path: /Users/mrstevns/Desktop/p2d testing/test-save-corruption.pclx
Resolved file path: /Users/mrstevns/Desktop/p2d testing/test-save-corruption.pclx
  File name: /Users/mrstevns/Desktop/p2d testing/test-save-corruption.pclx
  Recognized New zipped Pencil2D File Format (*.pclx) !
    Miniz found an error while reading the file. - 7, failed finding central directory
    Miniz found an error while closing file file. - 24, invalid parameter
  XML file: /private/var/folders/5c/2bmlz9fx2kzbd56j81ktpghr0000gn/T/Pencil2D/test-save-corruption_Y2xD_cplu6r9q/main.xml
  Data folder: /private/var/folders/5c/2bmlz9fx2kzbd56j81ktpghr0000gn/T/Pencil2D/test-save-corruption_Y2xD_cplu6r9q/data
  Working folder: /private/var/folders/5c/2bmlz9fx2kzbd56j81ktpghr0000gn/T/Pencil2D/test-save-corruption_Y2xD_cplu6r9q/
  Main XML file does not exist
System Info
Pencil2D version: 0.7.0.911 (stable)
Build ABI: x86_64-little_endian-lp64
Kernel: darwin, 21.6.0
Operating System: macOS 12.7
end
```

And here's the new one:
```
Raw file path: /Users/mrstevns/Desktop/p2d testing/test-save-corruption.pclx
Resolved file path: /Users/mrstevns/Desktop/p2d testing/test-save-corruption.pclx
  
[Project LOAD diagnostics]

  File name: /Users/mrstevns/Desktop/p2d testing/test-save-corruption.pclx
  Working dir: /private/var/folders/5c/2bmlz9fx2kzbd56j81ktpghr0000gn/T/Pencil2D/test-save-corruption_Y2xD_719f2v7t/
  Project format: .pclx
    
[Miniz sanity check]

    Found an error while reading the file. Error code: 7, reason: failed finding central directory
    Found an error while closing the file. Error code: 24, reason: invalid parameter
  
Error: Unable to extract project, miniz sanity check failed.

[System Info]

  Pencil2D version: 0.0.0.0 (dev)
  Build ABI: x86_64-little_endian-lp64
  Kernel: darwin, 21.6.0
  Operating System: macOS 12.7
  Language: en_GB
```

In addition the error dialog now also has a "Copy to clipboard" button, so the user can more easily paste the error report.

image

### DIFF
--- a/app/ui/errordialog.ui
+++ b/app/ui/errordialog.ui
@@ -83,6 +83,16 @@
     </layout>
    </item>
    <item>
+    <widget class="QLabel" name="label">
+     <property name="text">
+      <string>The following report contains vital information to figure out the cause of the error. Make sure to copy all of it, if you intend to report the bug.</string>
+     </property>
+     <property name="wordWrap">
+      <bool>true</bool>
+     </property>
+    </widget>
+   </item>
+   <item>
     <widget class="QTextEdit" name="details">
      <property name="readOnly">
       <bool>true</bool>
--- a/app/ui/errordialog.ui
+++ b/app/ui/errordialog.ui
@@ -85,7 +85,7 @@
    <item>
     <widget class="QLabel" name="label">
      <property name="text">
-      <string>The following report contains vital information to figure out the cause of the error. Make sure to copy all of it, if you intend to report the bug.</string>
+      <string>This report contains vital information. Copy all of it when submitting a bug.</string>
      </property>
      <property name="wordWrap">
       <bool>true</bool>
--- a/core_lib/src/graphics/vector/vectorimage.cpp
+++ b/core_lib/src/graphics/vector/vectorimage.cpp
@@ -23,6 +23,7 @@ GNU General Public License for more details.
 #include <QDebug>
 #include <QXmlStreamWriter>
 #include "object.h"
+#include "util.h"
 
 
 VectorImage::VectorImage()
@@ -82,6 +83,9 @@ bool VectorImage::read(QString filePath)
     {
         return false;
     }
+    ScopeGuard fileScope([&] {
+        file.close();
+    });
 
     QDomDocument doc;
     if (!doc.setContent(&file)) return false; // this is not a XML file
@@ -123,6 +127,9 @@ Status VectorImage::write(QString filePath, QString format)
         debugInfo << ("file.error() = " + file.errorString());
         return Status(Status::FAIL, debugInfo);
     }
+    ScopeGuard fileScope([&] {
+        file.close();
+    });
 
     if (format != "VEC")
     {
--- a/core_lib/src/graphics/vector/vectorimage.cpp
+++ b/core_lib/src/graphics/vector/vectorimage.cpp
@@ -83,9 +83,6 @@ bool VectorImage::read(QString filePath)
     {
         return false;
     }
-    ScopeGuard fileScope([&] {
-        file.close();
-    });
 
     QDomDocument doc;
     if (!doc.setContent(&file)) return false; // this is not a XML file
@@ -127,9 +124,6 @@ Status VectorImage::write(QString filePath, QString format)
         debugInfo << ("file.error() = " + file.errorString());
         return Status(Status::FAIL, debugInfo);
     }
-    ScopeGuard fileScope([&] {
-        file.close();
-    });
 
     if (format != "VEC")
     {
--- a/core_lib/src/qminiz.cpp
+++ b/core_lib/src/qminiz.cpp
@@ -83,6 +83,9 @@ Status MiniZ::compressFolder(QString zipFilePath, QString srcFolderPath, const Q
     mz_zip_zero_struct(mz);
 
     mz_bool ok = mz_zip_writer_init_file(mz, zipFilePath.toUtf8().data(), 0);
+    ScopeGuard mzScopeGuard2([&] {
+        mz_zip_reader_end(mz);
+    });
 
     if (!ok)
     {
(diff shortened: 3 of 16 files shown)


app/ui/errordialog.ui
```diff
@@ -82,6 +82,16 @@
      </item>
     </layout>
    </item>
+   <item>
+    <widget class="QLabel" name="label">
+     <property name="text">
+      <string>The following report contains vital information to figure out the cause of the error. Make sure to copy all of it, if you intend to report the bug.</string>
```
chchwy, line 88: I suggest a more concise message: "This report contains vital information. Copy all of it when submitting a bug." Shorter messages often increase the likelihood of a response.

core_lib/src/graphics/vector/vectorimage.cpp
```diff
@@ -82,6 +83,9 @@ bool VectorImage::read(QString filePath)
     {
         return false;
     }
+    ScopeGuard fileScope([&] {
+        file.close();
+    });
```
chchwy, line 88: you can use `OnScopeExit(file.close();)`

chchwy: BTW, QFile automatically closes the file upon destruction, so explicitly closing it is not necessary.

core_lib/src/qminiz.cpp
```diff
@@ -66,27 +68,32 @@ size_t MiniZ::istreamReadCallback(void *pOpaque, mz_uint64 file_ofs, void * pBuf
...
     dd << QString("Creating Zip %1 from folder %2").arg(zipFilePath, srcFolderPath);
 
     if (!srcFolderPath.endsWith("/"))
     {
+        dd << "Adding / to path";
         srcFolderPath.append("/");
     }
 
     mz_zip_archive* mz = new mz_zip_archive;
+    ScopeGuard mzScopeGuard([&] {
```
chchwy, line 81: `OnScopeExit(delete mz);`

core_lib/src/qminiz.cpp
```diff
@@ -166,17 +165,22 @@ Status MiniZ::uncompressFolder(QString zipFilePath, QString destPath)
...
+        delete mz;
+    });
+
     mz_zip_zero_struct(mz);
 
     mz_bool ok = mz_zip_reader_init_file(mz, zipFilePath.toUtf8().data(), 0);
-
-    ScopeGuard mzScopeGuard([&] {
+    ScopeGuard mzScopeGuard2([&] {
```
chchwy, line 175: You can combine `mzScopeGuard` and `mzScopeGuard2` into a single one.

core_lib/src/soundplayer.cpp
```diff
@@ -41,6 +42,10 @@ void SoundPlayer::init(SoundClip* clip)
 
     QFile file(clip->fileName());
     file.open(QIODevice::ReadOnly);
+    ScopeGuard fileScope([&] {
+        file.close();
```
chchwy, line 46: Same, this `file.close()` is not really necessary. If really want to close it explicitly, you can do it after `file.readAll()`

core_lib/src/structure/filemanager.cpp
```diff
@@ -92,21 +104,27 @@ Object* FileManager::load(const QString& sFileName)
...
         return nullptr;
     }
     if (!file.open(QFile::ReadOnly))
     {
+        dd << "Error: Main XML file is read only!";
         handleOpenProjectError(Status::ERROR_FILE_CANNOT_OPEN, dd);
         return nullptr;
     }
+    ScopeGuard fileScope([&] {
+        file.close();
```
chchwy, line 118: same as above.

chchwy:
Thanks for thoroughly revising all the project load/save error logs and adding the missing parts. This is very important. Apart from the explicit `QFile::close` calls, everything looks great to me!

chchwy (approved):

core_lib/src/graphics/vector/vectorimage.cpp
MrStevns: Hmm interesting, where did you read that? I haven't been able to find anything that confirms that. I can only find the QFIleDevice::AutoCloseHandle which is DontCloseHandle by default, as such that should not apply here.

Anyway, I prefer that we do it explicitly :)

core_lib/src/qminiz.cpp
MrStevns: Indeed, I just prefer having two different scope guards in this case, mzScopeGuard2 is only relevant if the writer has been opened, while the first one is always relevant.

core_lib/src/soundplayer.cpp
MrStevns: Does it matter? The good thing about the ScoepGuard is that it happens at the end of the scope, so as long as we've initiated the ScopeGuard, it'll happen at the end no matter what. Right?

core_lib/src/graphics/vector/vectorimage.cpp
chchwy: yeh, that's not very obvious. it's in the description of the `~QFile()`. 
https://doc.qt.io/qt-6/qfile.html#dtor.QFile

If you look at the source code, it's in the destructor of `QFileDevice`, which is the base class of `QFile`
https://github.com/qt/qtbase/blob/4f9cbd5b2892f0f72c74db05d05c67f25e7c99e7/src/corelib/io/qfiledevice.cpp#L236

chchwy: If you can find a good place to call close, it's good to close file handle early as it can be released immediately.
however, if you put `close()` in a scope guard, it duplicates the behaviour already handled by QFile's destructor.

core_lib/src/qminiz.cpp
chchwy: I was concerned about the destruction order of local objects, which can be a bit tricky. Then I checked the C++ standards, local objects are destructed in reverse order of constructions, so `mzScopeGuard2` will always execute before `mzScopeGuard`. This should be fine.

core_lib/src/soundplayer.cpp
MrStevns: Ah right, in that case i'll remove it again

PR #1898, opened 2025-01-02
#1897 - Fix view rotation effects affects various view handles

### LINKED ISSUE #1897: 🐛 [BUG] - Canvas rotation affects view handles
### Description

When rotating the canvas, the tolerance for grabbing the different view handles controlling the camera and the perspective overlays changes.

For example after rotating the canvas 90 deg, the handles are no longer draggable.

### Expected result

The handles are always draggable

### Actual result

The handles get more and more difficult to grab as you rotate closer to 90 or 270 degrees. All other handles work but the tolerance gets smaller and smaller, making it more difficult to grab at say 70 deg as well

### Steps to reproduce

```bash
1. Click on the camera layer
2. Select to move tool
3. Drag the border handles of the camera bounding box to see that it works
4. Rotate the canvas 90 deg (not the camera!)
3. Hover the mouse over the camera handles and try to drag
```

### Video or Image Reference

handle-bug

### System Information

```bash
- Pencil2D version: Version: 0.7.0.911
commit: 7e83608a957dfeacc739ec0a09a89291de0d4e6b
date: 2024-06-19_01:01:26

- Operation system: macOS
```

MrStevns:
Fixes #1897 as reported on discord. The same problem also existed with the overlay handles. prior to this fix.


chchwy (approved):

chchwy:
This PR was merged through a manual git rebase operation rather than through the web interface.

PR #1904, opened 2025-02-15
Write tool properties to disk when calling leavingThisTool() or ~ToolManager()

khoidauminh:
Tool properties are written to disk immediately when changed. If you try to slide a scalar value with a high temporal resolution device like a tablet or a high end mouse, the flooding events will cause the program to freeze/lag as it struggles to write every value to the QSettings object.

Previously #1853 was merged to mitigate this problem with quick sizing, but if you change the properties on the tool options panel, the lag is still there.

This PR makes the change so that the program will wait until you switch to another tool, or close the program, to write everything onto the disk.

### DIFF
--- a/core_lib/src/managers/toolmanager.cpp
+++ b/core_lib/src/managers/toolmanager.cpp
@@ -36,6 +36,14 @@ ToolManager::ToolManager(Editor* editor) : BaseManager(editor, __FUNCTION__)
 {
 }
 
+ToolManager::~ToolManager()
+{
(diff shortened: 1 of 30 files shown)


chchwy (approved):

khoidauminh:
I think I should go ahead and remove the changes in #1853 as well since it's now redundant

chchwy (approved):

khoidauminh:
Thank you!

PR #1905, opened 2025-02-27
Enable antialiasing option for move tool

khoidauminh:
Gives an option for the artist to transform without smoothing. Useful for pixel art.


MrStevns (approved):
Hi khoidauminh

Thanks for contributing and sorry for the slow review time.

The PR looks good to me.

It's a bit annoying that we can't see the drawing being affected by the anti-aliasing flag in real time but that's not your fault. It comes down to the way the transformed selection is painted on the canvas.

I have a PR in the works to address that specifically as well as some other concerns regarding the move and selection logic.

PR #1913, opened 2025-04-25
UI Feature: Custom Titlebar

MrStevns:
What started out as an experiment, has now become a full blown feature. I present to you, our own custom title bar.

A custom title bar gives us more control of the sizing of our widgets, especially in floating mode, as Qt will otherwise favor the OS min dimensions. Normally I wouldn't see this as a problem but the resizing system is Qt is so quirky sometimes that I experience that there are not enough handles to take control of the sizing properly.

With our new custom title bar, this is no longer the case. We are in full control of what we want to show, how it should be sized and what it should look like.

What our current one looks like on each respective platform:
| macOS | Windows | Linux |
| ------- | --------- | ------ |
| image image  | image | image | 

The new one:
| macOS | Windows | Linux |
| ------- | --------- | ------ |
| image | image | image |

The title bar has been built using the standard palette which should make sure that it looks correct on all platforms. I've designed some new icons as well, a normal one, an active one and the same in dark variations.

The most prominent changes can be seen on macOS, especially the dark mode titlebar which was previously shown in a terrible gradient... the icons were not high DPI either.

For the other platforms, the change should be minor. On windows the title bar looks nearly identical except for the icons of course. When I tested the build on Peppermint OS, previously the title bar was more embedded into the widget, where as with the new one the title bar background is as prominent as the other platforms. Personally I think it looks fine that way but I am also not daily driving a Linux distribution. What do you think?

When the title bar becomes too narrow to show both the title and the buttons, it will favor the title and hide the buttons. The buttons can however still be accessed by right-clicking. Although in the right-click menu, there's no animation currently when hovering the items, so it uses the active representation instead.

image

Visual highlights:
image

| Light | Dark |
| ----- | ----- |
| hover-effect | hover-effect2 |

Builds to try:
Windows
macOS
Linux


app/src/basedockwidget.cpp
```diff
@@ -34,8 +35,54 @@ BaseDockWidget::BaseDockWidget(QWidget* pParent)
                       "border-width: 1px; }");
     }
 #endif
+
+    mTitleBarWidget = new TitleBarWidget();
+    mNoTitleBarWidget = new QWidget();
```
chchwy, line 40: Will this `mNoTitleBarWidget` be deleted anywhere?

app/src/titlebarwidget.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+    mCloseButton->setIcon(closeIcon);
+    mCloseButton->setIconSize(iconSize);
+    mCloseButton->setFixedSize(iconSize + padding);
+    mCloseButton->installEventFilter(new ButtonAppearanceWatcher(closeButtonRes,
+                                                                 closeHoverButtonRes,
+                                                            this));
+
+    connect(mCloseButton, &QToolButton::clicked, this, [this] {
+        emit closeButtonPressed();
```
chchwy, line 85: Same here, the classic connection syntax should work?
```
    connect(mCloseButton, &QToolButton::clicked, this, &TitleBarWidget::closeButtonPressed);
```

app/src/titlebarwidget.h
```diff
@@ -0,0 +1,68 @@
...
+private:
+    QString flatButtonStylesheet() const;
+    void showEvent(QShowEvent* event) override;
+    void hideButtons(bool hide);
+    void hideButtonsIfNeeded(int width);
+
+    QWidget* createNormalTitleBarWidget(QWidget* parent);
+
+    QWidget* mNormalTitleBarWidget = nullptr;
+    QLayout* mContainerLayout = nullptr;
```
chchwy, line 55: Looks like this `mContainerLayout` doesn't have to be a member variable? A local variable in createNormalTitleBarWidget() is perfectly fine.

MrStevns: Yep looks like it, the same can be said with mNormalTitleBarWidget
it's likely from some work i discarded. I'll change it.

app/src/titlebarwidget.cpp
```diff
@@ -0,0 +1,216 @@
...
+    containerWidget->setMinimumSize(QSize(1,1));
+
+    return containerWidget;
+}
+
+QString TitleBarWidget::flatButtonStylesheet() const
+{
+    return "QToolButton {"
+            "border: 0;"
+            "}";
```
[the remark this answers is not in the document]
MrStevns: Done

app/src/titlebarwidget.cpp
MrStevns: You're right, I think the lambda is from when i had some local logic involved.
Fixed.

app/src/basedockwidget.cpp
chchwy: Thanks! It's good to ensure that there is an object that owns the title bar widget so it can always be freed properly.

app/src/titlebarwidget.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+    mMenu = new QMenu(this);
+    mMenu->addAction(closeHoverButtonRes.lightMode, tr("Close"), [=] {
+        emit closeButtonPressed();
+    });
+    mDockAction = mMenu->addAction(dockHoverButtonRes.lightMode, "", [=] {
+        emit undockButtonPressed();
+    });
+
+    setContextMenuPolicy(Qt::CustomContextMenu);
```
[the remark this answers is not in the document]
chchwy: Since both buttons/context menu do the same thing, is there a real use case where users would want to close the tools widget while it's minimized? What if we applied the subtraction principle here - keep the UI code simple until someone actually complains about it? What do you think?

app/src/titlebarwidget.cpp
```diff
@@ -0,0 +1,210 @@
...
+    QIcon dockIcon = dockButtonRes.iconForMode(isDarkmode);
+    mDockButton->setIcon(dockIcon);
+    mDockButton->setStyleSheet(flatButtonStylesheet());
+
+    mDockButton->setIconSize(iconSize);
+    mDockButton->setFixedSize(iconSize + padding);
+    mDockButton->installEventFilter(new ButtonAppearanceWatcher(dockButtonRes, dockHoverButtonRes, this));
+
+    connect(mDockButton, &QToolButton::clicked, this, [this] {
+       emit undockButtonPressed();
```
chchwy, line 103: This can use the classic connection syntax, too, just like the closeButtonPressed connection above.

chchwy:
Apart from the context menu, I don't have further questions. This PR looks good to me.

app/src/titlebarwidget.cpp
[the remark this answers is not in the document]
MrStevns: Sure, if you don't think it's necessary, then let's try without first.
I've removed it now

app/src/titlebarwidget.cpp
MrStevns: Fixed

PR #1914, opened 2025-04-25
Use the latest GPLv2 license text

musicinmybrain:
Replaces the contents of LICENSE.txt with the latest https://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.

- FSF is now remote-only, so there is no longer a postal address
- References to the Library GPL are now to the Lesser GPL
- Some placeholders have been changed
- Some headings have been realigned with different whitespace


MrStevns (approved):
Looks good to me, I'll merge it. Thanks 👍

PR #1916, opened 2025-04-29
UI Feature: Improve toolbox flow layout

MrStevns:
I've given our FlowLayout class an overhaul.

Pure row height calculation has been moved into its own separate function, rather than mixing everything with the "testOnly" flag. The justified logic has been updated such that it will always try to cram as many items on a line as possible, but still be constrained to a minimum spacing.

The result of that looks like this:

| Before | After |
| ------ | ----- |
| old-flowlayout | new-flowlayout |

The sizing logic for the Tools widget has been updated as well. If there's only one row, then the items will be aligned centered and otherwise the alignment will change to justified, until all items has been layout. When all items has been layout horizontally, the flow layout will default to centered placement again.

I've also introduced a new ToolBoxLayout class in order to change the alignment of the last line, as it will otherwise align justified with no concern of aligning to lines above.

With this PR, combined with #1913, we will be able to shrink to the following.
image

Note that due to how QDockWidget sizing works, in order to get more control of how the toolbox is layout, I had to move everything into a container widget.

### DIFF
--- a/app/app.pro
+++ b/app/app.pro
@@ -90,6 +90,7 @@ HEADERS += \
     src/generalpage.h \
     src/shortcutspage.h \
     src/timelinepage.h \
+    src/toolboxwidget.h \
     src/toolspage.h \
     src/basedockwidget.h \
     src/colorbox.h \
@@ -142,6 +143,7 @@ SOURCES += \
     src/generalpage.cpp \
     src/shortcutspage.cpp \
     src/timelinepage.cpp \
+    src/toolboxwidget.cpp \
     src/toolspage.cpp \
     src/basedockwidget.cpp \
     src/colorbox.cpp \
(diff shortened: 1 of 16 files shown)


chchwy:
I just played this PR around on my Macbook and found that my toolbox cannot be shrunk to the minimum width in the "after" GIF because of the title bar.

圖片

MrStevns:
Thanks for testing Matt

Based on the image you provided, are you certain you've pulled the latest changes?
it shouldn't be possible to have the items aligned that far to the left, instead it should align to center. 
I've just re-tested the branch and it still works as expected for me.

Although the layout can be a deciding factor in how much you can shrink, with a similar Layout i am able to get:
image

And with scrollbar:
image

chchwy:
Ah! I performed a clean rebuild and now I can see the minimum width getting smaller. 
圖片

chchwy:
Looks good to me!

PR #1917, opened 2025-05-26
Create Qt 6 builds for Linux with linuxdeploy

scribblemaniac:
This adds an additional nightly build package for Linux that uses Qt 6, similar to what is already done on other platforms. Because linuxdeployqt is no longer actively maintained and isn't certain to work with newer Qt 6 versions, this build uses linuxdeploy instead. This provides an opportunity for our users to both test Qt 6 and this new bundling tool for AppImages. After any bugs with linuxdeploy get ironed out, then we can fully transition to it (see #1574).


MrStevns (approved):
Looks good to me. Looks like we should begin thinking about testing more on Qt 6.. now that we have it available on all platforms.

.github/actions/create-package/create-package.sh
```diff
@@ -71,6 +46,43 @@ wayland-decoration-client,wayland-graphics-integration-client,wayland-shell-inte
...
+}
+
+bundle_package_linux_linuxdeploy() {
+${BUILD_CMD} curl -fsSL https://github.com/linuxdeploy/linuxdeploy/releases/download/continuous/linuxdeploy-x86_64.AppImage -o /usr/local/bin/linuxdeploy
+${BUILD_CMD} curl -fsSL https://github.com/linuxdeploy/linuxdeploy-plugin-qt/releases/download/continuous/linuxdeploy-plugin-qt-x86_64.AppImage -o /usr/local/bin/linuxdeploy-plugin-qt
+${BUILD_CMD} curl -fsSL https://github.com/linuxdeploy/linuxdeploy-plugin-appimage/releases/download/continuous/linuxdeploy-plugin-appimage-x86_64.AppImage -o /usr/local/bin/linuxdeploy-plugin-appimage
+${BUILD_CMD} curl -fsSL https://raw.githubusercontent.com/linuxdeploy/linuxdeploy-plugin-gstreamer/refs/heads/master/linuxdeploy-plugin-gstreamer.sh -o /usr/local/bin/linuxdeploy-plugin-gstreamer
+${BUILD_CMD} chmod 755 /usr/local/bin/linuxdeploy /usr/local/bin/linuxdeploy-plugin-qt /usr/local/bin/linuxdeploy-plugin-appimage /usr/local/bin/linuxdeploy-plugin-gstreamer
+export QMAKE=/usr/bin/qmake6
+${BUILD_CMD} linuxdeploy --appdir Pencil2D --plugin qt --plugin gstreamer --output appimage
```
chchwy, line 83: Would you mind adding indentation to this function like others? or any reason the spaces cannot be placed there?

chchwy (approved):

chchwy:
Thanks for taking care of this @scribblemaniac !

PR #1907, opened 2025-03-30
Timeline: Remove ability to add keyframe on double-click

MrStevns:
The amount of feedback we've received regarding this feature has made me re-think its existence.
As such I think it's appropriate to remove it again before the next bug fix release.

PR #1918, opened 2025-05-27
Fix SoundManager calculating frame length incorrectly

J5lx:
SoundManager duplicates SoundClip’s length calculation but without taking the ceil() of the calculated length. This can result in a length of 0 for very short clips, essentially preventing interaction with them. To avoid this, reuse SoundClip’s recalculation mechanism. I believe this is the might be the same issue as reported in [this forum post](https://discuss.pencil2d.org/t/sound-editing-bug-version-0-7/9823?u=j5lx), though the specific durations given don’t particularly make sense to me.

### DIFF
--- a/core_lib/src/managers/soundmanager.cpp
+++ b/core_lib/src/managers/soundmanager.cpp
@@ -146,9 +146,8 @@ void SoundManager::onDurationChanged(SoundPlayer* player, int64_t duration)
 
     double fps = static_cast<double>(editor()->fps());
 
-    double frameLength = duration * fps / 1000.0;
-    clip->setLength(static_cast<int>(frameLength));
     clip->setDuration(duration);
+    clip->updateLength(fps);
 
     editor()->layers()->notifyAnimationLengthChanged();
 


MrStevns (approved):
Not tested but the change makes sense to me.

scribblemaniac (approved):
I was going to look at this last weekend but got sidetracked by some tablet issues. The durations do make more sense if you consider that user typically animates at 8 fps, which would result in a frameLength of 0.8 for 0.1s, and 1.76 for 0.22s. I could reproduce the user's problems with moving and deleting the keyframes and even found an additional crash during playback. This appears to fix all of those issues :star2:

Good to go :+1:

MrStevns:
I was just reminded.. we should ideally be updating the changelog file on every bugfix or improvement we make, so we don't have to do it in bulks like last time.

I've already accumulated the ones I found here:
https://github.com/pencil2d/pencil/pull/1907

but given how long that PR has been lying there now.. I should have made a separate PR for it.

PR #1920, opened 2025-06-08
Fix polyline not creating a backup on end with keyboard

MrStevns:
Small fix for a bug that was just reported on the forum. Apparently we've had it for some time...
It might not be as easy to backmerge as some other things because this uses the old backup system where as that's changed in master.

I still think we should include it though.

I also fixed a behaviour inconsistency between keyboard and mouse where as double-clicking would include the current point on end, doing the same with keyboard would not.

source: https://discuss.pencil2d.org/t/the-polyline-tool-acts-weird-when-trying-to-undo/9833/3


chchwy:
Thanks @MrStevns, I have confirmed the issue is fixed. Good work!

Issue #1871, opened 2024-07-27
Quick Sizing cursor offset changes direction when view is flipped or rotated


HeCorr:
### Issue Summary
It seems the Quick Sizing cursor offset is relative to the canvas instead of absolute.

### Actual Results
When using the Quick Sizing shortcut, the cursor offset direction changes depending on the view rotation or whether it is flipped.

### Expected Results
I would expect the cursor to jump to the top-left of the pointer independently of the view's rotation angle and flip direction.

### Video or Image Reference

https://github.com/user-attachments/assets/0a086e35-9bba-4300-814a-a479f66ae42d

### Steps to reproduce
1. Make sure Quick Sizing functionality is enabled
2. Activate Brush tool
3. Enter Quick Sizing mode and note the offset direction
4. Rotate or flip the view
5. Repeat step 3.

### System Information
+ Pencil2D Version: `Nightly Build Jun 17 2024`

 

+ Operating System: `Windows 10`

 

+ RAM Size: `64 GB`

PR #1921, opened 2025-06-14
Refactor: Quick sizing cursor logic separation

MrStevns:
This is part of a bigger tool logic separation i'm working on...

As my changes were growing, at a similar rate of #1906, I decided to create a separate PR instead, to avoid PR's with lots of changes.

As such this PR focuses on the quick sizing part, separating it into its own tool and let StrokeTool deal with how the canvas cursor should be updated.

Some of the quick sizing logic had also leaked into CanvasCursorPainter, which shouldn't be necessary anymore. I've simplified CanvasCursor to the point where it just focuses on drawing a radial cursor.

In addition I've also updated the quick sizing logic so that it updates the offset based on the current position, rather than a fixed 45 deg angle.

cursor-offset

closes https://github.com/pencil2d/pencil/issues/1871 As it doesn't depend on a fixed angle anymore.

### DIFF
--- a/core_lib/src/tool/stroketool.h
+++ b/core_lib/src/tool/stroketool.h
@@ -103,8 +103,6 @@ protected:
     bool mCanvasCursorEnabled = false;
     QPointF mLastPixel { 0, 0 };
 
-    QPointF mAdjustPosition;
-
     StrokeInterpolator mInterpolator;
 
     const UndoSaveState* mUndoSaveState = nullptr;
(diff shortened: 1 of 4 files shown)


core_lib/src/tool/stroketool.h
```diff
@@ -108,11 +105,16 @@ public slots:
 
     QPointF mAdjustPosition;
```
chchwy: `mAdjustPosition` appears unused now, can you remove it?

MrStevns:
Thanks for reviewing! 🙏 

I have applied the suggestions mentioned and simplified the conditional logic a bit because there was some redundancy

chchwy:
I've done a quick test and it works perfectly. Thank you @MrStevns

PR #1928, opened 2025-08-28
Reposition frame dialog: fix cancel button not working

MrStevns:


MrStevns:
Hmm odd.. the tests work when i run it on my intel mac and use Qt 6. though i'm still on Monterey...

scribblemaniac:
Looks like it could be a similar issue to one I debugged last year caused by certain directories in the data dir path being symlinked in macOS. You should look at `validateDataPath` in util.cpp for how to properly compare directories. From what I remember, comparisons should be with Qt objects (QDir, QFile, etc.), but those objects need to be initialized with the *canonical* path.

MrStevns:
Hmm yes that seems to be the case. I figured that out a few moments ago, thanks :D 

It's a bit difficult to debug when it works on my machine though.. the strange thing is that the layer_bitmap test works while the layer_sound fails.

I'll look deeper into that function.

MrStevns:
I decided to use validateDataPath for the comparison rather than QDir.filepath, so the paths are resolved correctly.

It fixes the tests... but is it correct?

Alternatively I could have done `QDir(frame->fileName()) == QDir(dataDir.filePath(...)).canonicalPath()`

MrStevns:
I've reverted the test changes again as a better fix is provided by #1932

chchwy:
Thanks for catching this!

PR #1930, opened 2025-09-13
Filter release candidates out of update checker

scribblemaniac:
This is a really minimal solution to avoid release candidates from being presented as a new release when users check for updates. The releases.atom does not provide the information necessary to distinguish prereleases from full releases. The nightly build link was wrong so I updated that at the same time.

I would like to eventually use the GitHub API instead, which does provide prerelease information, and fall back to the atom, or perhaps even other sources in case GitHub is down or blocked. However I'm going to leave that as a future improvement as I want to get this fix into v0.7.1. Given how close v0.7.1 is to release, I think this small hack is the best short-term solution.

### DIFF
--- a/app/src/checkupdatesdialog.cpp
+++ b/app/src/checkupdatesdialog.cpp
@@ -111,7 +111,7 @@ void CheckUpdatesDialog::nightlyBuildCheck()
 {
     mTitleLabel->setText(tr("<b>You are using a Pencil2D nightly build</b>"));
     mDetailLabel->setText(tr("Please go %1 here %2 to check new nightly builds.")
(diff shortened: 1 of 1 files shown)


MrStevns:
I personally think it's good that we present release candidates to our users through the update checker. I do however also think that before doing that we might need provide the user with the choice between a stable release and a release candidate with accompanying release notes.

For now this quick fix is fine with me.

chchwy:
Thanks for the fix. Please change the merge target to `release/0.7.1` branch.

chchwy:
Alright I am going to merge it and cherry-pick it to the 0.7.1 branch

PR #1933, opened 2025-09-16
Fix inaccessible top menu bar items

### LINKED ISSUE #1929: 🐛 [BUG] - <title>
### Pencil2D version

nightly build 08/28

### Operating System

Windows

### Steps to reproduce

no response for edit and file tabs

### Expected Behaviour

I expected to open the file and editing tab

### Actual Behaviour

it wont open

### Other Information

I use a drawing tablet

scribblemaniac:
BaseDockWidget was creating empty QWidgets that were children of the main window. As children of the main window they were positioned at the top left of the window by default and I guess they were being set to visible when the main window is shown. These didn't render anything because they're empty, but were blocking mouse events for a small area.

By hiding them right away, we prevent this behavior. I also changed the parents of the title bar widgets to be the BaseDockWidget, so they will be moved to the top left of the dock widget instead. I think the BaseDockWidget makes more sense as a parent logically and assures that if the BaseDockWidget is deleted, its title bars will be deleted right away too (the dock widgets currently last the lifetime of the application so this won't make any different right now).

Fixes #1929

### DIFF
--- a/app/src/basedockwidget.cpp
+++ b/app/src/basedockwidget.cpp
@@ -36,10 +36,11 @@ BaseDockWidget::BaseDockWidget(QWidget* pParent)
     }
 #endif
 
-    mTitleBarWidget = new TitleBarWidget(pParent);
-    mNoTitleBarWidget = new QWidget(pParent);
+    mTitleBarWidget = new TitleBarWidget(this);
+    mNoTitleBarWidget = new QWidget(this);
 
     setTitleBarWidget(mTitleBarWidget);
+    mNoTitleBarWidget->hide();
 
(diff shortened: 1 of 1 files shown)


MrStevns (approved):
Thanks for fixing this, as I wasn't able to reproduce it.
Changes looks good to me

chchwy (approved):

chchwy:
OK, I will merge this PR and cherry pick to the release branch.

chchwy:
I just realized that this PR is for master. don't need to cherry-pick to `release/0.7.1` branch since the release branch doesn't have the latest title bar change.

MrStevns:
Ah... 😅  you're right.. in that case i've changed the milestone to 0.8.0 instead

PR #1936, opened 2025-09-22
Apply transformation before changing keyframe or layer

MrStevns:
I've dabbled with a few solutions and this is the least invasive fix I could come up with.

Fixes a bug where if you had a selection and the image had been transformed, then scrubbing to a different frame would make the new frame inherit the effect.

Before:
transform-bug

After:
transform-bug2

Some users mentioned that this was particularly annoying when scrubbing between frames without explicitly deselecting.

I should mention i'm not fond of adding arbitrary logic to core functions like setLayerIndex and scrubTo. Ideally I would like to maintain the state of the selection and transform as well, by tying the selection to the keyframe but that's a far more invasive which makes no sense for a hotfix release.

### DIFF
--- a/core_lib/src/interface/editor.cpp
+++ b/core_lib/src/interface/editor.cpp
@@ -155,7 +155,7 @@ void Editor::onCurrentLayerWillChange(int index)
 {
     Q_UNUSED(index)
 
-    if (select()->somethingSelected()) {
+    if (select() && select()->somethingSelected()) {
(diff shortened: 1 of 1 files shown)


chchwy:
Thanks & Good job @MrStevns I am going to merge this into 0.7.1 branch.

Would you like I cherry pick your fix to the master branch or you plan to do a more proper fix?

MrStevns:
I am working on reworked model for how we interact with a selection and where the state lives. That upcoming PR should ensure that this workaround becomes obsolete, as such I don't think you need to cherry pick those changes.

PR #1939, opened 2025-10-02
Fix keyframe opacity being adjusted globally in canvas rendering

MrStevns:
The issue was described on Discord earlier today. 

> hello, i have a little issue with the layer opacity. it started when i was making a fade out for some selected layers, first it was the hiena layers, until there it was all good, but when i made the same thing with the thunder effect,for some reason this applied for the whole other layers, i was expecting that the video was going to look like that when finished,but it turned out fine. 
here´s the video below showing what happened.

> https://github.com/user-attachments/assets/f0d9c725-7549-4047-9911-e84b86c393a8

As it was not immediately clear to me what the problem was from the video alone, I asked for the file to investigate. The problem the user is experiencing (that I failed to understand) is that the opacity value is being applied on the entire painter rather than the specific bitmap or vector painter.. same issue with onion skinning. It's easy to reproduce when selecting the "raio" layer the effect is also somewhat achievable on other layers.

Here's a before and after when "raio" is selected on the timeline:

Before:
raio-bug

After:
raio-fixed


chchwy (approved):

chchwy:
Nice catch! Thanks for improving this.

PR #1943, opened 2025-10-28
Fix some file extension issues with image and image sequence export

scribblemaniac:
#1736 Introduced a check to make sure that when a file is selected, that it has an extension matching one of the formats in the file extension filter list. It assumes that there is at least one filter, but this is not true for the image and image sequence exports, which do not have a filter and instead rely upon a dropdown dialog to determine the export format. This PR skips that check if there is no filter, otherwise an assertion will be triggered.

While fixing this, I noticed that single image export doesn't handle file extensions as well as the image sequence export. Notably, it doesn't force the output filename to match the output format, and it will try to export jpgs/bmps with transparency if you check off the transparency box before switching to those formats. I resolved both of these issues by simply copying what image sequence export does. I think there's got to be much better ways to handle this, but I hope that using this tried-and-tested code is sufficient for v0.7.1. I do have plans for support additional image formats in the future which will likely involve rewriting all this code anyway.

### DIFF
--- a/app/src/actioncommands.cpp
+++ b/app/src/actioncommands.cpp
@@ -535,6 +535,39 @@ Status ActionCommands::exportImage()
     QString exportFormat = dialog->getExportFormat();
     bool useTranparency = dialog->getTransparency();
 
+    QString extension = "";
+    QString formatStr = exportFormat;
+    if (formatStr == "PNG" || formatStr == "png")
+    {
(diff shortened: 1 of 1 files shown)


chchwy:
Thanks for catching this and submitting a fix!

chchwy:
The fix was cherry-picked to the master branch.

PR #1944, opened 2025-11-06
Project recovery dialog improvements

scribblemaniac:
Initially this was a fix for the 'x' button of the project recovery dialog not doing anything, however I ended up adding a couple other changes at the same time. There is now a Cancel option in that dialog which will ignore the unrecovered project until next time the program is started. This is an important option when running multiple Pencil2D instances (which you shouldn't do!), and it is in my opinion the most appropriate behavior to take when a user dismisses the dialog without explicitly choosing to open or delete the unrecovered project. This is also now the button that receives focus by default so if the user happens to press enter before reading the dialog, no harm is done.

I also have added support for handling multiple unrecovered projects. Previously, a dialog would only show up for the first unrecovered project, but now it show a dialog for every unrecovered project unless the user has opened one for recovery. It's not the most elegant solution, but it's not a situation that should come up often.

### DIFF
--- a/app/src/mainwindow2.cpp
+++ b/app/src/mainwindow2.cpp
@@ -652,7 +652,7 @@ bool MainWindow2::saveAsNewDocument()
 
 void MainWindow2::openStartupFile(const QString& filename)
 {
-    if (tryRecoverUnsavedProject())
+    if (checkForRecoverableProjects())
     {
         return;
     }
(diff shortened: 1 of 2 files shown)


chchwy:
Thanks for the contribution.

I would question why Pencil2D shouldn't allow multiple instances to run at the same time. Many drawing apps can do this. But we can discuss it later. Well done!

Issue #1894, opened 2024-11-23
🐛 [BUG] - Crash when exporting a project with a sound layer as a movie


DomenicoMammola:
### Description

Hello,
when I try to export as a movie my project I get this error message:

Command: /tmp/.mount_pencilyGVzmc/usr/plugins/ffmpeg -i /tmp/Pencil2D/Default_Y2xD_456aebye/data/sound_20241123_081849_683.wav -filter_complex [0:a:0] aformat=sample_fmts=fltp:sample_rates=44100:channel_layouts=mono,volume=1,adelay=0S|0S,apad=whole_len=3675[ad0] -map [ad0] -ar 44100 -acodec pcm_s16le -ac 2 -y -ss 0 -to 0.0833333 /tmp/Pencil2D-eqoHJY/tmpaudio.wav
ffmpeg version 2.8.17-0ubuntu0.1
 Copyright (c) 2000-2020 the FFmpeg developers
  built with gcc 5.4.0 (Ubuntu 5.4.0-6ubuntu1~16.04.12) 20160609
  configuration: --prefix=/usr --extra-version=0ubuntu0.1 --build-suffix=-ffmpeg --toolchain=hardened --libdir=/usr/lib/x86_64-linux-gnu --incdir=/usr/include/x86_64-linux-gnu --cc=cc --cxx=g++ --enable-gpl --enable-shared --disable-stripping --disable-decoder=libopenjpeg --disable-decoder=libschroedinger --enable-avresample --enable-avisynth --enable-gnutls --enable-ladspa --enable-libass --enable-libbluray --enable-libbs2b --enable-libcaca --enable-libcdio --enable-libflite --enable-libfontconfig --enable-libfreetype --enable-libfribidi --enable-libgme --enable-libgsm --enable-libmodplug --enable-libmp3lame --enable-libopenjpeg --enable-libopus --enable-libpulse --enable-librtmp --enable-libschroedinger --enable-libshine --enable-libsnappy --enable-libsoxr --enable-libspeex --enable-libssh --enable-libtheora --enable-libtwolame --enable-libvorbis --enable-libvpx --enable-libwavpack --enable-libwebp --enable-libx265 --enable-libxvid --enable-libzvbi --enable-openal --enable-opengl --enable-x11grab --enable-libdc1394 --enable-libiec61883 --enable-libzmq --enable-frei0r --enable-libx264 --enable-libopencv
  libavutil      54. 31.100 / 54. 31.100
[...]

PR #1959, opened 2025-12-09
Use Rocky Linux container for Linux nightly builds

scribblemaniac:
This PR moves the Linux nightly builds to Rocky Linux. By doing this, we can build on a system that is still receiving support, while still maintaing a low glibc version for maximum compatibility. The low glibc version version is critical for appimage builds as they will only run on systems with a glibc version equal to or greater than the one it was built with. A list of the main version changes that accompany this switch is listed below.

The only functional change I am aware of is that the version of gstreamer good plugins provided by Rocky Linux's repositories lacks the jack plugin. As far as I can tell this is an oversight upstream and hasn't be reported. I don't think this is a blocker as it is unlikely many of our appimage users are using a sound server targeted towards audio professionals.

### Qt 5 Build

Qt: 5.15.2 -> 5.15.3
FFmpeg: 2.8.17 -> 4.2.10
glibc: 2.23 -> 2.28
Examples of supported distros: Ubuntu 18.04+, Debian >= Buster, Fedora 29+
As far as I can tell all the systems we would be dropping with this change have been EOL for multiple years already.

### Qt 6 Build

Qt: 6.2.4 -> 6.6.2
FFmpeg: 4.4.2 -> 5.1.7
glibc: 3.35 -> 3.34
Examples of supported distros: Ubuntu 21.10+, Debian >= Bookworm, Fedora 35

### DIFF
--- a/.github/actions/install-dependencies/install-dependencies.sh
+++ b/.github/actions/install-dependencies/install-dependencies.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -ex
+set -e
 
 setup_linux() {
   # Because of how bare-bones our docker image is


chchwy (approved):

scribblemaniac:
Closes #1894. The ffmpeg version is also bumped by this change. I have added those versions to the main post.

scribblemaniac:
Note: one user reported on Discord issues with exporting using the nightly build including GIFs not looping and only the start of the animation getting exported on Linux. Through further testing, this build fixed that user's issues, almost certainly because of the updated ffmpeg version.

chchwy:
> Note: one user reported on Discord issues with exporting using the nightly build including GIFs not looping and only the start of the animation getting exported on Linux. Through further testing, this build fixed that user's issues, almost certainly because of the updated ffmpeg version.

Good news. Would you mind fixing the conflict so I can merge the PR?

chchwy:
OK I fixed the merge conflicts.

PR #1965, opened 2025-12-14
[GUI] Replace ToolboxOptions Spinslider/SpinBox with InlineSlider

MrStevns:
This PR introduces a new component I've been experimenting with for some time. It started out on my MyPaint experimental branch but now I think it's finally time to introduce it properly.

#### Horizontal space
Over the years we've had numerous discussions about the size of the interface, particularly the toolbox options and how to make it thinner. Prior to this PR we had a SpinSlider and a QSpinBox next to each other which was mostly fine although depending on the OS, these two components could take up quite a bit of horizontal space. With an application like ours, I think it's crucial that the canvas takes up the most of the space and the lighter we can make the UI, the better.

As with my recent PR's regarding the interface, this too should improve on that situation even more as well as improve on some UI annoyances I've had with the previous layout.

#### Bugs
Although i guess this might be a macOS only but the amount of time i've spend debugging weird issues with the QSlider between Qt 5 and Qt 6... Cosmetical issues mostly but still annoying.

### Introducing: InlineSlider & LineEditNumberWidget 
We now show both the title and the value inside the slider component. It ideally should allow for the widget to be a bit smaller, although this comes down to the OS.

The component can be interacted with in two different ways:
- Hold + drag movement, like a slider.
- Double click to edit like a textbox.

The double click behavior is handled by the `LineEditNumberWidget` component, which takes over as soon as you double click on the component or moves the focus to it with a keyboard.

inlineslider

Darkmode:
image

I've done my best to make sure the component is still accessible with keyboard and tabbing.

The slider has two modes:
- Leading 
- Middle

I don't think we have any tool options right now where the `Middle` style makes sense, so don't think too much of the feather slider using it above, that's just to show an example of it being used.

I've also implemented a fair share of code to ensure the slider retains its look when changing devicePixelRatio and when adjusting the component. Some of this could probably be moved to a general RoundedRect painter that for example the ColorSlider could benefit from as well...

Let me know what you think 😃

### DIFF
--- a/.github/actions/create-package/action.yml
+++ b/.github/actions/create-package/action.yml
@@ -1,4 +1,5 @@
 name: Create Package
+description: Create a release package for the target platform
 inputs:
   arch:
     description: Architecture
@@ -25,4 +26,4 @@ runs:
       INPUT_ARCH: ${{inputs.arch}}
       INPUT_QT: ${{inputs.qt}}
       INPUT_BUNDLER: ${{inputs.bundler}}
-      IS_RELEASE: ${{ startsWith(github.ref, 'refs/heads/release/') }}
+      IS_RELEASE: ${{ startsWith(github.ref, 'refs/heads/release/') || startsWith(github.ref, 'refs/tags/') }}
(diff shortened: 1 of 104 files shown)


MrStevns:
Branch is now up to date with master again

MrStevns:
Note: When #1997 has been merged, I will update this and make it use the new slider utils as this component implements some of the same logic.

PR #1966, opened 2025-12-15
Improve error handling when saving images

scribblemaniac:
This replaces `QImage::save` with `QImageWriter`, which functions the same but provides error codes and messages if the save fails. I also had to refactor some functions to return Status instead of bool to pass along this information. Some error dialogs were added to present users with that information in the event of an error. This applies to saving bitmap frames as well as exporting bitmap images/sequences.

There have been a couple user reports where the details indicate that images have failed to save, which is not very helpful without the reason *why* the images failed to save. This will provide much more insight into what could be going on if we continue to encounter issues like that.


chchwy (approved):

chchwy:
Thank you @scribblemaniac this is a solid improvement that addresses a real pain point, and will make debugging issues much easier in the future.

chchwy:
Cherry-picked the changes to the master branch

PR #1976, opened 2026-01-01
Fix undo/redo being ignored when scrubber is on an empty frame

MrStevns:
Small fix for the new undo/redo system.

To reproduce:

In an empty project:
1. Ensure there's a keyframe on frame 1 and put the scrubber on frame 2
2. Draw a stroke
3. Press undo

Expected:
Stroke is undone

Actual:
Nothing happens

### DIFF
--- a/core_lib/src/managers/undoredomanager.cpp
+++ b/core_lib/src/managers/undoredomanager.cpp
@@ -296,9 +296,10 @@ void UndoRedoManager::initCommonKeyFrameState(UndoSaveState* undoSaveState) cons
     }
 
     const int frameIndex = editor()->currentFrame();
-    if (layer->keyExists(frameIndex))
+    KeyFrame* frame = layer->getLastKeyFrameAtPosition(frameIndex);
+    if (frame)
     {
-        undoSaveState->keyframe = std::unique_ptr<KeyFrame>(layer->getLastKeyFrameAtPosition(frameIndex)->clone());
+        undoSaveState->keyframe = std::unique_ptr<KeyFrame>(frame->clone());
     }
     else if (layer->getKeyFrameWhichCovers(frameIndex) != nullptr)
     {
--- a/core_lib/src/managers/undoredomanager.cpp
+++ b/core_lib/src/managers/undoredomanager.cpp
@@ -296,14 +296,14 @@ void UndoRedoManager::initCommonKeyFrameState(UndoSaveState* undoSaveState) cons
     }
 
     const int frameIndex = editor()->currentFrame();
-    KeyFrame* frame = layer->getLastKeyFrameAtPosition(frameIndex);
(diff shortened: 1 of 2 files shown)


core_lib/src/structure/layervector.h
```diff
@@ -37,7 +37,7 @@ class LayerVector : public Layer
     void loadDomElement(const QDomElement& element, QString dataDirPath, ProgressCallback progressStep) override;
 
     VectorImage* getVectorImageAtFrame(int frameNumber) const;
-    VectorImage* getLastVectorImageAtFrame(int frameNumber, int increment) const;
+    VectorImage* getLastVectorImageAtFrame(int frameNumber, int increment = 0) const;
```
chchwy: I’m not a fan of default arguments. It usually complicates function overload resolution and makes the function signature ambiguous. But we can fix it later on.

core_lib/src/managers/undoredomanager.cpp
```diff
@@ -296,9 +296,10 @@ void UndoRedoManager::initCommonKeyFrameState(UndoSaveState* undoSaveState) cons
...
 
     const int frameIndex = editor()->currentFrame();
-    if (layer->keyExists(frameIndex))
+    KeyFrame* frame = layer->getLastKeyFrameAtPosition(frameIndex);
+    if (frame)
     {
-        undoSaveState->keyframe = std::unique_ptr<KeyFrame>(layer->getLastKeyFrameAtPosition(frameIndex)->clone());
+        undoSaveState->keyframe = std::unique_ptr<KeyFrame>(frame->clone());
     }
     else if (layer->getKeyFrameWhichCovers(frameIndex) != nullptr)
```
chchwy: Is this "else if" condition still reachable? or it's now dead code?

chchwy:
Thanks @MrStevns, I only have one minor question regarding the potential dead code in the "else if" block. Other than that, it's ready to be merged.

chchwy (approved):

core_lib/src/structure/layervector.h
MrStevns: I agree, I only did this because it's inconsistent compared to the bitmap layer. I can revert it but maybe we should consider removing the increment. I don't think we ever use it.

core_lib/src/managers/undoredomanager.cpp
MrStevns: Ah ... you're right. I always forget it's the other way around. First check if the keyframe is covered, and otherwise get the last keyframe. Fixed now.

We really ought to streamline the interface.

core_lib/src/structure/layervector.h
MrStevns: I've reverted the change now.

PR #1983, opened 2026-02-09
Fix eraser bug for vector layers

RayenChalouati:
Fixed issue where the eraser cleared vector paint incorrectly.
Tested locally and it works as expected.


MrStevns (approved):
Seems fine to me and it works as expected.

Thanks for the contribution 🙂

chchwy (approved):

PR #1992, opened 2026-03-08
[Security] Don't allow action to inject arbitrary data directly in to the pipeline

MrStevns:
This prevents an attacker from inputting malicious code directly into the pipeline because the variable will now be expanded when the code is executed, not before, which was the case with a yaml template variable.

Fixes the blockers we currently have in Sonarqube. 

Note: @chchwy please verify that it works as expected, as I have no developer account to test with.


chchwy:
Thanks! it should be fine as long as the CI pipeline passed.

PR #1998, opened 2026-03-31
macOS intel build fix

MrStevns:
- Forces the macOS runners to use intel images.
- Fix the package path. The script assumed a dmg file would always be uploaded for macOS, which fails silently when notarization is skipped.


chchwy:
Thanks! @MrStevns

PR #2020, opened 2026-07-11
Stop double-zipping workflow artifacts

j5lx:
This updates the build workflow to upload the build artefacts without re-compressing them, which [became possible earlier this year](https://github.blog/changelog/2026-02-26-github-actions-now-supports-uploading-and-downloading-non-zipped-artifacts/) and allows us to avoid the somewhat annoying double zip situation. Unfortunately, this requires some additional acrobatics because the upload action currently only allows uploading one non-zipped file at a time. Oh well. When this is merged I’ll adapt the nightly download page accordingly as well.

Also updated the checkout action while I was at it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,7 +66,7 @@ jobs:
     env: { CONTAINER: "${{matrix.container}}", BUILD_CMD: "" }
     steps:
       - name: Check out repository
-        uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


chchwy (approved):

scribblemaniac:
Glad to see this is finally possible! Thanks for implementing this change @j5lx and sorry I didn't get to it sooner.

Issue #1969, opened 2025-12-18
Move Working Directory from System Temp to Application Data


chchwy:
Currently, Pencil2D extracts .pclx files (literally a ZIP) to the system's temporary directory for working directory.

This could have two issues:
- Data loss: System cleanup tools or OS maintenance can delete temp files at any time, e.g., on restart, during disk cleanup, or third-party tools
- Recovery: working files in temp probably already gone after Pencil2D crashes.

## Proposed Solution
Move the working directory to platform-specific application data directories `QStandardPaths::AppLocalDataLocation`

Platform | Location
-- | -- 
Windows | C:\Users\username\AppData\Local\Pencil2D\
macOS | ~/Library/Application Support/Pencil2D/
Linux | ~/.local/share/Pencil2D/ (XDG_DATA_HOME)

Notes

- Working files always persist except someone explicitly cleans it up (Good for Crash recovery)
- AppData directories should always be writable
- Other people cannot access the personal AppData so it has better privacy

PR #2027, opened 2026-09-10
#1969 Move working directory from system temp to AppLocalDataLocation

chchwy:
This PR is for #1969

Previously, .pclx working files were extracted to the system temp directory (QDir::tempPath()), making them vulnerable to data loss from OS cleanup tools, disk cleaners, or system restarts. This also meant crash recovery was impossible since temp files would already be gone.

Working directories are now stored under:
- Windows: `C:\Users\<user>\AppData\Local\Pencil2D\working\`
- macOS: `~/Library/Application Support/Pencil2D/working/`
- Linux: `~/.local/share/Pencil2D/working/`


j5lx:
Very welcome change. However I believe these files should really be located in StateLocation. This won’t make much of a difference on Windows, where it ultimately ends up within AppLocalDataLocation anyway, but Linux has a separate location for that nowadays.

Unfortunately, Qt only added support for StateLocation in 6.7, so I suppose we’ll still have to fall back to AppLocalDataLocation as the next best thing when compiling against older versions. Not great but eh.

Also, with working files now having their own dedicated directory, I assume PFF_TMP_DECOMPRESS_EXT can be retired, no?

Anyway, good to see these files moved out of the system temp directory.

